PR #27, opened 2023-02-21
copy_paste utility function

hansent:
# Description

This adds a `copy_paste` utility function to copy paste specific areas from one image into another.  Similar to e.g. the functionality in this notebook:  https://colab.research.google.com/drive/1VK5FPT41pQt4pVxMRn8oaYxPfAD_jUOa?usp=sharing#scrollTo=03bUCdJRNXux

We are looking at using this functionality in this project https://github.com/roboflow/magic-scissors and thought it would live well as a general too in supervision.

I've been testing this using the following script:
```
import cv2
import numpy as np
import supervision

cart_image = cv2.imread("./cart1.jpg")

jelly_image = cv2.imread("./jelly.jpg")
jelly_polygon_raw = [640, 374.162, 601.785, 188.653, 566.854, 151.682, 524.031, 88.138,
                     464.306, 41.928, 397.819, 8.423, 360.632, 8.423, 299.781, 0.337,
                     232.169, 6.115, 169.063, 33.842, 110.463, 71.967, 67.64, 121.643,
                     27.073, 175.943, 0.028, 282.231, 10.17, 394.298, 43.976, 470.547,
                     108.212, 554.887, 203.994, 605.716, 222.027, 621.892, 406.35, 640,
                     410.855, 640, 565.729, 547.955, 640, 374.894, 640, 374.162]
jelly_polygon = np.array(jelly_polygon_raw).astype(np.int32).reshape(-1, 2)

output = supervision.draw.utils.copy_paste(
    jelly_image,  # source image to copy / cutout from
    jelly_polygon,  # polygon area to cutout from source
    cart_image,  # background / target image
    0.2,  # scale
    300,  # x
    200  # y
)

cv2.imshow('output', output)
cv2.waitKey(0)
cv2.destroyAllWindows()
```

### DIFF
--- a/supervision/draw/utils.py
+++ b/supervision/draw/utils.py
@@ -175,29 +175,49 @@ def draw_text(
     return scene
 
 
-def copy_paste(source_image: np.ndarray, source_polygon: np.ndarray, target_image: np.ndarray):
+def copy_paste(source_image: np.ndarray, source_polygon: np.ndarray, target_image: np.ndarray, scale: float = 1.0, x: int = 0, y: int = 0):
     """
     Copy and paste a region from a source image into a target image.
 
     Attributes:
         source_image (np.ndarray): The image to be copied.
-        source_polygon (np.ndarray): The polygon to be copied.
+        source_polygon (np.ndarray): The polygon area to be copied.
         target_image (np.ndarray): The image to be pasted into.
+        scale: scale factor to apply to the source image
+        x: x offset to paste into the target image at
+        y: y offset to paste into the target image at
 
 
     Returns:
         np.ndarray: The target image with the source image pasted into it.
     """
-    cv2.imshow('source', source_image)
-    cv2.imshow('target', target_image)
-    # print(source_image.shape)
+    # cv2.imshow('source', source_image)
+    # cv2.imshow('target', target_image)
+    # # print(source_image.shape)
 
-    width, height, depth = source_image.shape
-    mask = generate_2d_mask(source_polygon, (width, height))
-    cv2.imshow('mask', mask)
+    output = target_image.copy()
 
-    cutout = cv2.copyTo(source_image, mask, target_image)
-    cv2.imshow('cutout', cutout)
+    # generate mask image based on polygon
+    source_width = source_image.shape[0]
+    source_height = source_image.shape[1]
+    mask = generate_2d_mask(source_polygon, (source_width, source_height))
 
-    cv2.waitKey(0)
-    cv2.destroyAllWindows()
+    # cutout = cv2.bitwise_and(self.crop, self.crop, mask=self.mask)
+
+    # scale the source and mask
+    scaled_source = cv2.resize(source_image, None, fx=scale, fy=scale)
+    scaled_mask = cv2.resize(mask, None, fx=scale, fy=scale)
+    scaled_with = scaled_source.shape[0]
+    scaled_height = scaled_source.shape[1]
+
+    # generate a patch from the source image  / with background from target image where we are pasting
+    patch = np.where(
+        np.expand_dims(scaled_mask, axis=2),
+        scaled_source,
+        target_image[y:y+scaled_height, x:x+scaled_with, :]
+    )
+
+    # paste the patch area into the output image
+    output[y:y+scaled_height, x:x+scaled_with, :] = patch
+
+    return output
--- a/supervision/draw/utils.py
+++ b/supervision/draw/utils.py
@@ -3,6 +3,7 @@ from typing import Optional
 import cv2
 import numpy as np
 
+from supervision.detection.utils import generate_2d_mask
 from supervision.draw.color import Color
 from supervision.geometry.core import Point, Rect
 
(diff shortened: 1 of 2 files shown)


supervision/draw/utils.py
```diff
@@ -171,3 +172,58 @@ def draw_text(
...
+        target_image (np.ndarray): The image to be pasted into.
+        scale: scale factor to apply to the source image
+        x: x offset to paste into the target image at
+        y: y offset to paste into the target image at
+
+
+    Returns:
+        np.ndarray: The target image with the source image pasted into it.
+    """
+    # cv2.imshow('source', source_image)
```
SkalskiP, line 200: Please remove those commented lines

supervision/draw/utils.py
```diff
@@ -171,3 +172,58 @@ def draw_text(
...
+
+
+def copy_paste(
+    source_image: np.ndarray,
+    source_polygon: np.ndarray,
+    target_image: np.ndarray,
+    scale: float = 1.0,
+    x: int = 0,
+    y: int = 0,
+):
```
SkalskiP, line 184: Please add the return type.

supervision/draw/utils.py
```diff
@@ -171,3 +172,58 @@ def draw_text(
...
+
+
+    Returns:
+        np.ndarray: The target image with the source image pasted into it.
+    """
+    # cv2.imshow('source', source_image)
+    # cv2.imshow('target', target_image)
+    # # print(source_image.shape)
+
+    output = target_image.copy()
```
SkalskiP, line 204: This will be super slow. Do we want/need to copy image?

supervision/draw/utils.py
```diff
@@ -171,3 +172,58 @@ def draw_text(
...
+
+    # generate mask image based on polygon
+    source_width = source_image.shape[0]
+    source_height = source_image.shape[1]
```
SkalskiP, line 208: Let's do it in a single line.
```
source_width = source_image.shape[0]
source_height = source_image.shape[1]
```

supervision/draw/utils.py
```diff
@@ -171,3 +172,58 @@ def draw_text(
...
+    mask = generate_2d_mask(source_polygon, (source_width, source_height))
+
+    # cutout = cv2.bitwise_and(self.crop, self.crop, mask=self.mask)
```
SkalskiP, line 211: Let's remove those commented lines.

supervision/draw/utils.py
```diff
@@ -171,3 +172,58 @@ def draw_text(
...
+
+    # scale the source and mask
```
SkalskiP, line 213: The order of operations makes it inefficient.

1. Scale source <- expensive
2. Scale mask <- expensive
3. Extract patch from source <- cheap

Let's refactor and make it:

1. Extract patch from source <- cheap
2. Scale patch <- expensive

supervision/draw/utils.py
```diff
@@ -171,3 +172,58 @@ def draw_text(
...
+    scaled_source = cv2.resize(source_image, None, fx=scale, fy=scale)
+    scaled_mask = cv2.resize(mask, None, fx=scale, fy=scale)
+    scaled_with = scaled_source.shape[0]
+    scaled_height = scaled_source.shape[1]
```
SkalskiP, line 217: Let's do it in a single line. 

```
scaled_with = scaled_source.shape[0]
scaled_height = scaled_source.shape[1]
```

`scaled_with` <- typo

SkalskiP (changes requested):

supervision/draw/utils.py
hansent: mhh, I see what you're saying but I don't think that works with the current approach.

The `patch` is an image with both source image pixels and the background pixels from the target area.  So if we want to scale the patch, we would have to upscale the background first to generate the patch.  Maybe there is a better way to do the pasting / merging of pixels instead of this?:
```
    # generate a patch from the source image  / with background from target image where we are pasting
    patch = np.where(
        np.expand_dims(scaled_mask, axis=2),
        scaled_source,
        target_image[y : y + scaled_height, x : x + scaled_width, :],
    )
```

supervision/draw/utils.py
hansent: :+1:  can do in place on the target image instead

supervision/draw/utils.py
hansent: I think I'm skipping the step of extracting just a section of the source image (for the magic scissors use case its already cropped to bounding box of polygon)....so actually I think maybe want to add that part first, and then scale the section extracted/cropped from source.

SkalskiP: Oh, so the `patch` has the same shape as the target image? So if the image is 4K and the object we are copying is `400x400` we will still create a 4K patch?

Don't get me wrong, I can merge it as it is, but it sounds like something we could optimize quite fast. Because that line:

```
patch = np.where(
    np.expand_dims(scaled_mask, axis=2),
    scaled_source,
    target_image[y : y + scaled_height, x : x + scaled_width, :],
)
```

Is essentially a hidden image copy.

PR #34, opened 2023-03-03
added polygon zone to redme

ogjaylowe:
# Description
Added polygon zone to the readme so devs know it exist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,12 @@ Learn how to use YOLOv8, ByteTrack and **Supervision** to detect, track and coun
     </a>
 </p>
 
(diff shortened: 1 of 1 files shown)


SkalskiP:
I think we should have it in docs rather than `README.md`. What do you think?

capjamesg:
@SkalskiP This feature has gotten a lot of traction so I'd recommend we have it in the README, perhaps labelled as a "supporting utility" that works well with the library?

SkalskiP:
@capjamesg I don't think `README.md` is the right place.

PR #110, opened 2023-05-23
Add semantic segmentation polygon loading from Roboflow

capjamesg:
# Description

This PR adds support for loading the `points` values from Roboflow API detections into segmentation masks.

## Type of change

This code is used in `cvevals` for ingesting masks.


supervision/detection/core.py
```diff
@@ -324,6 +326,28 @@ def from_roboflow(cls, roboflow_result: dict, class_list: List[str]) -> Detectio
             class_id.append(class_list.index(prediction["class"]))
             confidence.append(prediction["confidence"])
 
+            if prediction.get("points"):
+                points = prediction["points"]
+
+                mask = np.zeros((int(height), int(width)), dtype=np.uint8)
```
SkalskiP, line 332: Most of that logic can be replaced with `polygon_to_mask` function that can create a mask given `polygon`. Please use it. You can find it in `supervision.detection.utils`.

supervision/detection/core.py
```diff
@@ -324,6 +326,28 @@ def from_roboflow(cls, roboflow_result: dict, class_list: List[str]) -> Detectio
...
+                    [(p["x"], p["y"]) for p in points], dtype=np.int32
+                ).reshape((-1, 1, 2))
+
+                cv2.fillPoly(mask, [points], 1)
+
+                mask = mask.astype(bool)
+
+                masks.append(mask)
+
+        if len(mask) > 0:
```
SkalskiP, line 343: Let's keep a single return and make it a bit more compact:

```python
return Detections(
    xyxy=np.array(xyxy),
    confidence=np.array(confidence),
    class_id=np.array(class_id).astype(int),
    mask=mask if mask else None,
)
```

supervision/detection/core.py
```diff
@@ -324,6 +326,28 @@ def from_roboflow(cls, roboflow_result: dict, class_list: List[str]) -> Detectio
...
+            return Detections(
+                xyxy=np.array(xyxy),
+                confidence=np.array(confidence),
+                class_id=np.array(class_id).astype(int),
+                mask=np.array(mask),
```
SkalskiP, line 348: Make sure that type of mask is `dtype=bool`.

supervision/detection/core.py
```diff
@@ -324,6 +326,28 @@ def from_roboflow(cls, roboflow_result: dict, class_list: List[str]) -> Detectio
             class_id.append(class_list.index(prediction["class"]))
             confidence.append(prediction["confidence"])
 
+            if prediction.get("points"):
```
SkalskiP, line 329: Let's prevent adding indented logic - reverse the logical condition and use `continue`.

```python
if "points" not in prediction:
    continue
```

SkalskiP:
Adding some unit tests for `from_roboflow` would be awesome 😎

SkalskiP (changes requested):

capjamesg:
@SkalskiP I have incorporated your feedback in my last commit. Please review. I have not added a test case, however.

SkalskiP:
@capjamesg do you have some code that I could use to test that code?

PR #125, opened 2023-06-07
Add ClassificationDataset API

capjamesg:
# Description

This PR introduces a new `ClassificationDataset` API for use in working with classification datasets using supervision. Users can:

- Create a ClassificationDataset object;
- Save data in the multiclass classification folder, and;
- Load data from a dataset structured as a multiclass classification folder.

## Type of change

I will add a test case once we have discussed the API.

## Docs

I will add the requisite docs once we have discussed the API.


supervision/__init__.py
```diff
@@ -1,5 +1,6 @@
 __version__ = "0.9.0"
 
+from supervision.classification.core import ClassificationDataset, Classifications
```
SkalskiP, line 3: This import is dead. This package does not exist.

supervision/dataset/classification.py
```diff
@@ -0,0 +1,239 @@
...
+            ...     root_directory_path="./dataset",
+            ... )
+        """
+        classes = os.listdir(root_directory_path + "/train")
+        classes.sort()
+
+        images = {}
+        annotations = {}
+
+        for split in ["train", "test", "val"]:
```
SkalskiP, line 224: The rule that we follow in `DetectionDataset` is that you create separate datasets objects of `train`, `test`, and `val`. This is pretty much the standard in deep learning community. 

You can take a look at those examples:
- Transformers training: https://colab.research.google.com/github/roboflow-ai/notebooks/blob/main/notebooks/train-huggingface-detr-on-custom-dataset.ipynb
- YOLO-NAS training: https://colab.research.google.com/github/roboflow-ai/notebooks/blob/main/notebooks/train-yolo-nas-on-custom-dataset.ipynb
- Detectron2 training: https://colab.research.google.com/github/roboflow-ai/notebooks/blob/main/notebooks/train-detectron2-segmentation-on-custom-data.ipynb

I would like the classification API to be consistent. So your `root_directory_path` should essentially lead to the directory that contains class subdirectories.

supervision/dataset/classification.py
```diff
@@ -0,0 +1,239 @@
...
+            for class_name in classes:
+                for image in os.listdir(
+                    root_directory_path + "/" + split + "/" + class_name
```
SkalskiP, line 227: Please do not build paths using `+ "/" +` . This is not secure. Python has the tools to do that. For example, `os.path.join`.

supervision/dataset/classification.py
```diff
@@ -0,0 +1,239 @@
...
+                ):
+                    images[image] = np.array([[[0, 0, 0]]])
```
SkalskiP, line 229: What is happening here?

supervision/dataset/classification.py
```diff
@@ -0,0 +1,239 @@
...
+                    os.makedirs(output_directory_path + "/" + split + "/" + class_name)
+
+                for image in dataset.images:
+                    if dataset.annotations[image].class_id[0] == self.classes.index(
+                        class_name
+                    ):
+                        full_dir = os.path.join(root_directory_path, image)
+
+                        cv2.imwrite(
+                            output_directory_path
```
SkalskiP, line 187: Please do not build paths using `+ "/" +` . This is not secure. Python has the tools to do that. For example, `os.path.join`.

supervision/dataset/classification.py
```diff
@@ -0,0 +1,239 @@
...
+            ... )
+        """
+        if not os.path.exists(output_directory_path):
+            os.makedirs(output_directory_path)
+
+        for split in ["train", "test", "valid"]:
+            if not os.path.exists(output_directory_path + "/" + split):
+                os.makedirs(output_directory_path + "/" + split)
+
+        train, test = self.split(split_ratio=0.8)
```
SkalskiP, line 170: Splitting the dataset should not be the responsibility of this method.

supervision/dataset/classification.py
```diff
@@ -0,0 +1,239 @@
...
+    """
+    Ensure that confidence is a 1d np.ndarray with (n, ) shape.
+    """
+    is_valid = isinstance(confidence, np.ndarray) and confidence.shape == (n,)
+    if not is_valid:
+        raise ValueError("confidence must be 1d np.ndarray with (n, ) shape")
+
+
+@dataclass
+class Classifications:
```
SkalskiP, line 31: This class should not live in the `dataset` package.

supervision/dataset/classification.py
```diff
@@ -0,0 +1,239 @@
...
+            >>> cd.as_multiclass_folder_structure(
+            ...     root_directory_path="./images",
+            ...     output_directory_path="./out",
+            ... )
+        """
+        if not os.path.exists(output_directory_path):
+            os.makedirs(output_directory_path)
+
+        for split in ["train", "test", "valid"]:
+            if not os.path.exists(output_directory_path + "/" + split):
```
SkalskiP, line 167: I refer to other comments about path construction.

supervision/dataset/classification.py
```diff
@@ -0,0 +1,239 @@
...
+                os.makedirs(output_directory_path + "/" + split)
+
+        train, test = self.split(split_ratio=0.8)
+        train, val = train.split(split_ratio=0.8)
+
+        for split, dataset in zip(["train", "test", "valid"], [train, test, val]):
+            for class_name in self.classes:
+                if not os.path.exists(
+                    output_directory_path + "/" + split + "/" + class_name
```
SkalskiP, line 176: I refer to other comments about path construction.

supervision/dataset/classification.py
```diff
@@ -0,0 +1,239 @@
...
+                ):
+                    os.makedirs(output_directory_path + "/" + split + "/" + class_name)
```
SkalskiP, line 178: I refer to other comments about path construction.

supervision/dataset/classification.py
```diff
@@ -0,0 +1,239 @@
...
+
+            >>> cd.as_multiclass_folder_structure(
+            ...     root_directory_path="./images",
+            ...     output_directory_path="./out",
+            ... )
+        """
+        if not os.path.exists(output_directory_path):
+            os.makedirs(output_directory_path)
+
+        for split in ["train", "test", "valid"]:
```
SkalskiP, line 166: I refer to other comments about separate `ClassificationDataset` object per split.

supervision/dataset/classification.py
```diff
@@ -0,0 +1,239 @@
...
+            if not os.path.exists(output_directory_path + "/" + split):
+                os.makedirs(output_directory_path + "/" + split)
+
+        train, test = self.split(split_ratio=0.8)
+        train, val = train.split(split_ratio=0.8)
+
+        for split, dataset in zip(["train", "test", "valid"], [train, test, val]):
```
SkalskiP, line 173: I refer to other comments about separate `ClassificationDataset` object per split.

SkalskiP (changes requested):

SkalskiP:
A summary:
- [ ] `Classifications` class should not be in the `datasets` package.
- [ ] Splitting dataset is not a responsibility of `as_multiclass_folder_structure`.
- [ ] Please do not use`+ "/" +` to build file paths.
- [ ] Classification API should be consistent with detections API. When users want to load train, test, and validation subsets, they should create three separate instances of `ClassificationDataset`. You can find more info in one of the code comments.
- [ ] Running `as_multiclass_folder_structure` and `from_multiclass_folder_structure` should result in precisely the same dataset.

supervision/dataset/classification.py
SkalskiP:
```
path = "/home" + "/" + "user" + "/" + "documents" # This will work on Unix-based systems, but not on Windows
```
```
path = "C:\\Users" + "\\" + "user" + "\\" + "documents" # This is how paths are typically structured on Windows
```

supervision/dataset/classification.py
capjamesg: Can you elaborate? I am confused.

supervision/dataset/classification.py
capjamesg: `images: Dict[str, np.ndarray]` is the data type for `images`. What should be the value in this dictionary?

supervision/dataset/classification.py
capjamesg: I assume we need a separate `split()` method?

supervision/dataset/classification.py
capjamesg: Where should it be?

SkalskiP: I'd probably create a new file `supervision.classification.core` and move that part of the code there.

supervision/dataset/classification.py
SkalskiP: I guess we already have `split` method. You implemented it already.

supervision/dataset/classification.py
SkalskiP: It should be `np.array` - the result of loading the image, at least for now. I plan to introduce lazy loading, but you should load the image here for now.

capjamesg: Got it!

capjamesg:
@SkalskiP I have responded to a lot of your feedback in my latest commit. Please review. I am still confused about this comment:

> Splitting dataset is not a responsibility of as_multiclass_folder_structure.

Can you outline in pseudocode the expected way this function would work?

supervision/__init__.py
SkalskiP: `Classifications` class should be in `supervision.classification.core`, but `ClassificationDataset` class should be in `supervision.dataset.core`

supervision/classification/core.py
```diff
@@ -0,0 +1,226 @@
...
+            >>> cd.as_multiclass_folder_structure(
+            ...     root_directory_path="./images/valid/",
+            ...     output_directory_path="./out",
+            ... )
+        """
+        if not os.path.exists(output_directory_path):
+            os.makedirs(output_directory_path)
+
+        for class_name in self.classes:
+            if not os.path.exists(os.path.join(output_directory_path, class_name)):
```
SkalskiP, line 167: I'd create a separate variable to store `os.path.join(output_directory_path, class_name)` value, and name in a descriptive way.

supervision/classification/core.py
```diff
@@ -0,0 +1,226 @@
...
+                os.makedirs(os.path.join(output_directory_path, class_name))
+
+            for image in self.annotations:
+                if self.annotations[image].class_id[0] == self.classes.index(
```
SkalskiP, line 171: What if `class_id` is not sorted? I guess, picking an element with a `0` index will be wrong.

supervision/classification/core.py
```diff
@@ -0,0 +1,226 @@
...
+                    class_name
+                ):
+                    full_dir = os.path.join(root_directory_path, class_name, image)
+
+                    cv2.imwrite(
+                        os.path.join(output_directory_path, class_name, image),
+                        cv2.imread(full_dir),
```
SkalskiP, line 178: As I said, there is no need to load that image. It is already loaded.

supervision/classification/core.py
```diff
@@ -0,0 +1,226 @@
...
+            ```python
+            >>> import supervision as sv
+
+            >>> cd = sv.ClassificationDataset.from_multiclass_folder_structure(
+            ...     root_directory_path="./dataset",
+            ... )
+        """
+        classes = os.listdir(root_directory_path)
+
+        classes = list(set(classes))
```
SkalskiP, line 204: We can do it simpler. Instead of:

```python
classes = list(set(classes))
classes.sort()
```

You can do:

```python
classes = sorted(set(classes))
```

supervision/classification/core.py
```diff
@@ -0,0 +1,226 @@
...
+        classes.sort()
+
+        images = {}
+        annotations = {}
+
+        for class_name in classes:
+            if not os.path.exists(os.path.join(root_directory_path, class_name)):
```
SkalskiP, line 211: Is it possible that `os.path.join(root_directory_path, class_name)` does not exist if we got that value from `os.listdir(root_directory_path)`?

supervision/classification/core.py
```diff
@@ -0,0 +1,226 @@
...
+                continue
+
+            for image in os.listdir(os.path.join(root_directory_path, class_name)):
+                images[image] = cv2.imread(
+                    os.path.join(root_directory_path, class_name, image)
```
SkalskiP, line 216: I'd create separate variable to store this value `os.path.join(root_directory_path, class_name, image)`.

supervision/classification/core.py
```diff
@@ -0,0 +1,226 @@
...
+                )
+                annotations[image] = Classifications(
+                    class_id=np.array([classes.index(class_name)]),
```
SkalskiP, line 219: You can calculate `classes.index(class_name)` at the top of the first for loop, it will not get changed until next `class_name`.

supervision/classification/core.py
```diff
@@ -0,0 +1,226 @@
...
+            >>> cd = sv.ClassificationDataset(...)
+
+            >>> cd.annotations["image.png"].get_top_k(1)
+
+            (array([1]), array([0.9]))
+        """
+        if self.confidence is None:
+            raise ValueError("confidence is None")
+
+        confidence = self.confidence.copy()
```
SkalskiP, line 69: Looks like:

```python
confidence = self.confidence.copy()
class_ids = self.class_id.copy()
```

is redundant. We do not need to do it.

supervision/classification/core.py
capjamesg: Good point. What is the best way to handle this?

capjamesg:
My latest commit incorporates all your feedback. I left a comment in response to one of your comments re: the class_id [0] issue.

supervision/dataset/core.py
```diff
@@ -330,3 +332,141 @@ def as_yolo(
...
+            ...     output_directory_path="./out",
+            ... )
+        """
+        os.makedirs(output_directory_path, exist_ok=True)
+
+        for class_name in self.classes:
+            os.makedirs(os.path.join(output_directory_path, class_name), exist_ok=True)
+
+            for image in self.annotations:
+                for class_id in self.annotations[image].class_id:
```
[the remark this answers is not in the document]
SkalskiP: I'd reimplemented that method from:

```python
os.makedirs(output_directory_path, exist_ok=True)

for class_name in self.classes:
    os.makedirs(os.path.join(output_directory_path, class_name), exist_ok=True)

    for image in self.annotations:
        for class_id in self.annotations[image].class_id:
            if class_id == self.classes.index(class_name):
                cv2.imwrite(
                    os.path.join(output_directory_path, class_name, image),
                    self.images[image],
                )
```

To:

```python
os.makedirs(output_directory_path, exist_ok=True)

for class_name in self.classes:
    os.makedirs(os.path.join(output_directory_path, class_name), exist_ok=True)
    
for image_name in self.images:
    classification = self.annotations[image_name]
    image = self.images[image_name]
    class_id = classification.class_id[0] \
        if classification.confidence is None \
        else classification.get_top_k(1)[0]
    class_name = self.classes[class_id]
    image_path = os.path.join(output_directory_path, class_name, image_name)
    cv2.imwrite(image_path, image)
```

supervision/dataset/core.py
```diff
@@ -330,3 +332,141 @@ def as_yolo(
...
+                    if class_id == self.classes.index(class_name):
+                        cv2.imwrite(
+                            os.path.join(output_directory_path, class_name, image),
+                            self.images[image],
+                        )
+
+    @classmethod
+    def from_multiclass_folder_structure(
+        cls, root_directory_path: str
+    ) -> "ClassificationDataset":
```
SkalskiP, line 433: Because we use `from __future__ import annotations` at the top of the file, we do not need `"` around `"ClassificationDataset"`.

test/classification/test_core.py
```diff
@@ -0,0 +1,39 @@
+from typing import List, TypeVar, Optional, Tuple
+from contextlib import ExitStack as DoesNotRaise
+
+import pytest
+import numpy as np
+
+from supervision.classification.core import Classifications
+
+T = TypeVar("T")
```
SkalskiP, line 9: Why did you decided to use `TypeVar`?

test/classification/test_core.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+@pytest.mark.parametrize(
+    'class_id, confidence, expected_result, exception',
+    [
+        (
+            [0, 1, 2, 3, 4],
```
SkalskiP, line 16: Why pass `list` and not just `np.array`?

test/classification/test_core.py
```diff
@@ -0,0 +1,39 @@
...
+        ),  # class_id with 5 numbers and 5 confidences
+        (
+            [0, 1, 2, 3, 4],
+            [0.1, 0.2, 0.3, 0.4],
+            None,
+            pytest.raises(ValueError)
+        ),  # class_id with 5 numbers and 4 confidences
+    ]
+)
+def test_top_k(
```
SkalskiP, line 29: I think we need to test with different `k` values. Do not assume that it is always `len(class_id)`.

test/classification/test_core.py
```diff
@@ -0,0 +1,39 @@
...
+            [0.1, 0.2, 0.9, 0.4, 0.5],
+            (np.array([2, 4, 3, 1, 0]), np.array([0.9, 0.5, 0.4, 0.2, 0.1])),
+            DoesNotRaise()
+        ),  # class_id with 5 numbers and 5 confidences
+        (
+            [0, 1, 2, 3, 4],
+            [0.1, 0.2, 0.3, 0.4],
+            None,
+            pytest.raises(ValueError)
+        ),  # class_id with 5 numbers and 4 confidences
```
SkalskiP, line 26: Let's add a case when `confidence=None`.

test/classification/test_core.py
```diff
@@ -0,0 +1,39 @@
...
+def test_top_k(
+    class_id: List[T],
+    confidence: Optional[List[T]],
+    expected_result: Optional[Tuple[List[T], List[T]]],
+    exception: Exception
+) -> None:
+    with exception:
+        result = Classifications(class_id=np.array(class_id), confidence=np.array(confidence)).get_top_k(len(class_id))
+
+        assert result[0].tolist() == expected_result[0].tolist()
```
SkalskiP, line 38: Please use `np.array_equal(result[0], expected_result[0])` not `result[0].tolist() == expected_result[0].tolist()`.

SkalskiP (changes requested):

capjamesg:
I have incorporated all of your feedback into my latest commit. We now have five test cases for the `get_top_k()` function.

Issue #160, opened 2023-06-28
Add `start` and `end` arguments to `get_video_frames_generator`


SkalskiP:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Description

Add two new arguments to the `get_video_frames_generator` function:
- `start` - index of the first frame that will be accessed
- `end` - index of the last frame that will be accessed

### Use case

We can use `start` and `end` arguments to define the range of video frames we want to access.

```python
>>> import supervision as sv

>>> for frame in sv.get_video_frames_generator(source_path='source_video.mp4', stride=2, start=100, end=200):
...     pass
```

SkalskiP:
@hardikdava most of the logic is already implemented here: https://github.com/roboflow/supervision/blob/main/supervision/utils/video.py

hardikdava:
@SkalskiP should we close this one? And start working on next thing?

SkalskiP:
@hardikdava absolutely :) I didn't notice it was still open after the merge. GH sometimes closes tied issues automatically.

Issue #138, opened 2023-06-14
Error when running PR checks 


danigarciaoca:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Question

Hi everyone! I've one question realted to PR and Test Workflow.

I've seen that, when commiting my PR (#135) I get an error in all checks related to Test Workflow due to problems in fetching the respository. Here are [the errors](https://github.com/roboflow/supervision/actions/runs/5265179974).

![245773658-ef3286fa-35ac-45f5-9e26-cd94df97c1a4](https://github.com/roboflow/supervision/assets/27818263/efda20d9-a347-4c3b-918d-7e01963422f6)

Does anyone have an idea about what can be happening? I've seen that some other users that have created and merged a PR haven't experienced this issue. Some examples of successful PRs:
- [Update `__init__.py` to include linezone](https://github.com/roboflow/supervision/pull/31), by @dceejay 
- [Add custom text for the in & out tags](https://github.com/roboflow/supervision/pull/44), by @iPoe 
- [Adding yolov8 masks to detections](https://github.com/roboflow/supervision/pull/101), by @hardikdava 

Thanks in advance!

SkalskiP:
Hi @dankresio 👋🏻 ! It is most likely a problem with our GitHub Action. This only happens if you try to marge from the fork of Supervision. I will look into it. 👀

dankresio:
Oh, thanks! I thought it was a problem with private repositories or my Github PAT.

Regarding this comment:
> This only happens if you try to marge from the fork of Supervision

Some other PRs (#101, #44) have been merged from the fork and there have been no problems with their checks. That was the point that was blowing my mind. Its weird...

hardikdava:
@SkalskiP I think this problem rises when we use different branch from a fork respository i.e. other than `main` branch.

SkalskiP:
@hardikdava that idea looks plausible. Let's keep our eyes on future PRs and try to confirm that.

hardikdava:
@SkalskiP I noticed the following thing:

fork_repo/any_branch -> supervision/other than main branch :+1: 
fork_repo/any_branch -> supervision/main branch :-1:

onuralpszr:
@SkalskiP  has been fix via #218 so we can close this too 🎉 🎉 🎉 🎉

SkalskiP:
@onuralpszr closing!

PR #162, opened 2023-06-28
video generator with start and end position feature

hardikdava:
# Description

Related to #160 

## Type of change

## Docs

### DIFF
--- a/supervision/dataset/core.py
+++ b/supervision/dataset/core.py
@@ -409,7 +409,7 @@ class DetectionDataset(BaseDataset):
         annotations_path: Optional[str] = None,
         min_image_area_percentage: float = 0.0,
         max_image_area_percentage: float = 1.0,
-        approximation_percentage: float = 0.0
+        approximation_percentage: float = 0.0,
     ) -> None:
         """
         Exports the dataset to COCO format. This method saves the images and their corresponding
@@ -439,7 +439,7 @@ class DetectionDataset(BaseDataset):
                 classes=self.classes,
                 min_image_area_percentage=min_image_area_percentage,
                 max_image_area_percentage=max_image_area_percentage,
-                approximation_percentage=approximation_percentage
+                approximation_percentage=approximation_percentage,
             )
 
     @classmethod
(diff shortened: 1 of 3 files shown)


hardikdava:
@SkalskiP Please review it and merge it. I think docs needs to update for `start` and `end` argument.

SkalskiP:
Awesome! 🔥  @hardikdava, did you test that on some example videos?

hardikdava:
@SkalskiP I did some testing with various input parameters. It seems to work fine. I tried to improve the function with simple logic. Feel free to change or test it.

hardikdava:
Hi @SkalskiP, I am wondering why these checks are failing. Do you know about it?

hardikdava:
@SkalskiP I have a feeling that `yield` is slower compare to normal video reading.

SkalskiP:
> I am wondering why these checks are failing. Do you know about it? 

We already have an issue to take a look at it https://github.com/roboflow/supervision/issues/138

> I have a feeling that `yield` is slower compare to normal video reading.

Yup. It is because you initially used `video.set` 2x. Once to set the initial position and once in the loop. I refactored it to use `video.set` only once.

SkalskiP:
There was also another bug when `end was `None`. I managed to fix it. Could you take a look at the code once again and maybe test it on your vid once again? I think everything works now, but it never hurts to have multiple people looking at the code.

SkalskiP:
Oh I don’t think it is a bug. We return different frame at the end because total frame count is not divisible by stride. If you would use stride=1 it would return the same last frame.

hardikdava:
@SkalskiP I tested the code. It seems to work fine without any issue. Should we start discussing next task?

SkalskiP:
Yeah! Let's do it tomorrow. Today I need to wrap a few things before `0.11.0` release :)

hardikdava:
Okay. Let's do it tomorrow then. Can't wait to contribute more to the community.

SkalskiP (approved):

PR #150, opened 2023-06-25
COCO Detection Dataset Import/Export support

hardikdava:
# Description

- Added a support for `COCO` detection dataset import and export
- Added a support for reading and writing `json` files
- Related issue #141 

## Type of change

-   [x] New feature (non-breaking change which adds functionality)
-   [x] This change requires a documentation update

## How has this change been tested?

- Tested locally with dataset coco128 (downloaded from roboflow)
- Import using `supervision.DetectionDataset`
- Visualized using `supervision.BoxAnnotator`
- Saved dataset using `supervision.DetectionDataset`

## Docs

-   [x] Docs updated?

PR #169, opened 2023-06-30
COCO categroy ids fix

hardikdava:
Fix for #150 issue

### DIFF
--- a/supervision/dataset/formats/coco.py
+++ b/supervision/dataset/formats/coco.py
@@ -82,8 +82,8 @@ def coco_annotations_to_detections(
         return Detections(
             class_id=np.asarray(class_ids, dtype=int), xyxy=xyxy, mask=mask
         )
-
-    return Detections(xyxy=xyxy, class_id=np.asarray(class_ids, dtype=int))
+    class_ids = np.asarray(class_ids, dtype=int)-1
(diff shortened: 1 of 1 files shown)


SkalskiP:
Hi @hardikdava 👋🏻 If I understand your code correctly, you always subtract `1` from the class id. What if the class id already starts with `0`? Or if class_id values from COCO json are no consecutive numbers? For example `0`, `1`, and `3`?

hardikdava:
Hi @SkalskiP , you are right. That can create an issue. But then what is the solution for this problem?

SkalskiP:
Hi @hardikdava 👋🏻! Here is my proposed solution: https://github.com/roboflow/supervision/pull/176. My idea is to use the same `map_detections_class_id` we used in `sv.Dataset.merge`. As a result, we will map whatever class ids exist in the COCO annotation file to a format consistent with the rest of the `Dataset` logic.

hardikdava:
@SkalskiP I will check it out and try to test it.

SkalskiP:
@hardikdava awesome! 🔥 Thanks a lot!

SkalskiP:
Closing this PR, as we developed alternative solution here: https://github.com/roboflow/supervision/pull/176

PR #170, opened 2023-07-02
redesigned annotators

SkalskiP:
This PR expands our range of Annotators and makes them more reusable.

bounding-box-annotator-example
box-corner-annotator-example
ellipse-annotator-example
label-annotator-example-2
mask-annotator-example

Issue #168, opened 2023-06-29
Advanced BoxAnnotator🤩


hardikdava:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Description

Current `sv.BoxAnnotator` is drawing boxes based on opencv rectangle. It can be improved for this kind of boxes. This looks pretty cool! 

- [ ] Advanced corner based bounding boxes
- [ ] Blur capacity (Direct object blur application)
- [ ] Boxes as mask

<img width="519" alt="Screenshot 2023-06-29 234656" src="https://github.com/roboflow/supervision/assets/39372750/d05d19d2-bac2-4797-b6f5-26c79239626e">
image source: depthai blogs

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

SkalskiP:
@hardikdava Haha I see you want to do annotators :) How about we do this first? https://github.com/roboflow/supervision/issues/49

hardikdava:
Preliminary result: 
![result](https://github.com/roboflow/supervision/assets/39372750/e55a8775-fa73-425d-b972-a3b15d55a202)

SkalskiP:
Uuuu! Looks cool! 🔥 

I drove for a few hours yesterday and thought a lot about our annotators. And I think we can use that opportunity to redesign that feature and make it even more flexible. I think we should split annotators and make them composable. What do you think @hardikdava? 

Here is my idea:

```python
import numpy as np

from abc import ABC, abstractmethod

from typing import List


# base annotator

class BaseAnnotator(ABC):

    def annotate(self, scene: np.ndarray, detections: Detections) -> np.ndarray:
        pass


class ComposableAnnotator(ABC):

    def __init__(self, annotators: List[BaseAnnotator]):
        self.annotators = annotators

    def annotate(self, scene: np.ndarray, detections: Detections) -> np.ndarray:
        annotated_image = scene
        for annotator in self.annotators:
            annotated_image = annotator.annotate(scene=scene, detections=detections)
        return annotated_image

# single responsibility annotator

class BoxAnnotator(BaseAnnotator):

    def annotate(self, scene: np.ndarray, detections: Detections) -> np.ndarray:
        pass


class PolygonMaskAnnotator(BaseAnnotator):

    def annotate(self, scene: np.ndarray, detections: Detections) -> np.ndarray:
        pass


class LabelAnnotator(BaseAnnotator):

[...]

Issue #49, opened 2023-03-22
Add object trace annotation to Supervision 🤩


SkalskiP:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

## Scope

- Create `Trace` and `TraceAnnotator` classes.
- `Trace` is responsible for calculating the history of bounding box positions within a given time (frame count) window. Every time you use the `update` method, you update the `self.trace` dictionary inside. `self.trace` stores the history of bounding box positions for every `trace_id`.
- `TraceAnnotator` draws `Trace` on frame using utilities from https://github.com/roboflow/supervision/blob/main/supervision/draw/utils.py

## API

```python
class Trace:

    def __init__(
        self,
        position: Position = Position.BOTTOM_CENTER,
        max_length: int = 10,
    ):
        self.trace: Dict[int, List[sv.Point]] = {}
        self.position = position
        self.max_length = max_length

    def update(self, detections: Detections) -> None:
        pass


class TraceAnnotator:

    def __init__(
        self,
        color: Color = Color.white(),
        thickness: int = 2
    ):
        pass

    def annotate(self, scene: np.ndarray, trace: Trace) -> np.ndarray:
        pass
```

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

hardikdava:
Hi @SkalskiP, I implemented something similar. I hope this will be helpful to you. You might need to refine it. 
```
import supervision as sv
from supervision.geometry.core import Point, Position
from typing import Dict, List, Optional, Union
from supervision.detection.core import Detections
from supervision.draw.color import Color, ColorPalette
import numpy as np
import cv2


class Trace:

    xy: np.ndarray
    class_id: Optional[np.ndarray] = None
    tracker_id: Optional[np.ndarray] = None

    def __init__(
        self,
        position: Position = Position.CENTER,
        max_length: int = 10,
    ):
        self.trace: Dict[int, List[sv.Point]] = {}
        self.position = position
        self.max_length = max_length
        self.frame_counter = 0
        # Counter, x, y, class, trackid
        self.storage = np.zeros((0, 5))

    def update(self, frame_counter: int, detections: Detections) -> None:
        if detections.xyxy.shape[0]>0:
            xyxy = detections.xyxy
            x = (xyxy[:, 0] + xyxy[:, 2]) / 2
            y = (xyxy[:, 1] + xyxy[:, 3]) / 2
[...]

Issue #16, opened 2023-02-07
Draw ellipse from bounding box


litaldavar:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Question

How do I draw ellipse from the bounding box from BoxAnnotator class with YOLOv8?

SkalskiP:
Hi, @litaldavar 👋🏻! Thanks a lot for your interest in Supervision. We don't have that feature yet, unfortunately. We talk about ellipses like this: https://youtu.be/QCG8QMhga9k

hardikdava:
@SkalskiP @litaldavar  Do we have any confirmation on final design about this issue?

SkalskiP:
Hi @hardikdava 👋🏻 I would go for a design like the one you can see in the [video](https://www.youtube.com/watch?v=QCG8QMhga9k). Here is the [notebook](https://colab.research.google.com/github/roboflow-ai/notebooks/blob/main/notebooks/how-to-track-football-players.ipynb) I used to process the video.

hardikdava:
Hey @SkalskiP I will take a look and try to include in our feature.

hardikdava:
Preview:
User can play with start and end angles to customize more.
![Screenshot from 2023-07-07 19-12-46](https://github.com/roboflow/supervision/assets/39372750/8adf7aa3-4f45-4d96-b27b-4ff740300204)

PR #171, opened 2023-07-03
cornered box annotator

hardikdava:
# Description

As discussed in #168  and #170 . Merging new annotation format for initial commit.

@SkalskiP it might needed to refine a bit.

## Type of change


hardikdava:
@SkalskiP Please merge this branch before further work. I already finished some implementation.

hardikdava:
@SkalskiP #49 is also included in this PR.

hardikdava:
@SkalskiP I changed code style to consistent style as `Annotators.annotate(scene, detection)` format. Please review it and merge the PR.

hardikdava:
@SkalskiP `EllispseAnotator` is also added to the `Anotator` API. #16

SkalskiP (approved):

supervision/detection/track.py
```diff
@@ -0,0 +1,68 @@
...
+        self.position = position
+        self.max_length = max_length
+        self.frame_counter = 0
+        self.storage = np.zeros((0, 5))
+
+    def update(self, frame_counter: int, detections: Detections) -> None:
+        if detections.xyxy.shape[0] > 0:
+            xyxy = detections.xyxy
+
+            if self.position == Position.CENTER:
```
SkalskiP, line 43: That part of the logic is already implemented in `sv.Detections.get_anchor_coordinates`. Let's refactor our logic here to use the `get_anchor_coordinates` method. Also:

- Let's expand `Position` with:
    -  `TOP_CENTER = "TOP_CENTER"`
    -  `TOP_LEFT = "TOP_LEFT"`
    -  `TOP_RIGHT = "TOP_RIGHT"`
    -  `CENTER_LEFT = "CENTER_LEFT"`
    -  `CENTER_RIGHT = "CENTER_RIGHT"`
    -  `BOTTOM_LEFT = "BOTTOM_LEFT"`
    -  `BOTTOM_RIGHT = "BOTTOM_RIGHT"`

- Let's expand the `get_anchor_coordinates` implementation to handle new `Position`'s
- We need to add unit tests for `get_anchor_coordinates`, but I can handle that. Just let me know when you'll be ready.

supervision/detection/track.py
```diff
@@ -0,0 +1,68 @@
...
+            >>> for frame in sv.get_video_frames_generator(source_path='source_video.mp4'):
+            >>>     detections = sv.Detections(...)
+            >>>     tracked_objects = tracker(...)
+            >>>     tracked_detections = sv.Detections(tracked_objects)
+            >>>     track_storage.update(tracked_detections)
+        """
+        self.position = position
+        self.max_length = max_length
+        self.frame_counter = 0
+        self.storage = np.zeros((0, 5))
```
SkalskiP, line 37: Curious why you decided to use `np.ndarray` as a storage engine.

SkalskiP (changes requested):

SkalskiP:
@hardikdava I left you few comments already. I'm taking a look at the rest of the code, as well as testing it. But there are a few things that I need to think about tomorrow. Please start implementing the changes I already listed. 🙏🏻

supervision/detection/track.py
hardikdava: @SkalskiP , I made changes and added extra anchor points.

supervision/detection/track.py
hardikdava: > Curious why you decided to use `np.ndarray` as a storage engine.

Hi @SkalskiP, I am trying to rewrite `TrackerStorage` in the format of `Detections`. I will let you know once I am finished.

hardikdava: @SkalskiP I am done with consistent format for `TrackerStorage`. You can review it now.

supervision/annotators/core.py
```diff
@@ -74,20 +173,604 @@ def annotate(self, scene: np.ndarray, detections: Detections) -> np.ndarray:
...
-        pass
+    def __init__(
+        self,
+        color: Union[Color, ColorPalette] = ColorPalette.default(),
+        thickness: int = 2,
+        text_color: Color = Color.black(),
+        text_scale: float = 0.5,
+        text_thickness: int = 1,
+        text_padding: int = 10,
+        color_by_track: bool = False,
```
SkalskiP, line 191: Let's add `label_formatter: Callable[[Detections], List[str]]` argument.

The default `label_formatter` could be something like this.

```python
def default_label_formatter(detections: Detections) -> List[str]:
    return [
        str(class_id) 
        for class_id 
        in detections.class_id
    ]
```

If you want to use class mapping in your `label_formatter` you could do it like this:

```python
def build_label_formatter(classes: List[str]) -> Callable[[Detections], List[str]]:
    def default_label_formatter(detections: Detections) -> List[str]:
        return [
            classes[class_id] 
            for class_id 
            in detections.class_id
        ]
    return default_label_formatter
```

SkalskiP (changes requested):

supervision/annotators/core.py
hardikdava: Hi @SkalskiP :wave: , Thanks for the idea. I think it is too much for just to create `label`. This type of callback is not a good idea. It makes complicated everything. Instead, what do you think of this:

```python

class LabelAnnotator(BaseAnnotator):
	def __init__(..., labelmaps: Optionl[Dict[int, str]]=None):
		self.labelmaps = labelmap
		
	def annotate(self,  scene: np.ndarray,  detections: Detections):
	    for det in detections:
			if self.labelmaps:
		       	label=self.labelmaps[classid]
		    if not self.labelmaps:
			    labels = f"{classid}"				  

```
An example of labelmaps would be like:

```
labelmaps = {0: 'person', 1: '...', 2: '...'}
```
My concern is that if user has to write callback function then mostly they are not going to use this from `supervision` but instead they will use their own.

SkalskiP: Hi @hardikdava 👋🏻, label map is, of course, a lot simpler, but a mall problem I see is not flexible :/ It doesn't allow you to build complicated labels.

How about adding both `classes: Optional[List[str]], label_formatter: Optional[Callable[[Detections], List[str]]]`?

hardikdava: @SkalskiP, That seems fine to me. User can have flexibility more when they want to use some other format.

SkalskiP: @hardikdava exactly! There will likely be multiple people asking about label customization. And I want to support them as well :)

hardikdava: It is done. Ready for test.

SkalskiP: Awesome! 🔥 Let me test it. We will probably need to add unit tests for `get_anchor_coordinates`. Would you like me to take care of that?

hardikdava: Okay. I will try. What is command line for running tests? basic pytest or something special?

hardikdava:
@SkalskiP I have added tests for `get_anchor_coordinates`

Issue #215, opened 2023-07-21
ConfusionMatrix error when no targets but predictions available


hardikdava:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar bug report.

### Bug

This is specific case when there is only background image but on objects are inside. Due to low accuracy of model, model is predicting `False Positives` on background image. I simulate this by  just removing `targets` from one of the image labels. 

Note: If you are using yolo format then open **first** `labels.txt` file and delete all the labels from it.

This action will raise following error:

```
    raise ValueError(
ValueError: Targets must have shape (N, 5). Got (0, 6) instead.
```

### Minimal Reproducible Example

```python

import supervision as sv
from ultralytics import YOLO


ds = sv.DetectionDataset.from_yolo(images_directory_path="../data/coco128/images/train2017",
                                   annotations_directory_path="../data/coco128/labels/train2017",
                                   data_yaml_path="../data/coco128.yaml")


model = YOLO(model="yolov8n.pt")


predictions = []
gts = []

for image_name, image in ds.images.items():
    yolo_det = model.predict([image], imgsz=320, verbose=True)[0]
    det = sv.Detections.from_yolov8(yolov8_results=yolo_det)

    predictions.append(det)
    gts.append(ds.annotations[image_name])


print(gts[0])
print(predictions[0])
mat = sv.ConfusionMatrix.from_detections(classes=ds.classes, conf_threshold=0.1, iou_threshold=0.5, predictions=predictions, targets=gts)

print(mat.matrix)
# mat.plot(save_path="mattrix.png", title="coco128-yolov8n")
```

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

kirilllzaitsev:
Hey @hardikdava. Could you try changes in https://github.com/roboflow/supervision/pull/216 and confirm that the issue is fixed? Using your code on my side, having set all targets to be empty:

```python
targets=[
        sv.Detections(xyxy=np.empty((0, 4)), class_id=np.empty((0)))
        for _ in range(len(predictions))
    ]
```

and subsequently running sv.ConfusionMatrix.from_detections, I get:
```python
np.sum(mat.matrix[-1]) == np.sum(mat.matrix)
True
```

Issue #212, opened 2023-07-21
ConfusionMatrix - plot bug


hardikdava:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar bug report.

### Bug

ConfusionMatrixAPI is haveing following error

```python
TypeError: unsupported operand type(s) for +: 'dict' and 'list'
```

### Minimal Reproducible Example

```python

import supervision as sv
from ultralytics import YOLO

ds = sv.DetectionDataset.from_yolo(images_directory_path="../data/coco128/images/train2017",
                                   annotations_directory_path="../data/coco128/labels/train2017",
                                   data_yaml_path="../data/coco128.yaml")

model = YOLO(model="yolov8n.pt")

predictions = []
gts = []

for image_name, image in ds.images.items():
    yolo_det = model.predict([image], imgsz=320, verbose=True)[0]
    det = sv.Detections.from_yolov8(yolov8_results=yolo_det)

    predictions.append(det)
    gts.append(ds.annotations[image_name])

mat = sv.ConfusionMatrix.from_detections(classes=ds.classes, conf_threshold=0.1, iou_threshold=0.5, predictions=predictions, targets=gts)

mat.plot(save_path="mattrix.png", title="coco128-yolov8n")

```

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

kirilllzaitsev:
@hardikdava , thanks for finding it. This PR should address it https://github.com/roboflow/supervision/pull/216

hardikdava:
@SkalskiP, `sv.ConfusionMatrix.convert_detections_to_tensor` should be a `staticmethod`.

SkalskiP:
@hardikdava agree! Let me make it static in #223 PR.

SkalskiP:
Solved with https://github.com/roboflow/supervision/pull/223

PR #216, opened 2023-07-21
Fixes to ConfusionMatrix

kirilllzaitsev:
# Description

- Fix ConfusionMatrix.plot() when classes aren't a list. Addresses https://github.com/roboflow/supervision/issues/212
- Fix input validation to account for empty targets/predictions. Addresses https://github.com/roboflow/supervision/issues/215

## Type of change

-   [+] Bug fix (non-breaking change which fixes an issue)

Reproduced the issue.

### DIFF
--- a/supervision/metrics/detection.py
+++ b/supervision/metrics/detection.py
@@ -216,20 +216,28 @@ class ConfusionMatrix:
                 f"Number of predictions ({len(predictions)}) and targets ({len(targets)}) must be equal."
             )
         if len(predictions) > 0:
-            if not isinstance(predictions[0], np.ndarray) or not isinstance(
-                targets[0], np.ndarray
-            ):
-                raise ValueError(
-                    f"Predictions and targets must be lists of numpy arrays. Got {type(predictions[0])} and {type(targets[0])} instead."
-                )
-            if predictions[0].shape[1] != 6:
-                raise ValueError(
-                    f"Predictions must have shape (N, 6). Got {predictions[0].shape} instead."
-                )
-            if targets[0].shape[1] != 5:
-                raise ValueError(
-                    f"Targets must have shape (N, 5). Got {targets[0].shape} instead."
-                )
+            first_non_empty_prediction = next(
+                (x for x in predictions if len(x) > 0), None
+            )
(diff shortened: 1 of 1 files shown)


supervision/metrics/detection.py
```diff
@@ -216,20 +216,28 @@ def _validate_input_tensors(
...
-                )
-            if predictions[0].shape[1] != 6:
-                raise ValueError(
-                    f"Predictions must have shape (N, 6). Got {predictions[0].shape} instead."
-                )
-            if targets[0].shape[1] != 5:
-                raise ValueError(
-                    f"Targets must have shape (N, 5). Got {targets[0].shape} instead."
-                )
+            first_non_empty_prediction = next(
```
SkalskiP, line 219: @kirilllzaitsev, why do we need to look for the first non-empty?

hardikdava: Nope,  I found the real issue.

hardikdava (changes requested):
here, if `convert_to_tensor` function is agnostic whether the `detections` or `target`. If it is `target` then `confidences` should not be considered as it will always be `1` and does not influence in result. But `_validation` function is checking fo r shape for `(N,5)` and after adding `confidence` makes shape to `(N,6)`.

SkalskiP:
@hardikdava, your explanation looks reasonable. I'll be working on an alternative fix.

SkalskiP:
@kirilllzaitsev and @hardikdava [here](https://github.com/roboflow/supervision/pull/217) is alternative fix for [#215](https://github.com/roboflow/supervision/issues/215) take a look

supervision/metrics/detection.py
kirilllzaitsev: @SkalskiP, prediction/target can be either empty or non-empty. Checking the first non-empty instance and assuming that all subsequent non-empty instances will have the same shape, I can raise ValueError if the shape is wrong.

SkalskiP:
@kirilllzaitsev and @hardikdava, I'm doing small housekeeping. I'll close this PR. https://github.com/roboflow/supervision/issues/215 is already fixed. I will submit https://github.com/roboflow/supervision/issues/212 fix later today.

PR #219, opened 2023-07-22
fix: 🐛 package version control if PackageNotFoundError occured

### LINKED ISSUE #213: Developement - version issue
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar bug report.

### Bug

I have not installed `supervision` but running directly for developement purpose.

But I got an following error when I import supervision code:

```
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: supervision
```

After a quick investigation, it is found that `__init__.py` of `supervision` where version information used. It is creating an issue. If I comment this line and the bug is gone.

@onuralpszr Can you take a look?

I think it should be ignorable, if valid version is not found then use `development` version. Though, I do not have concret idea, how to tackle it.

### Environment

_No response_

_No response_

### Additional

_No response_

onuralpszr:
# Description

Fix #213 for no install method (poetry or pip install .  "install package" so we can find version If none happen __version__ check will fail, to prevent that changed in to "development" tag for set to version.

### DIFF
--- a/supervision/__init__.py
+++ b/supervision/__init__.py
@@ -1,6 +1,10 @@
 import importlib.metadata as importlib_metadata
 
-__version__ = importlib_metadata.version(__package__)
+try:
+    # This will read version from pyproject.toml
+    __version__ = importlib_metadata.version(__package__ or __name__)
+except importlib_metadata.PackageNotFoundError:
+    __version__ = "development"
 
 
 from supervision.classification.core import Classifications


onuralpszr:
@SkalskiP @hardikdava  please check 

dev package installed version check (pip install .)

```bash
pip list

supervision            0.12.0
```

```python
>>> import supervision
>>> supervision.__version__
'0.12.0'
>>> exit()
```

Poetry install results

pip list

supervision                0.12.0   /Users/onuralp/Github/supervision (locally installed which means better for development)

```python
>>> import supervision
>>> supervision.__version__
'0.12.0'

```

Technically using poetry will give you "local install" so every change you made you will already able to use without re-install.

onuralpszr:
@SkalskiP using poetry actually better for development as well. (less headache) :)

hardikdava:
@onuralpszr @SkalskiP Should we update documentation for installation if we also support for poetry?

onuralpszr:
> @onuralpszr @SkalskiP Should we update documentation for installation if we also support for poetry?

we already did

SkalskiP:
@hardikdava take a look at this: https://roboflow.github.io/supervision/. We updated the installation guide yesterday. It is 🔥!

SkalskiP:
@onuralpszr merging! Thanks a lot 🙏🏻

PR #223, opened 2023-07-22
🛠️ #212 bug fix

SkalskiP:
# Description

It is the fix for #212. Turns out that the problem is not with the confusion matrix but with the loading of the YOLO dataset. I thought that `names` value in `data.yaml` can only be `List[str]` but it can also be `Dict[int, str]`. I added some logic to handle that.

- Tested in Google Colab

## Docs


SkalskiP:
@hardikdava please take a look. I tested the fix with the code you provided and it works.

Issue #205, opened 2023-07-20
coco segmentation data for polygon should be [polygon,] not polygon


yance-dev:
thanks for this great project, but I found that supervision export coco as :

```
"segmentation":  polygon,
```
https://github.com/roboflow/supervision/blob/20954a299660a1f227a772a0a7cd253761ddb9da/supervision/dataset/formats/coco.py#L127

but coco official [data format](https://cocodataset.org/#format-data:~:text=%22segmentation%22,RLE%20or%20%5Bpolygon%5D%2C) is :
```
"segmentation":  [polygon],
```
even only one polygon, it should be a list of polygons.

the third party like fifityone may not support this format.

that's all, thanks again.

SkalskiP:
Hi, @yance-dev 👋🏻! Thanks a lot for your interest in `supervision`. Let me take a look and confirm if the problem exists. 

Would you like to create a PR with fix?

SkalskiP:
Hi, @yance-dev. I just did some tests and it looks like everything work just fine on our side. Did you face any issues?

onuralpszr:
It does work because polygon is a "list" checkout here as well

https://github.com/roboflow/supervision/blob/20954a299660a1f227a772a0a7cd253761ddb9da/supervision/dataset/formats/coco.py#L111C12-L111C12

SkalskiP:
It looks like a false alarm this time. @yance-dev I'm closing the issue for now, but feel free to reopen it in the future.

yance-dev:
> It does work because polygon is already a "list" checkout here as well
> 
> https://github.com/roboflow/supervision/blob/20954a299660a1f227a772a0a7cd253761ddb9da/supervision/dataset/formats/coco.py#L111
> 
> https://github.com/roboflow/supervision/blob/20954a299660a1f227a772a0a7cd253761ddb9da/supervision/dataset/formats/coco.py#L113

polygon is a list. but coco `"segmentation":` should be  polygons which is dyadic array like `[[x1,y1,x2,y2,x3,y3,x4,y4],[x1,y1,x2,y2,x3,y3,x4,y4]]`

https://github.com/roboflow/supervision/blob/20954a299660a1f227a772a0a7cd253761ddb9da/supervision/dataset/formats/coco.py#L119C31-L119C31
 ```
                approximate_mask_with_polygons(
                    mask=mask,
                    min_image_area_percentage=min_image_area_percentage,
                    max_image_area_percentage=max_image_area_percentage,
[...]

PR #225, opened 2023-07-23
Format coco output data as per standard

mayankagarwals:
# Description

Fixes https://github.com/roboflow/supervision/issues/205

The issue was compatible with supervision because even though supervision exports data in the wrong format, its function to read coco dataset works for both the correct and malformed output format 
Refer: https://github.com/roboflow/supervision/blob/20954a299660a1f227a772a0a7cd253761ddb9da/supervision/dataset/formats/coco.py#L87
```
        polygons = [
            np.reshape(
                np.asarray(image_annotation["segmentation"], dtype=np.int32), (-1, 2)
            )
            for image_annotation in image_annotations
        ]
```
Regardless of the format of `image_annotation["segmentation"]`, data is loaded in expected shape 

With this fix, data stored will be in expected format. 

## Type of change

While I have not tested it by loading exported format in a third party tool, I have gone through the exported annotations file and validated the format. 
If we have a specific third party tool/ coco format validator in mind, we can make sure the updated file works well with it. 
Gist of exported json: https://gist.github.com/mayankagarwals/3fb243c7a776ee5b8986386aef60b76d

## Docs

### DIFF
--- a/supervision/dataset/formats/coco.py
+++ b/supervision/dataset/formats/coco.py
@@ -124,7 +124,7 @@ def detections_to_coco_annotations(
             "category_id": int(class_id),
             "bbox": [xyxy[0], xyxy[1], box_width, box_height],
             "area": box_width * box_height,
-            "segmentation": polygon,
+            "segmentation": [polygon],
             "iscrowd": 0,
         }
         coco_annotations.append(coco_annotation)
--- a/supervision/dataset/formats/coco.py
+++ b/supervision/dataset/formats/coco.py
@@ -124,7 +124,7 @@ def detections_to_coco_annotations(
             "category_id": int(class_id),
             "bbox": [xyxy[0], xyxy[1], box_width, box_height],
             "area": box_width * box_height,
-            "segmentation": [polygon],
+            "segmentation": [polygon] if polygon else [],
             "iscrowd": 0,
         }
         coco_annotations.append(coco_annotation)


SkalskiP:
Hi, @mayankagarwals 👋🏻! Awesome. Could you create Colab that would demonstrate dataset before and after are the same? It would make testing a lot simpler. You can use this one as starting point: https://colab.research.google.com/drive/1E2H8MkgFjMeJznOLUc6npaAQuFkRRz7-?usp=sharing

supervision/dataset/formats/coco.py
```diff
@@ -124,7 +124,7 @@ def detections_to_coco_annotations(
             "category_id": int(class_id),
             "bbox": [xyxy[0], xyxy[1], box_width, box_height],
             "area": box_width * box_height,
-            "segmentation": polygon,
+            "segmentation": [polygon],
```
yance-dev, line 127: if there is no polygon,segmentation will be `[[]]`. 

`[[]]` in python if condition will be `True`.

`[]` in python if condition will be `False`.


segmentation in coco dataset is `[]`  when there is no segmentatio which means `False` in python.Although it is a small change, we should be as cautious as possible considering that someone may make a Boolean judgment on it.

change `[polygon]` to `[polygon] if polygon else []` may avoid this problem.

mayankagarwals: Makes a lot of sense. Apart from truth value, it could also break anything using the size of the array.  Have added to the fix

mayankagarwals:
> Hi, @mayankagarwals 👋🏻! Awesome. Could you create Colab that would demonstrate dataset before and after are the same? It would make testing a lot simpler. You can use this one as starting point: https://colab.research.google.com/drive/1E2H8MkgFjMeJznOLUc6npaAQuFkRRz7-?usp=sharing

Would love to @SkalskiP. I'm able to write this test locally. Any documentation about what the process followed by supervision is for snapshot releases? I saw in makefile you are using Testpypi. I'll figure out a way if one doesn't exist already

SkalskiP:
@mayankagarwals, you can install your version of the `supervision` package directly from git. You can do it like that: `pip install git+https://github.com/username/repository.git@branch_name
`.

mayankagarwals:
Oh yeah, that makes sense. 

https://colab.research.google.com/drive/1E2H8MkgFjMeJznOLUc6npaAQuFkRRz7-?usp=sharing
This notebook should serve as a good enough test. I've raised a few exceptions where the library fails without the fix.

yance-dev:
> Oh yeah, that makes sense.
> 
> [colab.research.google.com/drive/1E2H8MkgFjMeJznOLUc6npaAQuFkRRz7-?usp=sharing](https://colab.research.google.com/drive/1E2H8MkgFjMeJznOLUc6npaAQuFkRRz7-?usp=sharing) This notebook should serve as a good enough test. I've raised a few exceptions where the library fails without the fix.
> 
> @yance-dev Please have a look too!

🎉 Great job, everything seems to be working fine now

SkalskiP:
@mayankagarwals, let me do some tests as well.

SkalskiP:
@mayankagarwals I'm not sure if I'm using it right, but I see `Exception: WRONG DATA FORMAT`

Screenshot 2023-07-24 at 12 33 08

mayankagarwals:
Hey @SkalskiP !, 

🧐 My hunch is Colab preserves the environment on restarting runtimes. I've added a small piece to uninstall existing versions before installing from git. That should fix it :). Please give it one more shot

SkalskiP:
@mayankagarwals looks like you were right. I've done a few more tests, and it works. Merging.

PR #226, opened 2023-07-23
Fix yolo dataset background instance loading

hardikdava:
# Description

## Type of change

Load `coco128` dataset which contains images but not annotations. This represents background image. `Supervision` is eliminating such cases.


SkalskiP:
@hardikdava, let me know when you'll be ready for review here.

hardikdava:
@SkalskiP the fix is added. Please review it and merge it.

SkalskiP:
@hardikdava I'm testing now. If I understand correctly, before that change, `coco128.yaml` should load only 126 entries, and after, we should have 128.

hardikdava:
> @hardikdava I'm testing now. If I understand correctly, before that change, `coco128.yaml` should load only 126 entries, and after, we should have 128.

Yeah, 128 is expected result.

SkalskiP:
@hardikdava tested! Merging.

Issue #183, opened 2023-07-06
Show Progress in time consuming tasks


hardikdava:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Description

It is highly suggestable that users have information on how process is going on especially on time consuming tasks such as dataset operations.

Example:
Loading a large dataset can be time consuming. It can take a while to load whole dataset. Meanwhile users might be wondering if it is working properly or not. In worst situaiton, the process might even fails due to circumstances like memory issue or dataset issue.

### Use case

- Dataset ops
- Image saving ops
- Video processing ops

### Additional

Solution:

Introducing Progressbar from `tqdm` can be very useful in such scenario. 

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

hardikdava:
@SkalskiP This can be wait for later.

SkalskiP:
Hi @hardikdava 👋🏻 I was also thinking about it. There are two things we need to consider:
- The progress bar needs to work in the terminal and notebook environment.
- The progress bar needs to be optional. You should be able to turn that off.

Let's keep that issue open. We will pick it up in the future 100%.

onuralpszr:
@SkalskiP  

Tqdm could be a solution for this 

https://github.com/tqdm/tqdm (works with terminal/notebooks) you see in their docs as well.

SkalskiP:
@onuralpszr oh I'm sure we will use `tqdm` to implement it. Awesome package.

mayankagarwals:
@hardikdava Let me know if you've already started working on this! Would be fun to pick this up

SkalskiP:
Hi, @mayankagarwals 👋🏻! As far as I know, no one started to work on this issue yet. Feel free to try :)

hardikdava:
@mayankagarwals Yeah, I haven't started working on this. Feel free to try it out :)

mayankagarwals:
Amazing. I think this problem will be solved differently for each of 
```
1. Dataset ops
2. Image saving ops
3. Video processing ops
```

based on the process taking the most time in each and how homogeneously they can be divided to track progress. 

For `Dataset ops`, I ran a small test on the function to load coco dataset 
```
ds = sv.DetectionDataset.from_coco(
    images_directory_path="/Users/mayankagarwal/Documents/Personal/CS/codebase/Supervision-Parent/coco-segmentation/train",
[...]

PR #234, opened 2023-07-24
Add tqdm support

mayankagarwals:
# Description

## Type of change

YOUR_ANSWER

## Docs

### DIFF
--- a/supervision/dataset/formats/coco.py
+++ b/supervision/dataset/formats/coco.py
@@ -2,6 +2,7 @@ import os
 from datetime import datetime
 from pathlib import Path
 from typing import Dict, List, Tuple
+from tqdm import tqdm
 
 import cv2
 import numpy as np
@@ -13,6 +14,7 @@ from supervision.dataset.utils import (
 from supervision.detection.core import Detections
 from supervision.detection.utils import polygon_to_mask
 from supervision.utils.file import read_json_file, save_json_file
+from supervision.utils.console import get_dataset_loading_tqdm_iterable
 
 
 def coco_categories_to_classes(coco_categories: List[dict]) -> List[str]:
@@ -150,7 +152,7 @@ def load_coco_annotations(
     images = {}
     annotations = {}
 
-    for coco_image in coco_images:
+    for coco_image in get_dataset_loading_tqdm_iterable(coco_images):
         image_name, image_width, image_height = (
             coco_image["file_name"],
             coco_image["width"],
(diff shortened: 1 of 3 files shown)


supervision/dataset/formats/coco.py
```diff
@@ -150,7 +152,7 @@ def load_coco_annotations(
     images = {}
     annotations = {}
 
-    for coco_image in coco_images:
+    for coco_image in get_dataset_loading_tqdm_iterable(coco_images):
```
artyaltanzaya, line 155: We could directly use tqdm without separate function like this: 

`for coco_image in tqdm(coco_images, desc="Loading Dataset: "):
`

mayankagarwals: This is done keeping extensibility in mind. The function is defined in separate file so that it's reusable for other dataset loading :)

SkalskiP: hi @mayankagarwals 👋🏻 I'm with @artyaltanzaya. It is awesome that you think about reusability, but like everything in programming, it is all about balance. We want to extract a single line of code into a separate function.

SkalskiP (changes requested):

supervision/dataset/formats/coco.py
mayankagarwals: > it is all about balance

That I agree with. My rationale is we will be duplicating this 

```
tqdm(coco_images, desc="Loading Dataset: ")
```
across coco, pascal voc and any other dataset we support/plan to support. 
Even if we assume tqdm is stable and won't make any breaking changes, for changing the description or adding any other argument, it needs to be repeated per dataset

To add to this, we will have tqdm imports in every file we plan on introducing progress tasks to  : Image/Video saving/processing ops, Dataset saving/processing ops, etc.

mayankagarwals: Current idea is to expose functions like 

```
def get_dataset_loading_tqdm_iterable
def get_dataset_saving_tqdm_iterable
```

You will see that get_dataset_loading_tqdm_iterable is exposed in a util file in this pull request

Which will be reused for these specific use cases. The idea is that we will need similar tqdm instances for specific tasks like dataset loading. 

Again, this is a new repository for me, so I can't strongly comment on what's the best way. Both will work. More than happy to remove the function for now :)

mayankagarwals: For example: 

Once we are done this PR, we will want to introduce the same for pascal voc. The change will be 

```python 
    for voc_image in get_dataset_loading_tqdm_iterable(voc_images):
```

instead of 

```python
    for voc_image in tqdm(voc_images, desc="Loading Dataset: "): 

```

mayankagarwals: Anyway, we are pausing work on this (refer : https://github.com/roboflow/supervision/issues/183)

PR #237, opened 2023-07-25
docs: 📝 MIT license classifier added

onuralpszr:
# Description

Toml file License corrected and license section also added into toml file as well. An oversight I miss in toml file


SkalskiP:
Hi @onuralpszr 👋🏻! Yesterday, I noticed that as well.

PR #256, opened 2023-08-01
Tracker latest

dbroboflow:
# Description

Refer to https://github.com/roboflow/supervision/pull/203

## Type of change

Updated tracker version

PR #258, opened 2023-08-01
tracker intermediate version

hardikdava:
This pr is intermediate version of #256 

- object class id is not a integrated in #256  but additional step was done from tracker result.
- additional step is removed
- typing added and fixed
- vectorized `box_iou` is integrated.


hardikdava:
@SkalskiP @dbroboflow I updated code for tracker, it is faster than before and fixed few things. Please merge it and work on #256  from there.

Issue #244, opened 2023-07-25
Add support for loading segmentation datasets in Pascal VOC format


kirilllzaitsev:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Question

@SkalskiP , hi, I have some questions regarding loading PASCAL VOC segmentation from XML.

1. A general one. I'm wondering why there are DetectionDataset, ClassificationDataset, but no SegmentationDataset that logically follows the sequence. Instead, DetectionDataset can handle both object detection and segmentation? Here, I refer to `DetectionDataset.from_yolo` being able to parse object detection and segmentation samples.

2. [This function](https://github.com/roboflow/supervision/blob/36cdcf1b128e38cb9a086290fbfb0130fa64ac5b/supervision/dataset/formats/pascal_voc.py#L122) loads only object detection data, so our task for this PR is to make it follow a convention for YOLO or [COCO](https://github.com/roboflow/supervision/blob/36cdcf1b128e38cb9a086290fbfb0130fa64ac5b/supervision/dataset/formats/coco.py#L135), keeping the backward compatibility?

mayankagarwals:
> [This function](https://github.com/roboflow/supervision/blob/36cdcf1b128e38cb9a086290fbfb0130fa64ac5b/supervision/dataset/formats/pascal_voc.py#L122) loads only object detection data, so our task for this PR is to make it follow a convention for YOLO or [COCO](https://github.com/roboflow/supervision/blob/36cdcf1b128e38cb9a086290fbfb0130fa64ac5b/supervision/dataset/formats/coco.py#L135), keeping the backward compatibility?


@kirilllzaitsev I went into this rabbit hole recently. Here are few things I observed: 
a. As you mentioned [coco](https://github.com/roboflow/supervision/blob/36cdcf1b128e38cb9a086290fbfb0130fa64ac5b/supervision/dataset/formats/coco.py#L72) and yolo both provide polygon loading support while our voc does not 
b. VOC dataset doesn't provide segmentation in annotations but just provides png images with the segmentation (It provides two folders `SegmentationClass` and `SegmentationObject`

SkalskiP:
@kirilllzaitsev This is an excellent question. Our segmentation API needs to be more intuitive. Problems that I see: 

- Lack of consistency. We have classification and detection datasets, but we do not have a segmentation dataset. Our detection dataset is doing both detection and segmentation. 
[...]

PR #263, opened 2023-08-02
Decouple mask and bndbox Detections (for VOC)

kirilllzaitsev:
# Description

Addresses [this issue](https://github.com/roboflow/supervision/issues/244). In the context of the VOC format, Detections are allowed to have either `xyxy` or `mask` field non-empty. Previous behavior to handle both `xyxy` and `mask` is supported, but a deprecation warning is issued.

The way to dump a dataset to disk is changed to a separate `XMLBuilder` class.

## Type of change

-   [+] This change requires a documentation update

Tested with either of (`xyxy`, `mask`), (`xyxy`), (`mask`) available.

## Docs

### DIFF
--- a/supervision/dataset/formats/pascal_voc.py
+++ b/supervision/dataset/formats/pascal_voc.py
@@ -72,7 +72,7 @@ class XMLBuilder:
     def to_string(self) -> str:
         return parseString(tostring(self.annotation)).toprettyxml(indent="  ")
 
-    def add_obj_det(self, root, xyxy: np.ndarray) -> Element:
+    def add_obj_det(self, root: Element, xyxy: np.ndarray) -> Element:
         bndbox = SubElement(root, "bndbox")
         xmin = SubElement(bndbox, "xmin")
         xmin.text = str(int(xyxy[0]))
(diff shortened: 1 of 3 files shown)


kirilllzaitsev:
@SkalskiP here is a starter patch of changes on the transition path to decoupled object detection and segmentation. Taking VOC format as a playground:
- logic for dumping a a Detections object is encapsulated in `XMLBuilder`
- having both `xyxy` and `mask` issues a deprecation message
- 
If the direction with `xyxy` and `mask` looks good, I will create a `SegmentationDataset`, further simplifying [`detections_to_pascal_voc`](https://github.com/kirilllzaitsev/supervision/blob/6fbeb7aeaa0b8f7b4cedec803610a0c53f28ac63/supervision/dataset/formats/pascal_voc.py#L104C5-L104C29). Otherwise, if the segmentation-related matters are too much for now, at least `XMLBuilder` looks like an enhancement.

SkalskiP:
@capjamesg @yeldarby I think you should know we are working on those changes. They might influence Autodistill.

PR #273, opened 2023-08-07
ultralytics tracker support added

hardikdava:
# Description

This pr enables directly convert tracker from ultralytics to `sv.Detections`


SkalskiP:
@hardikdava is it tested? We could squeeze that into `supervision-0.13.0` if it is ready and tested.

hardikdava:
Yeah, it is tested.

Issue #253, opened 2023-07-31
Drop `sv.Detections.from_yolov8` support


SkalskiP:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Description

https://github.com/roboflow/supervision/pull/222 introduces general support for all models offered by `ultralytics` package. We added `sv.Detections.from_ultralytics` method, and that makes `sv.Detections.from_yolov8` redundant and deprecated. We will drop `sv.Detections.from_yolov8` support with `0.15.0`.

capjamesg:
@SkalskiP Is there a reason to drop the deprecation notice so soon and remove the code entirely? We use the YOLOv5 and YOLOv8 data loaders in a few places on the Roboflow site that we'll need to update.

SkalskiP:
Hi, @capjamesg 👋🏻! Is `supervision-0.15.0` soon? Given the frequency of our releases, I think this task will go into effect no sooner than in 1.5 months.

1. YOLOv5 support stays as it is.
2. YOLOv8 support is getting migrated. All you must do is change `from_yolov8` into `from_ultralytics`. No other changes should be required.

PR #278, opened 2023-08-10
Remove yolo v8 support

mayankagarwals:
Fixes https://github.com/roboflow/supervision/issues/253

### DIFF
--- a/supervision/detection/core.py
+++ b/supervision/detection/core.py
@@ -170,38 +170,6 @@ class Detections:
             class_id=yolov5_detections_predictions[:, 5].astype(int),
         )
 
-    @classmethod
-    @deprecated(
-        "This method is deprecated and removed in 0.15.0 release. Use sv.Detections.from_ultralytics() instead."
-    )
-    def from_yolov8(cls, yolov8_results) -> Detections:
-        """
-        Creates a Detections instance from a [YOLOv8](https://github.com/ultralytics/ultralytics) inference result.
-
-        Args:
-            yolov8_results (ultralytics.yolo.engine.results.Results): The output Results instance from YOLOv8
-
-        Returns:
-            Detections: A new Detections object.
-
-        Example:
-            ```python
-            >>> import cv2
-            >>> from ultralytics import YOLO
-            >>> import supervision as sv
-
-            >>> image = cv2.imread(SOURCE_IMAGE_PATH)
-            >>> model = YOLO('yolov8s.pt')
-            >>> result = model(image)[0]
-            >>> detections = sv.Detections.from_yolov8(result)
-            ```
-        """
-        return cls(
-            xyxy=yolov8_results.boxes.xyxy.cpu().numpy(),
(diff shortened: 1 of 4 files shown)


mayankagarwals:
CC @hardikdava

SkalskiP:
Hi @mayankagarwals 👋🏻! A few overall comments:
- `from_yolov8` will be dropped entirely in `supervision-0.15.0`, so in 2 releases. Let's wait to remove the `from_yolov8` function. If you want to get that PR to merge now, bring back `from_yolov8`.
- We can already prepare to drop `from_yolov8` from documentation, README.md, examples.

SkalskiP:
@mayankagarwals It looks like you made a typo in `demo.ipynb`. I cant make it in the code. But double-check that, please. 🙏🏻

supervision/detection/core.py
```diff
@@ -170,39 +170,6 @@ def from_yolov5(cls, yolov5_results) -> Detections:
             class_id=yolov5_detections_predictions[:, 5].astype(int),
         )
 
-    @classmethod
```
SkalskiP, line 173: Bring back that implementation. We are going to drop it with `supervision-0.15.0`, so not in the next release.

SkalskiP (changes requested):

SkalskiP:
By the way, @hardikdava, I just noticed `sv.Classification.from_yolov8`. Do you think we should also create `sv.Classification.from_ultralytics` here as well, and deprecate `sv.Classification.from_yolov8`?

hardikdava:
@SkalskiP  of course, we should do that. I did not noticed it. Should I open a PR for it or you want to tackle that in this PR itself?

SkalskiP:
@hardikdava, I can do that. No problem.

supervision/detection/core.py
mayankagarwals: Done!. Makes sense

mayankagarwals:
> It looks like you made a typo in demo.ipynb. I cant make it in the code. But double-check that, please. 🙏🏻

Don't see anything suspicious. Let me know if there's something specific you'd like me to test!
image

demo.ipynb
mayankagarwals, line 1: @hardikdava @artyaltanzaya @SkalskiP 
Lot of diff with notebooks because of this https://github.com/jupyterlab/jupyterlab/issues/9444
Any recommended way to deal with it? Or should I let this be as is

SkalskiP: @mayankagarwals I'd prefer to avoid merging PRs with redundant changes like this. I suggest to rollback that commit. Make a manual change in just one line of the notebook and save it without running.

SkalskiP: > Weirdly, Any minor change in VS code notebook format results in 1k+ diffs. Something to dig later in to.

The beauty of Jupyter notebooks. 🤦🏻‍♂️

SkalskiP (approved):

SkalskiP:
It looks like we are ready to merge! 🔥

Issue #299, opened 2023-08-18
No error check for an empty input array to supervision.detection.utils.polygon_to_mask


ShubhamKanitkar32:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar bug report.

### Bug

`cv2.fillPoly` can not handle empty arrays. 

Error: `cv2.error: OpenCV(4.8.0) /io/opencv/modules/imgproc/src/drawing.cpp:2431: error: (-215:Assertion failed) p.checkVector(2, CV_32S) >= 0 in function 'fillPoly'`

### Environment

- Supervision: 0.13.0
- OS: Ubuntu 20.04
- Python: 3.8.10

### Minimal Reproducible Example

```
import supervision as sv
import numpy as np

polygon = np.array([[]], dtype=np.int64)
resolution_wh = (2048, 1536)

sv.detection.utils.polygon_to_mask(polygon, resolution_wh)
```

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

onuralpszr:
Thank you for reporting the issue @ShubhamKanitkar32  

But technically error check done by "opencv" side.  So I believe there is nothing to fix for this case, because sending empty array should be  cause problem already, or can you show me case mask would be okay if return "np.zeros((height, width))"  only ?

ShubhamKanitkar32:
Thank you for the response @onuralpszr 

There are cases when the `process_roboflow_result()` method is passed with empty (x, y) coordinates from the `roboflow` model response. In such scenarios, there either needs to be a way to make sure that they do not get added to the polygon [array](https://github.com/ShubhamKanitkar32/supervision/blob/develop/supervision/detection/utils.py#L364-L366) or Null check the polygon array before processing with fillPoly.

Apologies, the last part of your question is unclear to me. Hope ^this address it.

SkalskiP:
Hi, @ShubhamKanitkar32 👋🏻

Thanks for submitting the issue. Before making too quick concessions, @ShubhamKanitkar32, could you provide an example of the Roboflow response causing the exception?

As I understand, it can only happen when the user runs a segmentation model. This is only when the `points` key exists in response. The question that comes to my mind is how it is possible that a valid segmentation response can have empty `points` key.

ShubhamKanitkar32:
Hi @SkalskiP - sure, here is my testing setup.

- Yolov8 instance segmentation 
[...]

PR #300, opened 2023-08-18
Fix issue #299

ShubhamKanitkar32:
# Description

Fixes issue - https://github.com/roboflow/supervision/issues/299

## Type of change

Test script provided under the issue

### DIFF
--- a/supervision/detection/utils.py
+++ b/supervision/detection/utils.py
@@ -20,7 +20,10 @@ def polygon_to_mask(polygon: np.ndarray, resolution_wh: Tuple[int, int]) -> np.n
     """
     width, height = resolution_wh
     mask = np.zeros((height, width))
-    cv2.fillPoly(mask, [polygon], color=1)
+
+    # Empty numpy array check
+    if np.any(polygon):
+        cv2.fillPoly(mask, [polygon], color=1)
     return mask
 
 
--- a/supervision/detection/utils.py
+++ b/supervision/detection/utils.py
@@ -355,24 +355,25 @@ def process_roboflow_result(
         x_max = x_min + width
         y_max = y_min + height
 
-        xyxy.append([x_min, y_min, x_max, y_max])
-        class_id.append(class_list.index(prediction["class"]))
-        confidence.append(prediction["confidence"])
-
         if "points" not in prediction:
-            continue
-
-        if len(prediction["points"]) >= 3:
-            polygon = np.array(
-                [[point["x"], point["y"]] for point in prediction["points"]], dtype=int
-            )
-
+            xyxy.append([x_min, y_min, x_max, y_max])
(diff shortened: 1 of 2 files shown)


supervision/detection/utils.py
```diff
@@ -20,7 +20,10 @@ def polygon_to_mask(polygon: np.ndarray, resolution_wh: Tuple[int, int]) -> np.n
     """
     width, height = resolution_wh
     mask = np.zeros((height, width))
-    cv2.fillPoly(mask, [polygon], color=1)
+
+    # Empty numpy array check
```
artyaltanzaya, line 24: Thank you for addressing the bug; I suggest using if polygon.size > 0: to directly check for an empty array, ensuring against cases with all-zero coordinates.

SkalskiP:
Hi, @ShubhamKanitkar32 👋🏻! Thanks a lot for submitting a PR so fast. We definitely can't allow `supervision` to crash when processing `roboflow` results. The question that I have is: are we fixing the right thing? 

It seems reasonable for me for the `polygon_to_mask` to crash with invalid input. After all... you need at least three distinct points to draw a polygon on a 2d plane. 

Maybe we should move that fix to `sv.Detections.from_roboflow` and check if returned detection is valid? It is a bit surprising that the segmentation result can be an empty list of points and be valid.

ShubhamKanitkar32:
@SkalskiP no problem, happy to contribute!

I agree with you. It makes more sense for it to live under `sv.Detections.from_roboflow()` -> `process_roboflow_result()`

SkalskiP:
Awesome @ShubhamKanitkar32! 🔥 In that case, I have to ask you for two things:
- Accept our CLA. You can read more about it here: https://github.com/roboflow/supervision/pull/300#issuecomment-1684393454. This way, we can merge your PR later and credit you for your contribution.
- Move the conversation to issue: https://github.com/roboflow/supervision/issues/299 and provide me with the example of JSON you pass as an argument to `sv.Detections.from_roboflow()` that brakes the API. I'm interested not only in the list of points (I understand that it is empty) but the whole JSON object you pass.

Once again thanks for helping us to get to the bottom of this problem 🙏🏻

SkalskiP:
Hi @ShubhamKanitkar32 👋🏻! I think we are ready to implement. Take a look at my last comment under your issue: https://github.com/roboflow/supervision/issues/299#issuecomment-1686908979

SkalskiP (approved):

SkalskiP:
Hi, @ShubhamKanitkar32 I've made some changes to your implementation and added unit tests to make sure it works as expected. I'm merging changes into `develop`. Please check if everything works as expected.

SkalskiP:
btw, @onuralpszr I love per-commit

PR #330, opened 2023-08-29
Adjust text to diagonal and vertical lines

jcruz-ferreyra:
# Description

In this pull request, I've made some enhancements to `supervision/detection/line_counter.py` to provide more flexibility and control over the counting behavior of the `LineZone` class and the `LineZoneAnnotator` class.

- `LineZone` Class: I've introduced a new attribute called `direction` to the `LineZone` class. This attribute allows you to specify the direction in which elements will be counted. The possible values for this parameter are:
  - `in`: When viewing the line horizontally with the start point to the left, the `LineZone` annotates elements crossing up the line.
  - `out`: Similarly, when viewing the line horizontally with the start point to the left, the `LineZone` annotates elements crossing down the line.
  - `both` (default value): The `LineZone` annotates both elements crossing up and down the line.

Importantly, please note that the `direction` attribute does not affect the behavior of the `trigger` method. The `trigger` method continues to calculate elements crossing in both directions.

The primary impact of this enhancement is on the `annotate` method within the `LineZoneAnnotator` class. By introducing this attribute to the `LineZone` class, we can customize the counting behavior for each individual `LineZone`, avoiding the need to create multiple instances of `LineZoneAnnotator`.

- `LineZoneAnnotator` class, I've introduced a new function called `annotate_count` that is integrated into the existing `annotate` method. This enhancement streamlines the process of annotating count information for elements crossing the `LineZone`. The functionality of the `annotate_count` function can be conceptually divided into four key steps:

  1. **Text and Image Creation:** The function takes in the text for "in" and "out" counts as input. It then generates two background square images. The count text is annotated in the center of one image, and a text box is added to the other.

  2. **Rotation to Match LineZone Orientation:** Both background images are rotated to match the orientation of the `LineZone`. This ensures that the annotations align accurately with the direction of the line. For this step I've used `warpAffine` function from cv2 package.

  3. **Positioning and Cropping:** The rotated background images are positioned within the frame, taking into account the line's position and orientation. In cases where a portion of the images extends beyond the frame, they are cropped to fit within the frame boundaries.

  4. **Direct Annotation on Frame:** The original frame is annotated with the text box and text using pixel manipulation. By directly modifying the array representing the image, the count annotations are integrated into the frame.

## Type of change

The following colab notebook shows the result of the changes made. In it I tried LineZones with different orientation and direction parameter ("both","in", "out").

https://colab.research.google.com/drive/19yobSRlqgX_s8k9_oRw0QgIipKDjs7hx#scrollTo=nnD9XNEig9GD

No specific deployment considerations. Just one parameter added as an attribute for `LineZone` Class with default value "both" mantaining previous implementation.

## Docs

-   No changes made to Docs


jcruz-ferreyra:
@onuralpszr  Thank you for your feedback. I apologize for any inconvenience caused. It seems that I forked from the main branch instead of doing it directly from this branch. Sorry, I am working in fixing it.

onuralpszr:
@jcruz-ferreyra  thank you for re opening and no problem at all  :)

PR #349, opened 2023-09-02
fix: 🐛 Empty check for xyxy results for prevent ValueError and return as empty detection

onuralpszr:
# Description

We have multiple different detection and in case of video or multiple images it can be return empty so prevent that workflow break we should check all other detection type xyxy in case of empty boxes and continue 

## Type of change

### DIFF
--- a/supervision/detection/core.py
+++ b/supervision/detection/core.py
@@ -179,9 +179,6 @@ class Detections:
         """
         yolov5_detections_predictions = yolov5_results.pred[0].cpu().cpu().numpy()
 
-        if np.asarray(yolov5_detections_predictions[:, :4]).shape[0] == 0:
-            return cls.empty()
-
         return cls(
             xyxy=yolov5_detections_predictions[:, :4],
             confidence=yolov5_detections_predictions[:, 4],
@@ -220,9 +217,6 @@ class Detections:
             >>> detections = sv.Detections.from_yolov8(result)
             ```
         """
-
-        if np.asarray(yolov8_results.boxes.xyxy.cpu().numpy()).shape[0] == 0:
-            return cls.empty()
 
         return cls(
             xyxy=yolov8_results.boxes.xyxy.cpu().numpy(),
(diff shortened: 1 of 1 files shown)


onuralpszr:
@hardikdava  I was thinking bit more big before sending PR but anyway here we go could you check all of them please.

supervision/detection/core.py
```diff
@@ -178,6 +178,10 @@ def from_yolov5(cls, yolov5_results) -> Detections:
             ```
         """
         yolov5_detections_predictions = yolov5_results.pred[0].cpu().cpu().numpy()
+
+        if np.asarray(yolov5_detections_predictions[:, :4]).shape[0] == 0:
```
hardikdava, line 182: Tested and not needed.

supervision/detection/core.py
```diff
@@ -216,6 +220,10 @@ def from_yolov8(cls, yolov8_results) -> Detections:
             >>> detections = sv.Detections.from_yolov8(result)
             ```
         """
+
```
hardikdava, line 223: tested and not needed

hardikdava:
@onuralpszr I just tested and it seems this is not necessary at all. It is automatically handled when you convert output from pytorch to numpy.

supervision/detection/core.py
onuralpszr: .numpy() ones not needed okay. I will revert those but leave others then.
```suggestion
        if np.asarray(yolov5_detections_predictions[:, :4]).shape[0] == 0:
```

onuralpszr:
@hardikdava I revert the ones not needed and leave rest and add for sam too.

onuralpszr:
@hardikdava friendly ping :)

hardikdava (approved):
Everything seems fine to me. 👍

SkalskiP:
@onuralpszr up until now, all of those `.from_*` methods resulted in incorrect `sv.Detections` objects if the model result was empty? Do I understand that correctly?

onuralpszr:
> @onuralpszr up until now, all of those `.from_*` methods resulted in incorrect `sv.Detections` objects if the model result was empty? Do I understand that correctly?

Yes and we switch to empty for prevent "Value Error" in case of no detection

PR #363, opened 2023-09-12
Add Detections Filtering Methods to Improve Readability and Ease of Use

augustoolucas:
# Description

This pull request improves the existing Detections class by introducing intuitive and user-friendly filtering methods that enhance readability and simplify the process of filtering object detections. While the current approach of using numpy-like indexing for filtering is powerful, it can also become cumbersome and less readable, especially when dealing with mixed filtering conditions. The newly added filtering methods provide a more straightforward and expressive way to filter detections based on class, confidence, area, relative area, and box dimensions. Filtering by PolygonZone is not included as I think it should be added to the PolygonZone class.

The proposed addition of these new filtering methods **does not** remove support for numpy-like indexing, ensuring compatibility for users who prefer or rely on that approach.

The changes in this pull request include the addition of the following filtering methods to the Detections class:

1. [filter_by_class](https://github.com/augustoolucas/supervision/blob/filtering-methods/supervision/detection/core.py#L671): Allows filtering detections by class or classes, accepting a single class ID, a list of class IDs, or a numpy array of class IDs.

2. [filter_by_confidence](https://github.com/augustoolucas/supervision/blob/filtering-methods/supervision/detection/core.py#L702): Filters detections based on their confidence values, allowing for specifying a minimum and/or maximum confidence threshold.

3. [filter_by_area](https://github.com/augustoolucas/supervision/blob/filtering-methods/supervision/detection/core.py#L744): Filters detections based on their size, providing options to set a minimum and/or maximum area value for filtering.

4. [filter_by_relative_area](https://github.com/augustoolucas/supervision/blob/filtering-methods/supervision/detection/core.py#L780): Filters detections based on their area relative to the size of the image. It requires passing the total area of the image and allows setting minimum and/or maximum relative area thresholds.

5. [filter_by_box_dimensions](https://github.com/augustoolucas/supervision/blob/filtering-methods/supervision/detection/core.py#L836): Filters detections based on their bounding box dimensions, using width and/or height constraints. It allows for specifying minimum and/or maximum values for both width and height.

## Type of change

To test the changes I reproduced the documented filtering options, except for filtering by PolygonZone and mixed conditions. 

Results are not exactly the same as I don't know the model and image size used to generate the results presented on the docs.

### Filtering by specific class
```python

image = cv2.imread(os.path.join(PWD, "sv.jpg"))
result = inference_detector(model, image)
detections = sv.Detections.from_mmdetection(result)

class_0_detections = detections.filter_by_class(0)
assert class_0_detections == detections[detections.class_id == 0]

```
<p align="center">

</p>

### Filtering by a set of classes
```python
classes_0_2_3_detections = detections.filter_by_class([0, 2, 3])
assert classes_0_2_3_detections == detections[np.isin(detections.class_id, [0, 2, 3])]
```

<p align="center">

</p>

### Filtering by confidence
```python
high_confidence_detections = detections.filter_by_confidence(min_conf=0.5)
assert high_confidence_detections == detections[detections.confidence > 0.5]
```

<p align="center">

</p>

### Filtering by area
```python
big_detections = detections.filter_by_area(min_area=1000)
assert big_detections == detections[detections.area > 1000]
```
The following image is the result of an extra step of confidence filtering (min_conf=0.2) in addition to the area filtering above, just to make the final result less polluted.
<p align="center">

</p>

### Filtering by relative area
```python
image_area = image.shape[0] * image.shape[1]
small_detections = detections.filter_by_relative_area(
    image_area=image_area, max_relative_area=0.8
)
assert small_detections == detections[(detections.area / image_area) < 0.8]
```
The following image is the result of an extra step of confidence filtering (min_conf=0.2) in addition to the relative area filtering above, just to make the final result less polluted.

<p align="center">

</p>

### Filtering by box dimensions

```python
w = detections.xyxy[:, 2] - detections.xyxy[:, 0]
h = detections.xyxy[:, 3] - detections.xyxy[:, 1]
big_detections = detections.filter_by_box_dimensions(min_width=200, min_height=200)
assert big_detections == detections[(w > 200) & (h > 200)]
```

The following image is the result of an extra step of confidence filtering (min_conf=0.2) in addition to the box dimensions filtering above, just to make the final result less polluted.

<p align="center">

</p>

## Docs

### DIFF
--- a/supervision/detection/core.py
+++ b/supervision/detection/core.py
@@ -741,6 +741,42 @@ class Detections:
 
         return self[filtered_indices]
 
+    def filter_by_area(
+        self,
+        min_area: int = 0,
+        max_area: int = np.inf,
+    ) -> Detections:
+        """
+        Filter the Detections based on their size.
+
+        Args:
+            min_area (int):
+                The minimum area of the subset of the Detections
+            max_area (int):
+                The maximum area of the subset of the Detections
+
+        Returns:
+            (Detections): A subset of the Detections object.
+
+        Example:
+            ```python
+            >>> import supervision as sv
+
+            >>> detections = sv.Detections(...)
+
+            >>> big_detections = detections.filter_by_area(min_area=1000)
+
+            >>> small_detections = detections.filter_by_area(max_area=500)
+
+            >>> medium_detections = detections.filter_by_area(min_area=500,
+                                                              max_area=1000)
+            ```
+        """
+
+        filtered_indices = (self.area >= min_area) & (self.area <= max_area)
+
+        return self[filtered_indices]
+
     def __getitem__(
         self, index: Union[int, slice, List[int], np.ndarray]
     ) -> Detections:
--- a/supervision/detection/core.py
+++ b/supervision/detection/core.py
@@ -777,6 +777,60 @@ class Detections:
 
         return self[filtered_indices]
 
+    def filter_by_relative_area(
+        self,
+        image_area: int,
+        min_relative_area: float = 0,
+        max_relative_area: float = 1,
+    ) -> Detections:
+        """
+        Filter the Detections based on their size relative to the image size.
+
+        Args:
+            min_relative_area (float):
+                The minimum relative area of the subset of the Detections
+            max_relative_area (float):
+                The maximum relative area of the subset of the Detections
+
+        Returns:
+            (Detections): A subset of the Detections object.
+
+        Example:
+            ```python
+            >>> import supervision as sv
+
+            >>> detections = sv.Detections(...)
+
+            >>> image = ...
+
+            >>> height, width = image.shape[:2]
+
+            >>> image_area = height * width
+
+            >>> big_detections = detections.filter_by_relative_area(
+                    image_area=image_area,
+                    min_relative_area=0.8
+                )
+
+            >>> small_detections = detections.filter_by_relative_area(
+                    image_area=image_area,
+                    max_relative_area=0.3
+                )
+
+            >>> medium_detections = detections.filter_by_relative_area(
+                    image_area=image_area,
+                    min_relative_area=0.3,
+                    max_relative_area=0.7
+                )
+        """
+
+        relative_area = self.area / image_area
+        filtered_indices = (relative_area >= min_relative_area) & (
+            relative_area <= max_relative_area
+        )
+
+        return self[filtered_indices]
+
     def __getitem__(
         self, index: Union[int, slice, List[int], np.ndarray]
     ) -> Detections:
--- a/supervision/detection/core.py
+++ b/supervision/detection/core.py
@@ -690,7 +690,7 @@ class Detections:
 
             >>> class_0_detections = detections.filter_by_class(0)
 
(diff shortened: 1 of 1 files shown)


hardikdava:
Thanks for the PR. I like this idea personally. We will surely take a look into it. But be a patience we have overload on other requested features also.

hardikdava:
@SkalskiP @onuralpszr What do you think about this?

SkalskiP:
Hi @augustoolucas 👋🏻 First of all, I'm very sorry for the late reply. I've been on vacation for the last couple of weeks and am just getting back into the rhythm. And thank you very much for your interest in supervision. It means a lot to me that people volunteer their time and share their ideas

@hardikdava I was thinking about this PR. I deliberately designed the API to be consistent with pandas / numpy, so I'm more on the 'no' side. But I'm curious about your and @onuralpszr's opinion. 

I think we should invest time in creating examples and documentation so that people understand how effective the current API is. But I'm open for discussion :)

augustoolucas:
Hi @SkalskiP,  thank you for your work and feedback. If I may, I'd like to share my thoughts about it.

In my view, these methods don't compromise the consistency of the API with pandas/numpy as the changes maintain the support for the current filtering approach. It just provides an extra abstraction layer for users to focus on the intended filter rather than its implementation details.

Also, Pandas and Numpy are designed to be application-agnostic, this makes it not possible to have this kind of abstraction. Supervision's Detections class is more specific in representing a particular kind of data with known attributes. Therefore, having abstractions to handle it seems appropriate. Lastly, while it's not feasible to implement every possible filtering option, I believe that it's helpful to provide the basic ones.

SkalskiP:
Hi, @augustoolucas 👋🏻! Welcome back after the weekend. In general, I believe in the "less is more" principle. What worries me is that it adds more code that we need to maintain. Also, up until now, there is always one way of solving the problem. 

One thing that I know for sure is that if we will merge this PR we will need to move those utilities into a separate file. @onuralpszr, @hardikdava, and I, are already worried that our `Detection` class is getting larger and larger.

hardikdava:
@SkalskiP yes, I am completely agree with your thoughts on detection class is getting larger and larger. This will become difficult to maintain in the future.

augustoolucas:
Yes, for sure it is important to keep the codebase clean. We can further discuss the possibilities to make it happen, I'd be glad to help you on this.

SkalskiP:
@augustoolucas wanna help us out clean `sv.Detections`?

hardikdava:
@SkalskiP Since we all vote for not merging this. Let's close it and convert it into discussion for future. Thanks @augustoolucas for taking valuable time to contribute to supervision.

SkalskiP:
@hardikdava I think you're right. 

@augustoolucas once again thank you very much for your interest in supervision. If you'd like to help us with other tasks in the future, I'd be happy to see your next PR.

augustoolucas:
> @augustoolucas wanna help us out clean `sv.Detections`?



> 
> @augustoolucas once again thank you very much for your interest in supervision. If you'd like to help us with other tasks in the future, I'd be happy to see your next PR.

Sure! Is there already an ongoing discussion regarding this cleanup? Let me know where we can continue the discussions about it. Thank you guys.

PR #430, opened 2023-10-06
Fix return value when sv.Detections is empty

### LINKED ISSUE #429: Fix `sv.ByteTrack.tracker_id` return value when `sv.Detections` is empty
### Bug

When `sv.Detections` is empty (`len(detections) == 0`) `sv.ByteTrack` returns `tracker_id` equal `None`. Correct that behavior, make it more consistent, and return `np.array([], dtype=int)`.

### Fix

Add `else` statement here: https://github.com/roboflow/supervision/blob/89c1b63979d43f8ed651a9701ca8333034d0bc07/supervision/tracker/byte_tracker/core.py#L239

akashpambhar:
# Description

Fixes #429. When `sv.Detections` is empty (`len(detections) == 0`) `sv.ByteTrack` returns `np.array([], dtype=int)`

## Type of change

### DIFF
--- a/supervision/tracker/byte_tracker/core.py
+++ b/supervision/tracker/byte_tracker/core.py
@@ -249,6 +249,8 @@ class ByteTrack:
             detections.confidence = np.array(
                 [t.score for t in tracks], dtype=np.float32
             )
+        else:
+            return np.array([], dtype=int)
 
         return detections
 
--- a/supervision/tracker/byte_tracker/core.py
+++ b/supervision/tracker/byte_tracker/core.py
@@ -250,7 +250,7 @@ class ByteTrack:
                 [t.score for t in tracks], dtype=np.float32
             )
         else:
-            return np.array([], dtype=int)
+            detections.tracker_id = np.array([], dtype=int)
 
         return detections
 


SkalskiP:
Hi @akashpambhar 👋🏻! Thanks for the PR. 🙏🏻 Let me quickly test if everything works as expected.

supervision/tracker/byte_tracker/core.py
```diff
@@ -249,6 +249,8 @@ def update_with_detections(self, detections: Detections) -> Detections:
             detections.confidence = np.array(
                 [t.score for t in tracks], dtype=np.float32
             )
+        else:
+            return np.array([], dtype=int)
```
SkalskiP, line 253:
```suggestion
            detections.tracker_id = np.array([], dtype=int)
```

SkalskiP: Here we should update the `tracker_id` value.

SkalskiP (changes requested):

SkalskiP (approved):

SkalskiP:
Tested! Now, we are ready to merge. 👍🏻

Issue #418, opened 2023-10-05
Make `sv.ByteTrack` work with segmentation model


SkalskiP:
### Description

Our `sv.ByteTrack` only works with object detectors. If you use the segmentation model and push `sv.Detections` through `sv.ByteTracker`, you will lose information about masks.

### Code example

The code below will not draw any detections, as `byte_tracker.update_with_detections` call ic clearing all masks created by the model.

```python
import supervision as sv
from tqdm import tqdm
from ultralytics import YOLO

model = YOLO('yolov8x-seg.pt')
video_info = sv.VideoInfo.from_video_path(video_path='walking.mp4')
frame_generator = sv.get_video_frames_generator(source_path='walking.mp4')
byte_tracker = sv.ByteTrack()

with sv.VideoSink('walking-result.mp4', video_info=video_info) as sink:
    for frame in tqdm(frame_generator, total=video_info.total_frames):
        
        # segmentation info is created
        result = model.track(frame, imgsz=1920, verbose=False)[0]
        detections = sv.Detections.from_ultralytics(result)
        detections = byte_tracker.update_with_detections(detections)

        # segmentation info is no longer here
        annotated_frame = mask_annotator.annotate(
            scene=frame.copy(),
            detections=detections)

        sink.write_frame(frame=annotated_frame)
```

sitamgithub-MSIT:
@SkalskiP I want to work on this issue. Can you assign me this?

onuralpszr:
@sitamgithub-MSIT  I assigned to you, good luck :)

SkalskiP:
Closing this issue as it will be handled in https://github.com/roboflow/supervision/issues/754.

PR #436, opened 2023-10-07
#418 Make sv.ByteTrack work with segmentation model 

AntonioConsiglio:
**Description:**

This pull request addresses the following changes:

1. **ByteTrack Enhancement:**
   - Modified the `ByteTrack` module to retain information about the mask  #418 .

2. **Base Code Modification:**
   - Made adjustments to the base code to resolve the issue.

**Changes Made:**

- Added the attribute "mask" to the `STrack` class, initialized to `None` by default.
- Modified the `update` and `reactivate` methods of the `STrack` class to update the `mask` attribute with the new detection.

**Testing:**

Successfully tested the code with both object-detection and segmentation-models using the provided testing code:

```python
import supervision as sv
from tqdm import tqdm
from ultralytics import YOLO

# model = YOLO('yolov8x-seg.pt')
model = YOLO('yolov8x.pt')
video_info = sv.VideoInfo.from_video_path(video_path='walking.mp4')
frame_generator = sv.get_video_frames_generator(source_path='walking.mp4')
byte_tracker = sv.ByteTrack()
track_annotator = sv.TraceAnnotator()

with sv.VideoSink('walking-result.mp4', video_info=video_info) as sink:
    i = 0
    for frame in tqdm(frame_generator, total=video_info.total_frames):
        
        # Segmentation info is created
        result = model.track(frame, imgsz=1280, verbose=False)[0]
        detections = byte_tracker.update_with_detections(detections)

        # Segmentation info is no longer here
        annotated_frame = mask_annotator.annotate(
            scene=frame.copy(),
            detections=detections)
        annotated_frame = track_annotator.annotate(
            scene=annotated_frame.copy(),
            detections=detections)

        if i == 200:
            break
        i += 1

### DIFF
--- a/supervision/tracker/byte_tracker/core.py
+++ b/supervision/tracker/byte_tracker/core.py
@@ -11,7 +11,7 @@ from supervision.tracker.byte_tracker.kalman_filter import KalmanFilter
 class STrack(BaseTrack):
     shared_kalman = KalmanFilter()
 
-    def __init__(self, tlwh, score, class_ids):
+    def __init__(self, tlwh, score, class_ids, mask=None):
         # wait activate
         self._tlwh = np.asarray(tlwh, dtype=np.float32)
         self.kalman_filter = None
@@ -21,6 +21,7 @@ class STrack(BaseTrack):
         self.score = score
         self.class_ids = class_ids
         self.tracklet_len = 0
+        self.mask = mask
 
     def predict(self):
         mean_state = self.mean.copy()
@@ -74,6 +75,7 @@ class STrack(BaseTrack):
         if new_id:
             self.track_id = self.next_id()
         self.score = new_track.score
+        self.mask = new_track.mask if new_track.mask is not None else None
 
     def update(self, new_track, frame_id):
         """
@@ -94,6 +96,8 @@ class STrack(BaseTrack):
         self.is_activated = True
 
         self.score = new_track.score
+
+        self.mask = new_track.mask if new_track.mask is not None else None
 
     @property
     def tlwh(self):
@@ -231,9 +235,9 @@ class ByteTrack:
             ... )
             ```
         """
-
         tracks = self.update_with_tensors(
-            tensors=detections2boxes(detections=detections)
+            tensors=detections2boxes(detections=detections),
+            masks = detections.mask
         )
         detections = Detections.empty()
         if len(tracks) > 0:
@@ -249,12 +253,16 @@ class ByteTrack:
             detections.confidence = np.array(
                 [t.score for t in tracks], dtype=np.float32
             )
+            detections.mask = np.array(
+                [t.mask for t in tracks], dtype=bool
+            )
+
         else:
             detections.tracker_id = np.array([], dtype=int)
 
         return detections
 
-    def update_with_tensors(self, tensors: np.ndarray) -> List[STrack]:
+    def update_with_tensors(self, tensors: np.ndarray, masks:np.ndarray = None) -> List[STrack]:
         """
         Updates the tracker with the provided tensors and returns the updated tracks.
 
@@ -281,6 +289,12 @@ class ByteTrack:
         inds_second = np.logical_and(inds_low, inds_high)
         dets_second = bboxes[inds_second]
         dets = bboxes[remain_inds]
+        if masks is not None:
+            masks_hs = masks[remain_inds]
+            masks_ls = masks[inds_second]
+        else:
+            masks_hs = np.array([None]*len(remain_inds))
+            masks_ls = np.array([None]*len(inds_second))
         scores_keep = scores[remain_inds]
         scores_second = scores[inds_second]
 
@@ -290,8 +304,8 @@ class ByteTrack:
         if len(dets) > 0:
             """Detections"""
             detections = [
-                STrack(STrack.tlbr_to_tlwh(tlbr), s, c)
-                for (tlbr, s, c) in zip(dets, scores_keep, class_ids_keep)
+                STrack(STrack.tlbr_to_tlwh(tlbr), s, c,m)
+                for (tlbr, s, c,m) in zip(dets, scores_keep, class_ids_keep,masks_hs)
             ]
         else:
             detections = []
@@ -331,8 +345,8 @@ class ByteTrack:
         if len(dets_second) > 0:
             """Detections"""
             detections_second = [
-                STrack(STrack.tlbr_to_tlwh(tlbr), s, c)
-                for (tlbr, s, c) in zip(dets_second, scores_second, class_ids_second)
+                STrack(STrack.tlbr_to_tlwh(tlbr), s, c,m)
+                for (tlbr, s, c, m) in zip(dets_second, scores_second, class_ids_second,masks_ls)
             ]
         else:
             detections_second = []
--- a/supervision/tracker/byte_tracker/core.py
+++ b/supervision/tracker/byte_tracker/core.py
@@ -1,4 +1,4 @@
-from typing import List, Tuple
+from typing import List, Tuple, Optional
 
 import numpy as np
 
@@ -11,7 +11,7 @@ from supervision.tracker.byte_tracker.kalman_filter import KalmanFilter
 class STrack(BaseTrack):
     shared_kalman = KalmanFilter()
 
-    def __init__(self, tlwh, score, class_ids, mask=None):
+    def __init__(self, tlwh, score, class_ids, mask:Optional[np.array] = None):
         # wait activate
         self._tlwh = np.asarray(tlwh, dtype=np.float32)
         self.kalman_filter = None
@@ -259,13 +259,14 @@ class ByteTrack:
         return detections
 
     def update_with_tensors(
-        self, tensors: np.ndarray, masks: np.ndarray = None
+        self, tensors: np.ndarray, masks: Optional[np.array] = None
     ) -> List[STrack]:
         """
         Updates the tracker with the provided tensors and returns the updated tracks.
 
         Parameters:
             tensors: The new tensors to update with.
+            masks: The new masks associated to new tensors
 
         Returns:
             List[STrack]: Updated tracks.
@@ -288,11 +289,11 @@ class ByteTrack:
         dets_second = bboxes[inds_second]
         dets = bboxes[remain_inds]
         if masks is not None:
-            masks_hs = masks[remain_inds]
-            masks_ls = masks[inds_second]
+            masks_keep = masks[remain_inds]
+            masks_second = masks[inds_second]
         else:
-            masks_hs = np.array([None] * len(remain_inds))
-            masks_ls = np.array([None] * len(inds_second))
+            masks_keep = np.array([None] * len(remain_inds))
+            masks_second = np.array([None] * len(inds_second))
         scores_keep = scores[remain_inds]
         scores_second = scores[inds_second]
 
@@ -303,7 +304,7 @@ class ByteTrack:
             """Detections"""
             detections = [
                 STrack(STrack.tlbr_to_tlwh(tlbr), s, c, m)
-                for (tlbr, s, c, m) in zip(dets, scores_keep, class_ids_keep, masks_hs)
+                for (tlbr, s, c, m) in zip(dets, scores_keep, class_ids_keep, masks_keep)
             ]
         else:
             detections = []
@@ -345,7 +346,7 @@ class ByteTrack:
             detections_second = [
                 STrack(STrack.tlbr_to_tlwh(tlbr), s, c, m)
                 for (tlbr, s, c, m) in zip(
-                    dets_second, scores_second, class_ids_second, masks_ls
+                    dets_second, scores_second, class_ids_second, masks_second
                 )
             ]
         else:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by SkalskiP** [comment]
Hi, @AntonioConsiglio 👋🏻! Thanks a lot for submitting this PR. 🙏🏻 Let me do some tests.

**Comment by SkalskiP** [comment]
Looks like the code works. [Here](https://colab.research.google.com/drive/1PM8qYRvFS5oU9OBkalcbh9LAlPEuCVHt?usp=sharing) is the Google Colab I used. Here is the output:


https://github.com/roboflow/supervision/assets/26109316/c5a1a492-46ee-4dc5-8822-c2d9c903df1f

File: supervision/tracker/byte_tracker/core.py [hunk 1]
```diff
@@ -281,6 +288,12 @@ def update_with_tensors(self, tensors: np.ndarray) -> List[STrack]:
         inds_second = np.logical_and(inds_low, inds_high)
         dets_second = bboxes[inds_second]
         dets = bboxes[remain_inds]
+        if masks is not None:
```
SkalskiP, line 291: Can we use `_keep`, and `_second` suffixes here? And make the new code consistent with the old one?

supervision/tracker/byte_tracker/core.py
```diff
@@ -249,12 +252,16 @@ def update_with_detections(self, detections: Detections) -> Detections:
...
             detections.tracker_id = np.array([], dtype=int)
 
         return detections
 
-    def update_with_tensors(self, tensors: np.ndarray) -> List[STrack]:
+    def update_with_tensors(
+        self, tensors: np.ndarray, masks: np.ndarray = None
+    ) -> List[STrack]:
         """
         Updates the tracker with the provided tensors and returns the updated tracks.
```
SkalskiP, line 266: Please update the docstring to reflect the API change. Add and describe `masks` argument.

supervision/tracker/byte_tracker/core.py
```diff
@@ -249,12 +252,16 @@ def update_with_detections(self, detections: Detections) -> Detections:
...
+            detections.mask = np.array([t.mask for t in tracks], dtype=bool)
+
         else:
             detections.tracker_id = np.array([], dtype=int)
 
         return detections
 
-    def update_with_tensors(self, tensors: np.ndarray) -> List[STrack]:
+    def update_with_tensors(
+        self, tensors: np.ndarray, masks: np.ndarray = None
```
SkalskiP, line 263: Let's use Optional typing here. So: `masks: Optional[np.ndarray] = None`

SkalskiP (changes requested):

SkalskiP:
Hi, @CLAassistant 👋🏻! I've done a code review and testing. The code works as expected 🔥! Thanks a lot. I left three small comments regarding the implementation. Could you take a look? We are so close to merging it.

supervision/tracker/byte_tracker/core.py
AntonioConsiglio: I've done it )

SkalskiP:
@AntonioConsiglio, let me know once you are done.

SkalskiP (approved):

SkalskiP:
Awesome work @AntonioConsiglio! 🔥 Thanks for the quick turnaround. A very important contribution. Merging.

SkalskiP:
Hi @AntonioConsiglio 👋🏻 While working on a new Supervision example, I noticed this PR turned out to brake parts of supervision logic. Fir that reason, I'm forced to Revert it. I'm sorry. 

@AntonioConsiglio, let me know if you still wont to work on this peace of logic.

AntonioConsiglio:
> Hi @AntonioConsiglio 👋🏻 While working on a new Supervision example, I noticed this PR broke parts of supervision logic. For that reason, I'm forced to Revert it. I'm sorry.
> 
> @AntonioConsiglio, let me know if you still want to work on this logic.
> 
> @onuralpszr this is also a big lesson for me. No direct merging into `develop`. Everything goes through PR.

Of course, let me know what issue did u find

SkalskiP:
@AntonioConsiglio in short: updated ByteTrack was breaking [`traffic_analysis`](https://github.com/roboflow/supervision/tree/develop/examples/traffic_analysis) example. After some debugging, it turned out that if you were running the detection (not segmentation) model, the updated code still filled the `sv.Detection.mask` field. However, the shape of that `np.array` was incorrect. `sv.Detections.mask` is expected to be 3D `np.ndarray`, but it was 1D.

You can probably reproduce that error by running that example from this branch. 

After some thinking, I'd like to approach the same problem differently. In the code you proposed, we are putting our masks inside `STrack` instances. That can work but apart from the bug I noticed, it has one big downside. 

What if, in the future, we add more information to `sv.Detections`? What if someone extends `sv.Detections` and builds `CustomDetections(sv.Detections)` and stores some additional information? In short: this approach solves this local problem of supporting the segmentation model but does not scale to support future challenges. 

I think it would be a lot better to solve our problem [here](https://github.com/roboflow/supervision/blob/35142fc7551c1433cdbceba7042986d820f39fc4/supervision/tracker/byte_tracker/core.py#L237). At this stage we have `List[STrack]` and `sv.Detections`. We can use `box_iou_batch` to match tracks with detections - they will have the highest IoU, and then transfer `tracker_id` information from `List[STrack]` to `sv.Detections`. This way, we no longer need to care about copying all information from `sv.Detections`. Instead, we add information to `sv.Detections`.

AntonioConsiglio:
> @AntonioConsiglio in short: updated ByteTrack was breaking [traffic_analysis](https://github.com/roboflow/supervision/tree/develop/examples/traffic_analysis) example. After some debugging, it turned out that if you were running the detection (not segmentation) model, the updated code still filled the `sv.Detection.mask` field. However, the shape of that `np.array` was incorrect. `sv.Detections.mask` is expected to be 3D `np.ndarray`, but it was 1D.
> 
> You can probably reproduce that error by running that example from this branch.
> 
> After some thinking, I'd like to approach the same problem differently. In the code you proposed, we are putting our masks inside `STrack` instances. That can work but apart from the bug I noticed, it has one big downside.
> 
> What if, in the future, we add more information to `sv.Detections`? What if someone extends `sv.Detections` and builds `CustomDetections(sv.Detections)` and stores some additional information? In short: this approach solves this local problem of supporting the segmentation model but does not scale to support future challenges.
> 
> I think it would be a lot better to solve our problem [here](https://github.com/roboflow/supervision/blob/35142fc7551c1433cdbceba7042986d820f39fc4/supervision/tracker/byte_tracker/core.py#L237). At this stage we have `List[STrack]` and `sv.Detections`. We can use `box_iou_batch` to match tracks with detections - they will have the highest IoU, and then transfer `tracker_id` information from `List[STrack]` to `sv.Detections`. This way, we no longer need to care about copying all information from `sv.Detections`. Instead, we add information to `sv.Detections`.
> 
> I did that some time ago in this [notebook](https://colab.research.google.com/github/roboflow-ai/notebooks/blob/main/notebooks/how-to-track-football-players.ipynb). It was created before Supervision existed, so data is structured a bit differently, but the principle stays the same. The core logic lives here:
> 
> ```python
> def match_detections_with_tracks(
>     detections: List[Detection], 
>     tracks: List[STrack]
> ) -> List[Detection]:
>     detection_boxes = detections2boxes(detections=detections, with_confidence=False)
>     tracks_boxes = tracks2boxes(tracks=tracks)
>     iou = box_iou_batch(tracks_boxes, detection_boxes)
>     track2detection = np.argmax(iou, axis=1)
>     
>     for tracker_index, detection_index in enumerate(track2detection):
>         if iou[tracker_index, detection_index] != 0:
>             detections[detection_index].tracker_id = tracks[tracker_index].track_id
>     return detections
> ```
> 
> @AntonioConsiglio, would you be willing to try this approach?

Hi @SkalskiP , well i figured out what was going on and this will be simply solved with this line fo code:
```python
 masks = np.array([t.mask for t in tracks if t.mask is not None], dtype=bool)
            detections.mask = masks if len(masks) > 0 else None
```
This was a problem before because when no mask where present, the array was filled by N - None where N is the number of detections tracked. So before to assign the mask attribute I check if the t.mask is None (it means no mask for the tracked element) so if no Mask are present I assign None to the Detection.mask and I saw that you check the dimension of the mask only if it is not equal to None in your function. 

I don't see how can subclassing sv.Detections (if the attribute will be the same) can makes problems:

"What if, in the future, we add more information to sv.Detections? What if someone extends sv.Detections and builds CustomDetections(sv.Detections) and stores some additional information? In short: this approach solves this local problem of supporting the segmentation model but does not scale to support future challenges"

Feel free to ask if i was not clear enough :)

SkalskiP:
@AntonioConsiglio 

> I don't see how subclassing sv.Detections (if the attributes remain the same) can cause problems:

It is because you will lose extra information stored in that class, for example:

```python
class CustomDetections(sv.Detections):
    class_names: List[str]

# here, detections contains `class_names` value
detections = CustomDetections(...)

tracker = sv.ByteTrack()

# here, detections no longer contains class `class_names` value
```

AntonioConsiglio:
``> @AntonioConsiglio
> 
> > I don't see how subclassing sv.Detections (if the attributes remain the same) can cause problems:
> 
> It is because you will lose extra information stored in that class, for example:
> 
> ```python
> class CustomDetections(sv.Detections):
>     class_names: List[str]
> 
> # here, detections contains `class_names` value
> detections = CustomDetections(...)
> 
> tracker = sv.ByteTrack()
> 
> # here, detections no longer contains class `class_names` value
> detections = tracker.update_with_detections(detections)
> ```

Did you try it?
Because I've tested it whit this code
```python

class CustomDetections(sv.Detections):
    def __init__(self,*args,**kwargs):
        super().__init__(*args,**kwargs)
        self.class_name:list = []

with sv.VideoSink('walking-result.mp4', video_info=video_info) as sink:
    i = 0
    for frame in tqdm(frame_generator, total=video_info.total_frames):
        
        # segmentation info is created
        result = model.track(frame, imgsz=1280, verbose=False)[0]
        detections = CustomDetections.from_ultralytics(result)
        detections = byte_tracker.update_with_detections(detections)
```
class_name attribute was not "deleted".

image

SkalskiP:
@AntonioConsiglio, two things:

- Is the order of tracks the same as the order of detections? Because I'm almost sure it is not guaranteed.
- What if some object wont get `tracker_id`?

AntonioConsiglio:
> @AntonioConsiglio, two things:
> 
> * Is the order of tracks the same as the order of detections? Because I'm almost sure it is not guaranteed.
> * What if some object wont get `tracker_id`?

@SkalskiP 
The order of track - detection is guaranteed for one reason:

In the Tracking process you check which new detection is most similar to the previous one and assign the value of that detection to the tracked object (that if the tracking algorithm is not wrong should be the correct one)

Than in this order you populate the detection object (basically you change the value of the attribute of the CustomDetection). So basically all the attribute will have the tracked objects order after the tracking process

for the second question:

Why an object should not get a tracker_id if you assign it every time there is a new Object to track? also, why this can be related  to the change I've done?

SkalskiP:
> In the Tracking process you check which new detection is most similar to the previous one and assign the value of that detection to the tracked object (that if the tracking algorithm is not wrong should be the correct one)

The current process does not guarantee that `tracks` and `detections` have the same order. It simply overwrites `detections` content with information from `tracks`. 

> Why an object should not get a tracker_id if you assign it every time there is a new Object to track? also, why this can be related to the change I've done?

We have proven with @paulguerrie last week that when a new object appears on the scene during the video, it does not get `tracker_id` for the first few frames. Essentially what is happening right now, is when you call `detections = tracker.update_with_detections(detections)` those new objects are removed from `detections` for the first few frames because they are not in `tracks`. Using `box_iou_batch` would allow us to change that behavior. We could assign them some default `tracker_id` like `-1`.

AntonioConsiglio:
> > In the Tracking process you check which new detection is most similar to the previous one and assign the value of that detection to the tracked object (that if the tracking algorithm is not wrong should be the correct one)
> 
> The current process does not guarantee that `tracks` and `detections` have the same order. It simply overwrites `detections` content with information from `tracks`.
>
 
Yeah but tracks object were updated with the new detections value before, so each new detecion is assign to the corrisponding track object

> > Why an object should not get a tracker_id if you assign it every time there is a new Object to track? also, why this can be related to the change I've done?
> 
> We have proven with @paulguerrie last week that when a new object appears on the scene during the video, it does not get `tracker_id` for the first few frames. Essentially what is happening right now, is when you call `detections = tracker.update_with_detections(detections)` those new objects are removed from `detections` for the first few frames because they are not in `tracks`. Using `box_iou_batch` would allow us to change that behavior. We could assign them some default `tracker_id` like `-1`.

I will check this part and let you know.

SkalskiP:
> I will check this part and let you know.

Sure ;)

AntonioConsiglio:
> > I will check this part and let you know.
> 
> Sure ;)
@SkalskiP 
In my test I notice that the ID are assigned immediately (see picture)
image

So i can't see the problem you have mentioned before at inference pipeline

SkalskiP:
@AntonioConsiglio the video is not loading :/

AntonioConsiglio:
> @AntonioConsiglio the video is not loading :/

I attached a shorter version

AntonioConsiglio:
I'm wating @SkalskiP

PR #466, opened 2023-10-12
New feature HeatmapAnnotator

kapter:
# Description

Annotator for drawing heatmap on an image using provided detections.
Heat is accumulated over time. Drawn as a semi-transparent overlay as blured circles.

## Demo

https://github.com/roboflow/supervision/assets/6329901/4a7aa682-59ea-4c34-b569-8e6b40cea5d3

## Type of change

## Docs

### DIFF
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -84,18 +84,19 @@ class BoundingBoxAnnotator(BaseAnnotator):
             )
         return scene
 
+
 class HeatmapAnnotator:
     """
     A class for drawing heatmap on an image using provided detections.
     Heat is accumulated over time. Drawn as a semi-transparent overlay as blured circles.
     """
+
     def __init__(
         self,
         position: Optional[Position] = Position.BOTTOM_CENTER,
         opacity: float = 0.2,
         radius: int = 40,
-        kernel_size: Optional[int]  = 25,
-        
+        kernel_size: Optional[int] = 25,
     ):
         """
         Args:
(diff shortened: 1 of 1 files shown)


supervision/annotators/core.py
```diff
@@ -85,6 +85,118 @@ def annotate(self, scene: np.ndarray, detections: Detections) -> np.ndarray:
...
+            radius (int): Radius of the heat circle.
+            kernel_size (Optional[int]): Kernel size for blurring the heatmap.
+        """
+        self.position = position
+        self.opacity = opacity
+        self.radius = radius
+        self.kernel_size = kernel_size
+        self.heatmask = None
+
+    def get_xy(
```
SkalskiP, line 115: Hi @kapter 👋🏻! I think we already have that logic implemented: https://github.com/roboflow/supervision/blob/9ded70c8daae214e53cb7b33559c50817af14e68/supervision/detection/core.py#L603

SkalskiP (changes requested):

supervision/annotators/core.py
kapter: It's my mistake, I saw same logic in LabelAnnotator and repeat it but now I understood why.
I can create new commit tonight.

kapter:
Thanks for response.
You can get annotator attribute heatmask. This ndarray consist info about counts of detections in every pixel.
I will try this on long video from road-parking.

AlainPilon:
Is'nt the 90 on this line used for the decay? (havnet had enough time to look over the whole thing)
https://github.com/kapter/supervision/blob/1324cda60080884d8c867dd44288aca285f5a250/supervision/annotators/core.py#L159C16-L159C44

Just giving access to this variable (as a float) would be an easy solution.

SkalskiP:
Hi @AlainPilon 👋🏻 wanna propose some changes? I will probably work on merging this annotator tomorrow. So if you have some ideas how to change implementation this is good time :)

AlainPilon:
@SkalskiP nothing I can personally fix in the next few days. A quick change would have been to move the decay value into the signature but I feel that if we start doing this, we will end up with 20 kwargs to control every aspect of the annotator. 

A better solution would be to split the apparence of the annotator and its behaviour into 2 config objects but it is not something I can work on for a while. On the positive side, I do have to create a custom annotator for work which is going to use this pattern so, once I am done, I could retrofit the change in this one.

kapter:
I think the main purpose of the annotator is to give a quick representation of the frequency of objects appearing in a particular part of the frame. It seems there is no need to complicate it.

kapter:
I found this
image
May be later add option: track vs detection heat map?

AlainPilon:
To me, the example you showed is a mix of tracking the movement over a long duration + the heat map with a small decay. Visualization of a parking lot comes to mind and would be a great use case for this.

I agree that for a first version of the feature, a simple version could work. But since Supervision is a framework it has to be open to the multiple potential usage. This openness could be with configuration options OR by making it very easy to overwrite existing annotators.

I have nothing against the current version of the class, other than thinking the `90` value should be a well named constant or a params. Because right now it is a magic number and unless you run some maths, it is very hard to figure out ahead of time the behaviour of the heat value over time.

kapter:
Preview

https://github.com/roboflow/supervision/assets/6329901/95172fb0-d18d-40ff-930b-62010e776040

SkalskiP:
Merging! We need to update the docs. But I'll do it myself! 🔥 Thank you @kapter! 🙏🏻

kapter:
Thank you, @SkalskiP, for your assessment.

PR #483, opened 2023-10-16
Add 404 page

capjamesg:
This PR adds a 404 page to `supervision`.

Here is the current page someone sees when they navigate to a broken URL:

Screenshot 2023-10-16 at 09 02 43

Here is the page this PR adds:

Screenshot 2023-10-16 at 09 03 06

### DIFF
--- a/custom_theme/404.html
+++ b/custom_theme/404.html
@@ -1,0 +1,7 @@
+{% extends "main.html" %} 
+
+{% block content %}
+    <h1>Page Not Found</h1> 
+    <p>The page you have requested was not found.</p>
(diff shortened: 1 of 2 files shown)


SkalskiP:
@capjamesg, should this PR solve the problem with this URL: https://supervision.roboflow.com/detection/annotate/

onuralpszr:
> @capjamesg, should this PR solve the problem with this URL: https://supervision.roboflow.com/detection/annotate/

is *.roboflow.com behind CDN/proxy (like cloudflare) because I am still getting problematic page. If that is the case could you please purge cache ?

Issue #461, opened 2023-10-12
Extend `sv.Detections` with `data` field to support custom payload


SkalskiP:
### Description

Add `data: Dict[str, np.ndarray] = field(default_factory=dict)` to `sv.Detections` to support the custom payload. That can be `class_name`, or other detection-related metadata. To ensure full compatibility, the `__post_init__`, `__iter__`, `__eq__`, `merge` and `__getitem__` methods should be updated.

### ⚠️ Disclaimer

Maintaining data consistency is critical. We use `np.ndarray`s to store information: 

- `xyxy` with `(N, 4)` shape
- `mask` with `(N, H, W)` shape
- `confidence` with `(N, )` shape
- `class_id` with `(N, )` shape
- `tracker_id` with `(N, )` shape

Notice that the shape of every `np.ndarray` is dictated by `N` - detections count. When designing `data: Dict[str, np.ndarray]`, we must ensure that every value in that dictionary has `(N, ...)` shape.

subhajit20:
hey @SkalskiP want to work on this.
Can you assign it to me?

SkalskiP:
Yup, @subhajit20 you are assigned! Let me know if you will need any help. That task might be a bit tricky.

subhajit20:
> Yup, @subhajit20 you are assigned! Let me know if you will need any help. That task might be a bit tricky.

can you guide me on how to solve this?

hiteshhedwig:
hi, @SkalskiP can you explain or provide an example. what's problem does this task we are trying to solve? it'd really help in the context of solving it

SkalskiP:
Hi @subhajit20 and @hiteshhedwig 👋🏻! The idea is to make `sv.Detections` more flexible and applicable to more use cases. 

Imagine you want to use supervision in an OCR project. In such projects, it is usually a matter of knowing what text was read and where it was located. You can store text bounding boxes in `sv.Detection.xyxy`, but where do you put associated text? Right now, this cannot be done with `sv.Detections` alone. 

We aim to allow users to store additional information in this example text extracted from the image.

```python
detections = sv.Detections(...)
extracted_text = [...]
detections.store('extracted_text', extracted_text)
```

Of course, OCR is just an example. There is an infinite amount of similar scenarios.

beasteers:
This is the modification I've been using.

The nice thing about this is that you can add arbitrary data the same as you would define `mask` or `confidence`.

```python
detections = Detections(
    xyxy=np.array([...]),
[...]

PR #497, opened 2023-10-18
add data payload dict #461

hiteshhedwig:
# Description

Addressing the following - #461  data payload enhancement feature in `Detections`.  
Adding data payload to `Detections`, in order to allow more flexibility. 

## Type of change

YOUR_ANSWER

## Docs

### DIFF
--- a/supervision/detection/core.py
+++ b/supervision/detection/core.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
-from dataclasses import astuple, dataclass
-from typing import Any, Iterator, List, Optional, Tuple, Union
+from dataclasses import astuple, dataclass, field
+from typing import Any, Iterator, List, Optional, Tuple, Union, Dict
 
 import numpy as np
 
@@ -12,6 +12,17 @@ from supervision.detection.utils import (
     xywh_to_xyxy,
 )
 from supervision.geometry.core import Position
+
+
+def _validate_array(value: Any, n: int, name: str) -> None:
+    """
+    Validates an array ensuring it's either None or a numpy array with its first dimension being n.
+    """
+    is_valid = value is None or (
+        isinstance(value, np.ndarray) and value.shape[0] == n
+    )
+    if not is_valid:
+        raise ValueError(f"{name} must be None or np.ndarray with its first dimension being n.")
 
 
 def _validate_xyxy(xyxy: Any, n: int) -> None:
@@ -58,6 +69,9 @@ def _validate_tracker_id(tracker_id: Any, n: int) -> None:
     if not is_valid:
         raise ValueError("tracker_id must be None or 1d np.ndarray with (n,) shape")
 
+def _validate_data_payload(data_payload: Dict[str, np.ndarray], n: int) -> None:
+    for key, value in data_payload.items():
+        _validate_array(value, n, f"array in data payload - ({key})")
 
 @dataclass
 class Detections:
@@ -74,6 +88,8 @@ class Detections:
             `(n,)` containing the class ids of the detections.
         tracker_id (Optional[np.ndarray]): An array of shape
             `(n,)` containing the tracker ids of the detections.
+        data (Dict[str, np.ndarray]): A dictionary to support custom payload. This 
+            can store detection-related metadata.
     """
 
     xyxy: np.ndarray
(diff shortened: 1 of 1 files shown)


hiteshhedwig:
@SkalskiP - hi, let me know what do you think about the PR. I tried adding as much as i could. But as you said, its a little tricky. Plus, in your example #461 . 
I am following the guideline as you said - 
Notice that the shape of every `np.ndarray` is dictated by `N` - detections count. When designing `data: Dict[str, np.ndarray]`, we must ensure that every value in that dictionary has `(N, ...)` shape.

But it contradicts your example of OCR in the comment. storing `extracted_text` into data payload. its going to be standalone string, right? . maybe `(N, ...)` shape doesnt fit well in OCR example? Please correct me if you feel so.

Let me know, what do you think about the PR

SkalskiP:
Hi @hiteshhedwig 👋🏻! Sorry for lagging with the review. I'll taka a look on Monday.

supervision/detection/core.py
```diff
@@ -592,12 +627,27 @@ def __all_not_none(item_list: List[Any]):
         class_id = np.hstack(class_id) if __all_not_none(class_id) else None
         tracker_id = np.hstack(tracker_id) if __all_not_none(tracker_id) else None
 
+        # Extract all unique keys from the data_list
```
SkalskiP, line 630: Please, remove comments

supervision/detection/core.py
```diff
@@ -692,12 +742,15 @@ def __getitem__(
         """
         if isinstance(index, int):
             index = [index]
+        # Indexing the custom data dictionary
```
SkalskiP, line 745: Please, remove comments

supervision/detection/core.py
```diff
@@ -59,6 +67,11 @@ def _validate_tracker_id(tracker_id: Any, n: int) -> None:
         raise ValueError("tracker_id must be None or 1d np.ndarray with (n,) shape")
 
 
+def _validate_data_payload(data_payload: Dict[str, np.ndarray], n: int) -> None:
```
SkalskiP, line 70: Let's rename the function to `_validate_data`.

SkalskiP (changes requested):

supervision/detection/core.py
```diff
@@ -592,12 +627,27 @@ def __all_not_none(item_list: List[Any]):
         class_id = np.hstack(class_id) if __all_not_none(class_id) else None
         tracker_id = np.hstack(tracker_id) if __all_not_none(tracker_id) else None
 
+        # Extract all unique keys from the data_list
+        all_keys = set().union(*(d.keys() for d in data_list))
```
SkalskiP, line 631: When we merge `Detections` objects we should make sure that their `data` dictionaries have the same `key`.

SkalskiP (changes requested):

supervision/detection/core.py
```diff
@@ -105,22 +122,39 @@ def __iter__(
             Optional[float],
             Optional[int],
             Optional[int],
+            Dict[str, np.ndarray],
```
SkalskiP, line 125: That type should be `Dict[str, Any]`

SkalskiP:
Hi @hiteshhedwig! I made the review and left a few comments. Overall, we are moving in the right direction! 🔥

supervision/detection/core.py
hiteshhedwig: Done

supervision/detection/core.py
hiteshhedwig: Done

supervision/detection/core.py
hiteshhedwig: Thanks! Done

supervision/detection/core.py
hiteshhedwig: see it now? i am using `warnings.warn("All data payload dictionaries in Detections object do not have the same keys!")` just incase dictionaries dont have the same key, code doesn't silently execute without letting the user know? I thought, raising valueerror could be little tricky in this case. what do you think ?

hiteshhedwig: should we raise value error, or just warning ? or None?

hiteshhedwig:
@SkalskiP Hi, can you let me know if some changes are still recommended ?

SkalskiP:
Hi @hiteshhedwig 👋🏻 ! I'm closing this PR as this change was already implemented in the `0.18.0` release. Thanks a lot for the help and time you spend crafting this code.

Issue #268, opened 2023-08-06
Weighted Box Fusion


hardikdava:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Description

Current object detection models removes overlapping boxes by `nms` which can reduce accuracy of the final results. It can be avoided by `Weighted Box Fusion` which can accepts all the prediction whether from a single model or multiple models. 

Reference: [Weighted Box Fusion](https://arxiv.org/pdf/1910.13302.pdf)
Implementation: [ensemble-boxes](https://github.com/ZFTurbo/Weighted-Boxes-Fusion/tree/master)

### Use case

```python
import supervision as sv

wbf = sv.WeightedBoxFusion()

res_a = model_a(image)
det_a = sv.Detection(res_a)

res_b = model_b(image)
det_b = sv.Detection(res_b)

wbf_detections = wbf([det_a, det_b])
```

### Additional

```python

class WeightedBoxFusion:

	def __init__(self):
		pass
	
	def __call__(self, detections: Union[Detections, List[Detections]]) -> Detections:
		pass
		result_detections = Detections(...)
		return result_detections 

```

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

hardikdava:
@SkalskiP yes, it is better in terms of accuracy but not in terms of speed. You can learn about it more and direct comparision on this blogpost: https://learnopencv.com/weighted-boxes-fusion/

AlainPilon:
@hardikdava Did you implement the WBF? Even if Supervision doesnt accept the feature, I could use it for my current project.

SkalskiP:
@hardikdava, if you have that implemented somewhere already, I'd love to see it. I know that you are probably super busy, but some of the open-source contributors may be able to add it to our codebase.

AlainPilon:
I found this: https://github.com/ZFTurbo/Weighted-Boxes-Fusion which does most of the work, we just need the detections coordinates to be as % instead of absolute pixel values. The packages also does NMS which would be a feature overlap with the existing Supervision code. I dont know how we should handle this: 

- extract just want we need for WBF, or
- replace the existing NMS code with the one in the package (and get a few extra features for free). 

My working version of Supervision has drifted quite a bit from the current release. Anything I do on my side would probably not be mergeable in main.

hardikdava:
[...]

PR #500, opened 2023-10-18
Add Non-Maximum Merging (NMM) to Detections

mario-dg:
# Description

Currently it is only possible to apply Non-Maximum Suppression (NMS) to Detections.
In an attempt to increase the accuracy of object detection and segmentation, especially when using the InferenceSlicer,
I added the ability to apply NMM instead of NMS to the merged Detections [merge_detections](https://github.com/mario-dg/supervision/blob/6f4704625b16ba69068b3a19f6d55bc21c80c434/supervision/detection/tools/inference_slicer.py#L125C8-L125C8).

NMM merges bounding boxes and masks, that would have originally been discarded, when applying NMS.
This results in more accurate detections/segmentations.

The implementation of this algorithm closely follows the one from [SAHI](https://github.com/obss/sahi/blob/b115f08f0c0aeeb151adf24e47c222d3483cc931/sahi/postprocess/combine.py#L533).

In addition, also inspired by SAHI, the InferenceSlicer can increase the detection accuracy of larger objects, by running inference
on the whole image on top of all slices [perform_standard_pred](https://github.com/mario-dg/supervision/blob/6f4704625b16ba69068b3a19f6d55bc21c80c434/supervision/detection/tools/inference_slicer.py#L122C8-L122C8). The whole image detections will be merged with the sliced image detections.

I would be happy to discuss implementation changes and design choices.
This PR should serve as a first implementation and is far from perfect.
Any ideas, critic or comments are more than welcome :)

## Type of change

NMS:
nms

NMM:
nmm

SAHI:
sahi

The code sample that was used to create these two images:
```python
import cv2
import numpy as np
import supervision as sv

from supervision.detection.core import Detections
from supervision.detection.tools.inference_slicer import InferenceSlicer

from PIL import Image
from ultralytics.models.yolo import YOLO


model = YOLO("yolov8s.pt")
def callback(x):
    result = model(x[:, :, ::-1])[0]
    dets = Detections.from_ultralytics(result)
    return dets[dets.confidence > 0.3]

# get image from here: https://ultralytics.com/images/bus.jpg

slicer = InferenceSlicer(callback=callback, 
                         slice_wh=(512, 512), 
                         overlap_ratio_wh=(0.4, 0.4), 
                         iou_threshold=0.1, 
                         merge_detections=True, 
                         perform_standard_pred=True)

ba = sv.BoundingBoxAnnotator()
la = sv.LabelAnnotator()
image = np.array(Image.open("bus.jpg"))

nmm_dets = slicer(image)
nmm_ann = ba.annotate(scene=image.copy(), detections=nmm_dets)
nmm_ann = la.annotate(scene=nmm_ann, detections=nmm_dets)
cv2.imwrite("nmm.jpg", nmm_ann)

slicer.merge_detections = False

nms_dets = slicer(image)
nms_ann = ba.annotate(scene=image.copy(), detections=nms_dets)
nms_ann = la.annotate(scene=nms_ann, detections=nms_dets)
cv2.imwrite("bms.jpg", nms_ann)

```

The two additional parameters that were added to InferenceSlicer are optional and default to `False`, thus neither breaking, nor changing the results of existing implementations.

## Docs

-   [ ] Docs updated? What were the changes: I will update the docs in the near future.


SkalskiP:
Hi @mario-dg 👋🏻! Sorry for lagging with the review. I will take a look at it on Monday.

mario-dg:
Hi @SkalskiP👋, no worries! 
Must have been quite busy with the latest Release! Congrats on that one🚀

kadirnar:
This is great feature. When will it be merged? @SkalskiP

SkalskiP:
@kadirnar, I need to review it first. I didn't have time to work on it yet.

SkalskiP:
Hi @mario-dg 👋🏻 We aim to significantly enhance our `InferenceSlicer` implementation. One of the challenges we've encountered involves limitations associated with applying `non_max_suppression` in the area of overlapping tiles. Would you still be interested in continuing to assist us?

mario-dg:
Hey @SkalskiP, thanks for coming back to this PR!
Sure, I would love to continue🚀

SkalskiP:
Hi @mario-dg 👋🏻 that's awesome! 

We are planning a major wave of updates in InferenceSlicer that includes:

- Support for segmentation models
- Support for batch processing (running inference on multiple tiles at once)
- Improved post-processing - merging instead of removing overlapping detections

This PR looks like the perfect candidate to achieve the last of these goals.

LinasKo:
Hey @mario-dg :wave:

I'm here to help out if you need anything.

For some broader context, we're thinking of implementing #268 with this method and later linking the better of the two (or both) with the `InferenceSlicer` in a separate PR.

I was wondering - what's the state of this? At the point where you contributed, was it ready to test & merge?
Would you mind changing the code so it's a standalone `utils` method, not part of `InferenceSlicer` yet?

Lastly, if you have the time, could you test it on a video?
You can find one [here](https://storage.googleapis.com/com-roboflow-marketing/supervision/video-examples/beach-1.mp4)

mario-dg:
Thats great! Supervision might become a much more suitable lightweight alternative to SAHI, once we've improved its capabilities. Anything specific I could work on?

SkalskiP:
@mario-dg 🔥 Please start by solving conflicts. This PR is quite old and we need to make it up to date. Than @LinasKo will review the PR.

LinasKo:
Hi @mario-dg :wave:

We spoke with **SkalskiP** - I'll see if I can review your code today.
We know of a few changes still needed, but I need to ramp-up a bit.

The broader context is that there's another candidate solution #268 and we'd like to compare the two before adding to `InferenceSlicer`.

mario-dg:
Alright @SkalskiP, will get to it this evening.

supervision/detection/utils.py
```diff
@@ -110,6 +110,196 @@ def non_max_suppression(
...
+    Returns:
+        np.ndarray: A numpy array of shape `(, 4)` where the new
+            bounding box is the merged bounding box of `bbox1` and `bbox2`.
+    """
+    left_top = np.minimum(bbox1[0][:2], bbox2[0][:2])
+    right_bottom = np.maximum(bbox1[0][2:], bbox2[0][2:])
+    return np.array([np.concatenate([left_top, right_bottom])])
+
+
+def get_merged_class_id(id1: np.ndarray, id2: np.ndarray) -> np.ndarray:
```
LinasKo, line 242: Calling `id1.item()` on an array longer than 1 element throws an error.

While I like how the merge behavior is encapsulated, if the users find this function, it'd feel very unintuitive. (It's tightly coupled with `_merge_object_detection_pair`, which is private)

It feels less clean, but I'd much rather have the logic locked within `_merge_object_detection_pair` than exposed in the `utils` module. I don;t think any of these are used in `greedy_nmm` or `batched_greedy_nmm`, which can stay here.

This also applies to all functions defined below.

supervision/detection/utils.py
```diff
@@ -110,6 +110,196 @@ def non_max_suppression(
...
+    Merges two class ids into one.
+
+    Args:
+        id1 (np.ndarray): The first class id.
+        id2 (np.ndarray): The second class id.
+
+    Returns:
+        np.ndarray: The merged class id.
+    """
+    return np.array([max(id1.item(), id2.item())])
```
LinasKo, line 253: Let's switch to `min`. If we're merging two objects, we'd slightly favor IDs that were issued earlier.

supervision/detection/utils.py
```diff
@@ -110,6 +110,196 @@ def non_max_suppression(
...
+    Merges two confidences into one.
+
+    Args:
+        confidence1 (np.ndarray): The first confidence.
+        confidence2 (np.ndarray): The second confidence.
+
+    Returns:
+        np.ndarray: The merged confidence.
+    """
+    return np.array([max(confidence1.item(), confidence2.item())])
```
LinasKo, line 269: We aren't as confident about the larger area - there's outlier cases where this is dangerous.

In theory, is there a way to draw weights from this - to know which area contributed how much; to derive the weights?
Alternatively, I think `mean` of the two is more acceptable than `max`.

supervision/detection/utils.py
```diff
@@ -110,6 +110,196 @@ def non_max_suppression(
...
+    Merges two tracker ids into one.
+
+    Args:
+        tracker_id1 (int): The first tracker id.
+        tracker_id2 (int): The second tracker id.
+
+    Returns:
+        int: The merged tracker id.
+    """
+    return max(tracker_id1, tracker_id2)
```
LinasKo, line 300: Same as with IDs - we'd prefer the one issued first. Let's change to min.

supervision/detection/utils.py
```diff
@@ -110,6 +110,196 @@ def non_max_suppression(
...
+            are the height and width of the mask, respectively.
+
+    Returns:
+        np.ndarray: A numpy array of shape `(H, W)` where the new mask is the
+            merged mask of `mask1` and `mask2`.
+    """
+    return np.logical_or(mask1, mask2)
+
+
+def get_merged_tracker_id(tracker_id1: int, tracker_id2: int) -> int:
```
LinasKo, line 289: `tracker_id` can have N trackers - one for each object in `Detections`. It's a bit inconsistent in how it's used within the codebase at the moment ( #1092 ). The bottom line is: I think the API of this function should be the same as with `class_id`.

Later, as part of #1092, we'll add support for special `-1` values and update `NMM` functions.

supervision/detection/utils.py
LinasKo, line 1: @SkalskiP, we both agreed that `utils` is getting bloaty by now and needs a split. I suggest we hold off until this PR is done, the weighted boxes PR is done and do a `utils` refactor either before or after adding into the `inference_slicer`.

This removes the complexity for contributors now, and we'd have more clear-cut lines along which to split the file. There's probably more than 2 resulting files anyway - might as well do it thoughtfully.

supervision/detection/core.py
```diff
@@ -729,6 +790,64 @@ def box_area(self) -> np.ndarray:
         """
         return (self.xyxy[:, 3] - self.xyxy[:, 1]) * (self.xyxy[:, 2] - self.xyxy[:, 0])
 
+    def with_nmm(
```
LinasKo, line 793: This is a very clean implementation and we like it greatly. However, since there's another method we may or may not add, the function API might change, so let's remove this from inference slicer for now. At the same time, it's a good format for a function to be used when testing.

supervision/detection/tools/inference_slicer.py
```diff
@@ -36,6 +36,12 @@ class InferenceSlicer:
             slices in the format `(width_ratio, height_ratio)`.
         iou_threshold (Optional[float]): Intersection over Union (IoU) threshold
             used for non-max suppression.
+        merge_detections (Optional[bool]): Whether to merge the detection from all
+            slices or simply concatenate them. If `True`, Non-Maximum Merging (NMM),
+            otherwise Non-Maximum Suppression (NMS), is applied to the detections.
+        perform_standard_pred (Optional[bool]): Whether to perform inference on the
```
LinasKo, line 42: Let's not touch `InferenceSlicer` for now. We can add this on step 3 - after we have NMM and weighted boxes methods.

Alternatively, if you make a separate PR, it'd be easier to test in isolation - likely easier to merge too :)

LinasKo (changes requested):
Hey @mario-dg,

I've reviewed the PR. Frankly, I think it's a really well structured, well-documented contribution. I especially like how you managed to match the function API and documentation to what we have in `with_nms` :slightly_smiling_face: 

I've noted down the changes that we'd need. Some extra tests would be good to do as well.
* You can find one here: https://storage.googleapis.com/com-roboflow-marketing/supervision/video-examples/beach-1.mp4
* There's some code to run the models here. We're especially interested in the case of `from_inference` as that may produce the `data` object I mentioned in a few comments. https://supervision.roboflow.com/develop/how_to/detect_small_objects/

Let me know if you have any questions.

---

Hey @SkalskiP, as mentioned in one of the comments - you were right about `data` missing in `__setitem__`.

The broader question is: what's contained inside it? Do we have a list of possible dict entries? We would want to unify into a single object, so it's a different 'merge' than what we had before in `Detections.merge`.

supervision/detection/core.py
```diff
@@ -59,6 +67,27 @@ def _validate_tracker_id(tracker_id: Any, n: int) -> None:
...
+    if pred1.tracker_id and pred2.tracker_id:
+        merged_tracker_id = get_merged_tracker_id(pred1.tracker_id, pred2.tracker_id)
+
+    return Detections(
+        xyxy=merged_bbox,
+        mask=merged_mask,
+        confidence=merged_conf,
+        class_id=merged_class_id,
+        tracker_id=merged_tracker_id,
+    )
```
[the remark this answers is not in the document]
mario-dg: Like @SkalskiP said, the PR is pretty old, hence the data field is missing.

supervision/detection/utils.py
```diff
@@ -110,6 +110,196 @@ def non_max_suppression(
     return keep[sort_index.argsort()]
 
 
+def greedy_nmm(predictions: np.ndarray, threshold: float = 0.5) -> Dict[int, List[int]]:
```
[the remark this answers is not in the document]
mario-dg: I get that, keep the naming more consistent, right?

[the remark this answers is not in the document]
LinasKo: Yup. And slightly fewer acronyms.
Most people haven't heard of `nmm` so `non_max_merge` is slightly more approachable.

mario-dg:
Thank you @LinasKo for this in depth review. I think I got most of what you've commented.
I estimate the changes you've requested to be done this work week.

SkalskiP:
> Hey @SkalskiP, as mentioned in one of the comments - you were right about `data` missing in `__setitem__`.
> 
> The broader question is: what's contained inside it? Do we have a list of possible dict entries? We would want to unify into a single object, so it's a different 'merge' than what we had before in `Detections.merge`.

The whole point of the `data` field is to give people flexibility. The `data` can contain anything. It seems to me that the most challenging task will involve linking two or more objects together. At the moment, I'm leaning towards copying the data from the detection with the higher confidence value. @LinasKo, what do you think?

supervision/detection/utils.py
SkalskiP: Yup, good idea!

LinasKo:
> The whole point of the `data` field is to give people flexibility. The `data` can contain anything. It seems to me that the most challenging task will involve linking two or more objects together. At the moment, I'm leaning towards copying the data from the detection with the higher confidence value. @LinasKo, what do you think?

Got it; then there's no ideal solution. In light of that - more confidence is a nice tie-breaker.

@mario-dg, can I request a change? If we're relying on confidence as a tie-breaker, let's make the `class_id` and `tracker_id` resolution use the same logic - instead of using `min` and `max`, let's pick the IDs from the more confident of the two merged objects.

supervision/detection/utils.py
LinasKo: Update: let's use the ID from the most confident object being merged, as mentioned in https://github.com/roboflow/supervision/pull/500#issuecomment-2046649898

supervision/detection/utils.py
LinasKo: Update: let's use the ID from the most confident object being merged, as mentioned in https://github.com/roboflow/supervision/pull/500#issuecomment-2046649898

mario-dg:
@LinasKo yes sure! Was about to start with your first comments, so I will keep that in mind and use the confidences as a base to determine which ids will be picked in the merging process.

supervision/detection/utils.py
mario-dg: I don't quite get this. Where does your 'area' come from?
Should the merged confidence still be `mean` or `max`, since we are using the more confident Detection to determine the `data`, `class_id` and `tracker_id`?

supervision/detection/tools/inference_slicer.py
LinasKo: For us, continuing with this PR would be best, even if it means removing some code.

At the same time, if you find it easier to make a new one or have started work already - that's fine too :)

supervision/detection/utils.py
LinasKo: Just to clarify - NMM doesn't give a number to say how much each area contributed to the new merged object, right? If it did, we could multiply that with confidences and get a normalized new confidence.

If it doesn't have that, I'll make a judgement call - let's use `mean`. Yes, this is inconsistent with how we get `data`and `ids`, but it's better than `max` which is equivalent to saying that we're just as confident about the newly merged area as we were about any of its parts.

mario-dg: Got it!
We do have the bounding boxes for each detection pair that is getting merged.
Using the merged bounding box, I could get the relative area each individual bounding box occupies in the merged bounding box to calculate the contribution of each object.
Does that sound like a solution or should we just stick to `mean`?

supervision/detection/core.py
```diff
@@ -660,6 +689,38 @@ def get_anchor_coordinates(self, anchor: Position) -> np.ndarray:
 
         raise ValueError(f"{anchor} is not supported.")
 
+    def __setitem__(
```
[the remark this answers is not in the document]
mario-dg: `__setitem__` is implemented to work with the `data` field, hence there is already a (partial) implementation.
But I also agree with you, resulting in me removing my `__setitem__` from the code.
This also solves the conversation below.

LinasKo:
Hey @mario-dg,

Good work! I'm reviewing it now.

For testing, you might find something useful in [this Colab](https://github.com/roboflow/supervision/pull/1108#issuecomment-2050206052). If you remove the inference slicer and its callbacks, you should be able to evaluate the runtime of your functions on an image with many realistic detections.

I suggest you also remove surplus tests - the Ultralytics runs quickest and is fairly easy to install. Anything with `inference` and `timm` wouldn't be needed.

LinasKo:
Other than that, I like how this looks.
It would be cool to see how fast it runs.

LinasKo:
Upon closer inspection, `mypy` revealed issues relating to datatypes.

Specifically, many of the variables we have such as `class_id` and `tracker_id` may be `None`. Also, in the output we expect numpy arrays, whereas the code passes in integer and float types.

We'd like to merge today, so I'll see if I can fix it.

mario-dg:
Hey @LinasKo, sorry for being absent lately. Due to being pretty busy with work and my masters thesis, I'm currently not able to test or benchmark.

LinasKo:
It's alright; we've all been through that :wink:

Thank you very much for your contribution. I'll make sure it gets verified and merged soon!

LinasKo:
@SkalskiP, Ready to review.

I've closed prior PR, and will include InferenceSlicer integration in a separate one.

Tests on 2 images, benchmarks. no videos or segmentation
https://colab.research.google.com/drive/1v0MPlG1tQctX5-koh0l6h1NcB6eWJ_YY?usp=sharing

I haven't tested with videos, and segmentations are known to fail. Will fix that with `Detections.merge` fixes this week anyway.

supervision/detection/utils.py
```diff
@@ -274,6 +274,137 @@ def box_non_max_suppression(
...
+        curr_keep_to_merge_list = non_max_merge(predictions[curr_indices], threshold)
+        curr_indices_list = curr_indices.tolist()
+        for curr_keep, curr_merge_list in curr_keep_to_merge_list.items():
+            keep = curr_indices_list[curr_keep]
+            merge_list = [curr_indices_list[i] for i in curr_merge_list]
+            keep_to_merge_list[keep] = merge_list
+    return keep_to_merge_list
+
+
+def get_merged_bbox(bbox1: np.ndarray, bbox2: np.ndarray) -> np.ndarray:
```
SkalskiP, line 387: We have `clip_boxes` and `pad_boxes` functions. Let's rename this to `merge_boxes`.

SkalskiP: If possible, update typing from simple `np.ndarray` to `npt.NDArray[np.float64]`.

SkalskiP: We also usually use `xyxy` naming convension, so let's rename arguments to `xyxy1` and `xyxy2`?

supervision/detection/utils.py
```diff
@@ -274,6 +274,137 @@ def box_non_max_suppression(
...
+            the format `(x_min, y_min, x_max, y_max)`.
+        bbox2 (np.ndarray): A numpy array of shape `(, 4)` where the
+            row corresponds to a bounding box in
+            the format `(x_min, y_min, x_max, y_max)`.
+
+    Returns:
+        np.ndarray: A numpy array of shape `(, 4)` where the new
+            bounding box is the merged bounding box of `bbox1` and `bbox2`.
+    """
+    left_top = np.minimum(bbox1[0][:2], bbox2[0][:2])
```
SkalskiP, line 403: Would vectorizing that function and merging multiple boxes simultaneously be a big lift? It comes down to replacing `0` with `:`.

supervision/detection/utils.py
SkalskiP, line 387: Please add some unit tests )

supervision/detection/utils.py
```diff
@@ -274,6 +274,137 @@ def box_non_max_suppression(
     return keep[sort_index.argsort()]
 
 
+def non_max_merge(
```
SkalskiP, line 277: We have `box_non_max_suppression`. Let's rename it to `box_non_max_merge`.

supervision/detection/core.py
```diff
@@ -1150,3 +1180,160 @@ def with_nms(
...
+                    merged_detection = self._merge_object_detection_pair(
+                        self[keep_ind], self[merge_ind]
+                    )
+                    self._set_at_index(keep_ind, merged_detection)
+            result.append(self[keep_ind])
+
+        return Detections.merge(result)
+
+    @staticmethod
+    def _merge_object_detection_pair(det1: Detections, det2: Detections) -> Detections:
```
SkalskiP, line 1245: Please don't add logic like this directly into `Detections` class. This class it too much already and it's getting harder and harder to use. Let's move that logic somewhere else, `detections/utils.py` for example.

SkalskiP: I would also love to see some unit tests here :)

SkalskiP (changes requested):

LinasKo:
Hey @SkalskiP, I pushed an update; here's some questions:

1. Removed `non_max_merge` altogether - it wasn't used.
1. There's now tests for `merge_object_detection_pair` and `box_non_max_merge`, but not for `batch_box_non_max_merge`. Would you like to see those too?
2. `merge_object_detection_pair`, I kept in detections/core.py, as due to it using `Detections`, it'd be a circular dependency if moved to `utils`. Alternatively, It would work if I annotate with `"Detections"` types without importing.
3. `with_nmm` is between 3x and `1x slower than `with_nms`.
   * For 1000 iterations & 38 detections, it's 12x slower. (3.9s vs 48s total)
   * For 1000 iterations & 758 detections, it was 3x slower in prior tests. (412s vs (still running, will update))
   * (1-iteration runs to be added, but they won't be better)
   * We spoke of optimization for this previously, but in this PR I paid little care for vectorization. Would you like me to speed this up?
4. I wanted to check with you again regarding datatypes. Do we want the change to these functions or every detections/utils method? that's 117 changes.
5. `box_non_max_merge` tie-breaks preferring later detections. E.g. given equal confidence, and sufficient overlap, we would merge detections 0 and 1 into detection 2. I think it's too small to care about - do we agree on that?

SkalskiP:
> Removed `non_max_merge` altogether - it wasn't used.

Awesome! Less is more!

> There's now tests for `merge_object_detection_pair` and `box_non_max_merge`, but not for `batch_box_non_max_merge`. Would you like to see those too?

No. That's okey.

LinasKo:
Thanks, that's helpful :)

* I've now realized that for the method we planned to move to utils, annotating with `"Detections"`. Is not enough. It needs to know about the `Detections` class as it calls the constructor.
  * I strongly suggest we keep it in `core.py`. Otherwise it's either a circular import, or we need to pass the class as an argument which adds complication.
  * Is there a third, non-complex option?
* Indeed, I was unclear when I asked about datatypes. What I meant was `npt.NDArray[np.float64]`. We can change the type annotations only in code introduced by this PR, or we can change it in the whole `core.py` and `utils.py` files. Alternatively, we can wait for @onuralpszr to push his changes ([Issue 1021](https://github.com/roboflow/supervision/issues/1021)), and then do a second pass, annotating every `np.ndarray` that's left (We'll need to do that anyway - some changes will have been made on other branches).

LinasKo:
Added `np.NDArray` types for functions affected by this change.

I believe I've addressed the change requests.

supervision/detection/core.py
```diff
@@ -1066,6 +1068,33 @@ def __setitem__(self, key: str, value: Union[np.ndarray, List]):
 
         self.data[key] = value
 
+    def _set_at_index(self, index: int, other: Detections):
```
SkalskiP, line 1071: Wouldn't placing this code as part of the __setitem__ method makes more sense? The flow below feels quite natural to me. 

```python
detections_1 = sv.Detections(...)
detections_2 = sv.Detections(...)
detections_1[0] = detections_2[0]
```

supervision/detection/core.py
```diff
@@ -1150,3 +1179,154 @@ def with_nms(
...
+    area_det1 = (det1.xyxy[0][2] - det1.xyxy[0][0]) * (
+        det1.xyxy[0][3] - det1.xyxy[0][1]
+    )
+    area_det2 = (det2.xyxy[0][2] - det2.xyxy[0][0]) * (
+        det2.xyxy[0][3] - det2.xyxy[0][1]
+    )
+
+    merged_x1, merged_y1 = np.minimum(det1.xyxy[0][:2], det2.xyxy[0][:2])
+    merged_x2, merged_y2 = np.maximum(det1.xyxy[0][2:], det2.xyxy[0][2:])
+    merged_xy = np.array([[merged_x1, merged_y1, merged_x2, merged_y2]])
```
SkalskiP, line 1306: I guess that one should be named `merged_xyxy`.

supervision/detection/core.py
```diff
@@ -1150,3 +1179,154 @@ def with_nms(
...
+
+    if det2.mask is None or det1.mask is None:
+        merged_mask = winning_det.mask
```
SkalskiP, line 1309: Curious about this line. What if `winning_det` does not have the mask, but the other one does?

supervision/detection/core.py
```diff
@@ -1150,3 +1179,154 @@ def with_nms(
...
+                merged_detection = merge_object_detection_pair(
+                    self[keep_ind], self[merge_ind]
+                )
+                self._set_at_index(keep_ind, merged_detection)
+            result.append(self[keep_ind])
+
+        return Detections.merge(result)
+
+
+def merge_object_detection_pair(det1: Detections, det2: Detections) -> Detections:
```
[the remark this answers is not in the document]
SkalskiP: In general, I think we have a naming problem. Our current `marge` should be called `concatenate`, and this should be just `merge.`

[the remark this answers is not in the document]
SkalskiP: Should we allow for merging detections that have different "non-None" fields?

supervision/detection/utils.py
```diff
@@ -274,6 +275,81 @@ def box_non_max_suppression(
     return keep[sort_index.argsort()]
 
 
+def box_non_max_merge(
```
SkalskiP, line 278: This one seems like a helper function that does not need to be exposed in the API.

supervision/detection/core.py
```diff
@@ -1150,3 +1179,154 @@ def with_nms(
...
+            predictions = np.hstack(
+                (
+                    self.xyxy,
+                    self.confidence.reshape(-1, 1),
+                    self.class_id.reshape(-1, 1),
+                )
+            )
+            keep_to_merge_list = box_non_max_merge_batch(predictions, threshold)
+
+        result = []
```
SkalskiP, line 1230: It feels to me like the whole logic here

```
result = []
for keep_ind, merge_ind_list in keep_to_merge_list.items():
    for merge_ind in merge_ind_list:
        merged_detection = merge_object_detection_pair(
            self[keep_ind], self[merge_ind]
        )
        self._set_at_index(keep_ind, merged_detection)
    result.append(self[keep_ind])

return Detections.merge(result)
```

It should be hidden inside `box_non_max_merge.`

SkalskiP (changes requested):

supervision/detection/core.py
SkalskiP: `__setitam__`

```
detections_2[0]
detections_2["class_name"]
```

`__getitam__`

```
detections_2[0]
detections_2[1:3]
detections_2[[1, 2, 3]]
detections_2[[False, True, False]]
detections_2["class_name"]
```

supervision/detection/core.py
```diff
@@ -1150,3 +1179,154 @@ def with_nms(
...
+        detections = sv.Detections.from_inference(result)
+
+        merged_detections = merge_object_detection_pair(
+            detections[0], detections[1])
+        ```
+    """
+    if len(det1) != 1 or len(det2) != 1:
+        raise ValueError("Both Detections should have exactly 1 detected object.")
+
+    if det2.confidence is None:
```
SkalskiP, line 1288: Let's just assume the field value in both detections is `None` or the field value in both detections is not `None`. No mixed scenarios.

supervision/__init__.py
```diff
@@ -44,6 +44,8 @@
 from supervision.detection.tools.smoother import DetectionsSmoother
 from supervision.detection.utils import (
     box_iou_batch,
+    box_non_max_merge,
```
SkalskiP, line 47: Let's not expose `box_non_max_merge` in public API.

supervision/detection/utils.py
```diff
@@ -274,6 +275,81 @@ def box_non_max_suppression(
...
+        above_threshold = ious >= iou_threshold
+        keep_to_merge_list[idx] = np.flip(order[above_threshold]).tolist()
+        order = order[~above_threshold]
+
+    return keep_to_merge_list
+
+
+def box_non_max_merge_batch(
+    predictions: npt.NDArray[np.float64], iou_threshold: float = 0.5
+) -> Dict[int, List[int]]:
```
SkalskiP, line 322: I think this function should return `List[List[int]]`. Essentially define groups of detections that should be merged together.

supervision/detection/utils.py
```diff
@@ -274,6 +275,81 @@ def box_non_max_suppression(
...
+
+        above_threshold = ious >= iou_threshold
+        keep_to_merge_list[idx] = np.flip(order[above_threshold]).tolist()
+        order = order[~above_threshold]
+
+    return keep_to_merge_list
+
+
+def box_non_max_merge_batch(
+    predictions: npt.NDArray[np.float64], iou_threshold: float = 0.5
```
SkalskiP, line 321: Let's mimic the behavior of `mask_non_max_suppression` and `box_non_max_suppression`. If predictions is `(n, 6)`, shape `box_non_max_merge_batch` is class-based, and if it is `(n, 5)`, it is class agnostic.

SkalskiP (changes requested):

supervision/detection/utils.py
LinasKo: renamed to `_box_non_max_merge_all`, removed from __init__.py.

Note: you'll see `box_non_max_merge` which is the prior batch function. Now it's the main one.

supervision/detection/core.py
LinasKo: _set_at_index was not required. I removed it entirely, and did not add any logic to __setitem__.

supervision/detection/core.py
```diff
@@ -1150,3 +1152,203 @@ def with_nms(
...
+    """
+    Verify that for each optional field in the Detections, both instances either have
+    the field set to None or both have it set to non-None values.
+
+    `data` field is ignored.
+
+    Raises:
+        ValueError: If one field is None and the other is not, for any of the fields.
+    """
+    attributes = ["mask", "confidence", "class_id", "tracker_id"]
```
SkalskiP, line 1345: Can we try to get that list automatically?

supervision/detection/core.py
```diff
@@ -1150,3 +1152,203 @@ def with_nms(
...
+    detections_1 = detections[0]
+    for detections_2 in detections[1:]:
+        box_iou = box_iou_batch(detections_1.xyxy, detections_2.xyxy)[0]
+        if box_iou < threshold:
+            break
+        detections_1 = _merge_inner_detection_object_pair(detections_1, detections_2)
+    return detections_1
+
+
+def _verify_fields_both_defined_or_none(
```
SkalskiP, line 1333: I'd use `validate`.

supervision/detection/core.py
```diff
@@ -1150,3 +1152,203 @@ def with_nms(
...
+    """
+    if len(detections_1) != 1 or len(detections_2) != 1:
+        raise ValueError("Both Detections should have exactly 1 detected object.")
+
+    _verify_fields_both_defined_or_none(detections_1, detections_2)
+
+    if detections_1.confidence is None and detections_2.confidence is None:
+        merged_confidence = None
+    else:
+        area_det1 = (detections_1.xyxy[0][2] - detections_1.xyxy[0][0]) * (
```
SkalskiP, line 1267: no abbreviations :)

supervision/detection/core.py
```diff
@@ -1150,3 +1152,203 @@ def with_nms(
...
+    )
+    merged_xyxy = np.array([[merged_x1, merged_y1, merged_x2, merged_y2]])
+
+    if detections_1.mask is None and detections_2.mask is None:
+        merged_mask = None
+    else:
+        merged_mask = np.logical_or(detections_1.mask, detections_2.mask)
+
+    if detections_1.confidence is None and detections_2.confidence is None:
+        winning_det = detections_1
```
SkalskiP, line 1293: No abbreviations :)

supervision/detection/core.py
SkalskiP: Think about moving `detections_1.xyxy[0]` to a separate helper variable.

supervision/detection/core.py
```diff
@@ -1150,3 +1152,203 @@ def with_nms(
...
+    elif detections_1.confidence[0] >= detections_2.confidence[0]:
+        winning_det = detections_1
+    else:
+        winning_det = detections_2
+
+    winning_class_id = winning_det.class_id
+    winning_tracker_id = winning_det.tracker_id
+    winning_data = winning_det.data
```
SkalskiP, line 1301: Think about dropping those extra variables. :)

supervision/detection/core.py
```diff
@@ -1150,3 +1152,203 @@ def with_nms(
...
+                )
+            )
+
+        merge_groups = box_non_max_merge(
+            predictions=predictions, iou_threshold=threshold
+        )
+
+        result = []
+        for merge_group in merge_groups:
+            unmerged_detections = [self[i] for i in merge_group]
```
SkalskiP, line 1205: Maybe we don't need that list comprehension, just use `detections[indexes]`.

LinasKo: My explanation was wrong.

We're doing this not to copy the result (that's in another case), but to create a list of single-object detections. `[Detections, Detections, Detections, ...]`.

I believe this is the most concise way.

supervision/detection/core.py
LinasKo: I tried it, it's cumbersome, I'll add the code + tests in a separate PR and we can choose whether to keep it.

supervision/detection/core.py
LinasKo: Dropped

supervision/detection/core.py
LinasKo: Expanded

supervision/detection/core.py
LinasKo: Create a new variable

supervision/detection/core.py
LinasKo: Now called `validate_fields_both_defined_or_none` - removed `_` too.

LinasKo:
Hi @SkalskiP,

I've tidied this up - I believe it can now be merged.

SkalskiP (approved):

SkalskiP:
Thanks a lot, @LinasKo and @mario-dg! 🙏🏻 It's merged!

mario-dg:
That's awesome! Didn't image that this PR would turn out as big as it got. Thanks guys!🚀

PR #516, opened 2023-10-23
Add process video example

capjamesg:
This PR adds the `Process Video` "how to" to the supervision documentation.

### DIFF
--- a/docs/how_to/process_video.md
+++ b/docs/how_to/process_video.md
@@ -30,10 +30,10 @@ def callback(scene: np.ndarray, index: int) -> np.ndarray:
 
     annotated_image = bounding_box_annotator.annotate(
         scene=scene, detections=detections)
-    
+
     annotated_image = label_annotator.annotate(
         scene=annotated_image, detections=detections, labels=labels)
-    
+
     return annotated_image
 ```
 
--- a/docs/how_to/process_video.md
+++ b/docs/how_to/process_video.md
@@ -1,1 +1,74 @@
-🚧 Page under construction.
+With supervision, you can run inference on or otherwise process video frames with ease. In this guide, we will show how to run inference on each frame in a video and save the results to a new file.
+
+To process video, we will use the [sv.process_video](https://supervision.roboflow.com/utils/video/#process_video) method.
+
+## Define a Callback Function
+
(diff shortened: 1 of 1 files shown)


docs/how_to/process_video.md
```diff
@@ -1 +1,74 @@
...
+    detections = sv.Detections.from_ultralytics(results)
+
+    bounding_box_annotator = sv.BoundingBoxAnnotator()
+    label_annotator = sv.LabelAnnotator()
+
+    labels = [
+        results.names[class_id]
+        for class_id
+        in detections.class_id
+    ]
```
LinasKo, line 29: Let's let the annotator resolve the names itself. If a user plugs in a model without class_names, it will work just as well.

Here's an example from another file:

```
bounding_box_annotator = sv.BoundingBoxAnnotator()

annotated_image = bounding_box_annotator.annotate(
annotated_image = label_annotator.annotate(
    scene=annotated_image, detections=detections)
```

It auto-determines if there's a class name and displays it.

SkalskiP:
@capjamesg and @LinasKo 👋🏻, do you think we still need this if we have https://supervision.roboflow.com/latest/how_to/track_objects/? It seems to me that this tutorial covers all of those points. I don't want to have too many `how-to` guides, as users may get lost.

That's why I'd recommend closing it.

LinasKo:
Hi @SkalskiP ,

I had the same thought, but here's my thinking: as a type of user, I want to detect stuff in a video. I don't want to track things, and I don't want to read through every how-to until I find which one contains the basic example I need.

I do not think that "track objects" evokes a thought about "video" automatically.

A solution would be to include video processing in all, or the most popular articles, or change the title of tracking to mention "video".

---

In fact, the info about videos is pretty rare, so I've only added `sv.process_video` to my mental toolbox because of this page. Same as with inference, it feels like there's multiple ways to achieve similar goals - sinks, process_video, frame gen, so what would help is having a tradeoff and recommended use case page.

Saving time, instead I'd rather merge this.

SkalskiP:
Than I'd just rename the `How to` guide to `Detect and Track Objects on Video`. Trust me we don't want to have two guides that are 70-80% the same.

LinasKo:
If the goal is less duplication, let's close this one.

Renaming title of How To section to mention video: https://github.com/roboflow/supervision/pull/1223

PR #556, opened 2023-10-31
build(remove): 🔥 remove opencv headless install method

onuralpszr:
### Description

I see many issues cause of minimal vs full and certain deployments has conflicts between opencv and headless opencv so for that reason. I propose to remove headless install method. I also deleted related docs as well. 


SkalskiP:
Hi @onuralpszr 👋🏻 Any how we plan to support `headless` install if not this way?

onuralpszr:
> Hi @onuralpszr 👋🏻 Any how we plan to support `headless` install if not this way?

no dep install supervision and then install all deps via manually is one of them.

onuralpszr:
> Hi @onuralpszr 👋🏻 Any how we plan to support `headless` install if not this way?

Btw I opened this for discussion and decisions but I was also observed in repo many issue because of this.

SkalskiP:
> no dep install supervision and then install all deps via manually is one of them.

Curious if there are other alternatives 🤔 💭

onuralpszr:
> > no dep install supervision and then install all deps via manually is one of them.
> 
> Curious if there are other alternatives 🤔 💭

My problem is I can't split both packages with one default package I have to make them both optional which you don't want that.

onuralpszr:
> I second this PR. I would like to install supervision without opencv-python-headless library, which does not seem possible at the moment.

It is possible If you want to install without headless. You can install via --no-deps" parameter to ignore all deps and install deps yourself. Or you can install normally when you can swap by removing headless and install opencv-python  and it will work. If you have different packages does require opencv-python already you can install everything and then you can do swapping at last. We are going to re-discuss this headless change again later after release for now that's the options you can do. I hope that will help you out. 

Thank you.

ozangungortuhh:
Thanks a lot @onuralpszr for the answer. At the moment I added `pip uninstall opencv-python-headless` at the end of my Dockerfile and it's fine, what I meant was that there is no installation option for `supervision` without installing `opencv-python-headless` package at all.

SkalskiP:
@onuralpszr, do you have any idea how to distribute supervision in both models? or do we intend to just drop headless mode?

onuralpszr:
> @onuralpszr, do you have any idea how to distribute supervision in both models? or do we intend to just drop headless mode?

I might be have an idea let me experiment and will tell you how based on that we can decide it.

SkalskiP:
> I might be have an idea let me experiment and will tell you how. After that we can decide how to proceed it.

Unfortunately, I don't have time to discuss this today. But I'd be happy to listen next week. We can even schedule a call.

onuralpszr:
https://github.com/roboflow/supervision/pull/1530 merged and close this merge

Issue #460, opened 2023-10-12
[IconAnnotator] - mark objects with custom icons / images


SkalskiP:
### Description

Create `IconAnnotator` - whose task is to mark objects related to [`sv.Detections`](https://supervision.roboflow.com/detection/core/#detections) with custom icons.

![273726799-6b0ea345-bb22-4c3d-9e67-6a91371f852e](https://github.com/roboflow/supervision/assets/26109316/2f184480-c424-4724-a494-189d59592d13)

It should also allow you to control the icon position in relation to the box using [`sv.Position`](https://supervision.roboflow.com/geometry/core/#position).

### API

```python
class IconAnnotator(BaseAnnotator):

    def __init__(
        self,
        icon_path: str,
        position: Position = Position.TOP_CENTER
    ):
        pass

    def annotate(
        self,
        scene: np.ndarray,
        detections: Detections,
    ) -> np.ndarray:
        pass
```

### Additional

- Note: Please share a Google Colab with minimal code to test the new feature. We know it's additional work, but it will definitely speed up the review process. Each change must be tested by the reviewer. Setting up a local environment to do this is time-consuming. Please ensure that Google Colab can be accessed without any issues (make it public). Thank you! 🙏🏻

Rajarshi-Misra:
@SkalskiP I'm interested in working on this one.

SkalskiP:
@Rajarshi-Misra you are assigned!

Shishuii:
@SkalskiP Is this still available ? I would be glad to contribute to this. :)

onuralpszr:
@Shishuii I am re-assigning to you. You can start working on it.

Shishuii:
Hi @onuralpszr , 
Below is the final output of IconAnnotator.
Shall I submit a PR for review?
![final_ouput_v2](https://github.com/roboflow/supervision/assets/22843318/426a70d3-5804-4161-b3c1-90052d13d807)

onuralpszr:
@Shishuii looking nice do please PR let's talk about on PR as well.

Shishuii:
Hi @onuralpszr , 
Please review the PR.
The PR can be tested on this colab link :
https://colab.research.google.com/drive/1rjwO76QwA-W5WhA0Tvy3uSFtmCtbd9rA?usp=sharing

dfreilich:
@SkalskiP I'd like to take this (based on the work in https://github.com/roboflow/supervision/pull/579).

[...]

PR #579, opened 2023-11-06
Add icon annotator

Shishuii:
# Description

This change address #460 

## Type of change

Tested locally , Also link for the colab is [link](https://colab.research.google.com/drive/1rjwO76QwA-W5WhA0Tvy3uSFtmCtbd9rA?usp=sharing)

Let me know if any changes are required

### DIFF
--- a/supervision/__init__.py
+++ b/supervision/__init__.py
@@ -16,6 +16,7 @@ from supervision.annotators.core import (
     EllipseAnnotator,
     HaloAnnotator,
     HeatMapAnnotator,
+    IconAnnotator,
     LabelAnnotator,
     MaskAnnotator,
     TraceAnnotator,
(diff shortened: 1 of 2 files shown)


onuralpszr:
@Shishuii hello and here is my thoughts, I think we should remove "color" and remove "background clean" and put as it is. Because we can put actual prepared png or png-list to handle it. I also notice my weirdness on background remove on certain different pics so it may create different issues as well. So please re-check those as well. 

Thank you for your hard work !

Shishuii:
Sure @onuralpszr , I will work on these and let you know.

LinasKo:
Closing, implemented via #930.

Thank you for the contribution, @Shishuii. While dating to before my time in Roboflow, as far as I can tell, `IconAnnotator` had a bit of a shaky history regarding us reviewing it. Still, without your PR, we may not have built it.

PR #630, opened 2023-11-29
Better Polygons from Masks

thomasf1:
# Description

Improved the mask_to_polygons function to handle masks that have exclusions and Masks that do not align

Example:
Detections of Doughnut: Even if the Model correctly detects the Doughnut, the Polygon always was only the outer shape. With this change the hole in the middle is excluded from the Polygon
Person behind a table: When the Model identifies the Person correctly, the Part above and below the Table would have been two separate Polygons. With this change, it would be one Polygon connected.

## Type of change

-   [X] Somewhere in between between Bug fix and New feature (non-breaking change which adds functionality)
-> Arguably brings the Polygon functions closer to expected behavior

Example:

Mask Annotator: (unchanged)
person_mask

Polygon Conversion before PR:
person_polygon

Polygon Conversion after PR:
person_polygon2

Also changed the Polygon Annotator to display a filled Polygon

## Docs

The original issue is not Documented. If the change to draw_polygon is accepted as well, the documentation for the function needs to be updated.

### DIFF
--- a/supervision/detection/utils.py
+++ b/supervision/detection/utils.py
@@ -240,7 +240,6 @@ def mask_to_polygons(mask: np.ndarray) -> List[np.ndarray]:
     )
 
     contours_approx = []
-    polygons = []
     for contour in contours:
         epsilon = 0.001 * cv2.arcLength(contour, True)
         contour_approx = cv2.approxPolyDP(contour, epsilon, True)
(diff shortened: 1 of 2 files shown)


SkalskiP:
Hi, @thomasf1! 👋🏻 Could you accept CLA?

PR #634, opened 2023-12-04
Add `ColorAnnotator`

capjamesg:
# Description

This PR adds the `ColorAnnotator` class to the `supervision` annotator suite.

Example:

Screenshot 2023-12-04 at 14 15 22

## Type of change

Use the following Notebook to test the feature:

https://colab.research.google.com/drive/1X6KDQB2mw52JBDBibowU7cREA82bFZxs?usp=sharing

N/A

## Docs

### DIFF
--- a/docs/annotators.md
+++ b/docs/annotators.md
@@ -229,6 +229,27 @@
 
     </div>
 
+=== "Color"
+
+    ```python
+    >>> import supervision as sv
+
+    >>> image = ...
+    >>> detections = sv.Detections(...)
+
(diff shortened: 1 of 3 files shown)


SkalskiP:
After some brainstorming, we decided to close this PR as you can already get the same effect with the supervision `BoxMaskAnnotator`. Just set the opacity to 1. But we'll rename `BoxMaskAnnotator` to `ColorAnnotator` as this name seems more accurate. This change will be done in a separate PR.

PR #658, opened 2023-12-07
Remove upper limit on Python version

sanmai-NL:
# Description

See: https://discuss.python.org/t/requires-python-upper-limits/12663

The current upper limit breaks packages that depend on `supervision` and claim compatibility with a wider range of Python versions, like CPython 3.12.:

> 🔒 Lock failed
> Unable to find a resolution because the following dependencies don't work on all Python versions in the range of the project's `requires-python`: >=3.11,<3.13.
>   python>=3.8,<3.12 (from <Candidate supervision@0.16.0 from https://pypi.org/simple/supervision/>)
>   python>=3.8,<3.12 (from <Candidate supervision@0.17.0 from https://pypi.org/simple/supervision/>)
> A possible solution is to change the value of `requires-python` in pyproject.toml to >=3.11,<3.12.

However, these packages must sometimes set a wider range as on some Linux distributions, like Fedora, CPython 3.12 is the standard, and the package itself would refuse to install if `requires-python` were to satisfy `<3.12`.

Since `supervision` is actively maintained, should some future Python version be explictly unsupported, the upper limit can be set again. But then it would be informed and deliberate, rather than avoiding unknown modern Python versions.

It will be tested in CI.

## Docs

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ classifiers=[
 ]
 
 [tool.poetry.dependencies]
-python = ">=3.8,<3.12.0"
+python = ">=3.8"
 numpy = ">=1.21.2"
 matplotlib = ">=3.5.0"
 pyyaml = ">=5.3,<=6.0.1"


SkalskiP:
Hi @sanmai-NL 👋🏻 

We will solve the same problem with different PRs. Add support for Python `3.12` and release `supervision-0.17.1` soon.

sanmai-NL:
@onuralpszr and @SkalskiP, You use Poetry. The Poetry team advises this:

> That all goes to say, that best practices are likely this:
> 
> 1. do not constrain Requires-Python based on a test matrix, but instead hard incompatibility only (e.g. Python extension ABI incompatibility)
> 2. Poetry must still solve on a limited range of Pythons, but those should not drive Requires-Python

https://github.com/python-poetry/poetry/issues/6921#issuecomment-1308973524

I understand you want to test against new Python versions, but the way you now use this `requires-python` metadata field goes too far.

onuralpszr:
> @onuralpszr and @SkalskiP, you use Poetry. The Poetry team advises this:
> 
> > That all goes to say, that best practices are likely this:
> > 
> > 1. do not constrain Requires-Python based on a test matrix, but instead hard incompatibility only (e.g. Python extension ABI incompatibility)
> > 2. Poetry must still solve on a limited range of Pythons, but those should not drive Requires-Python
> 
> [python-poetry/poetry#6921 (comment)](https://github.com/python-poetry/poetry/issues/6921#issuecomment-1308973524)
> 
> I understand you want to test against new Python versions, but the way you now use this `requires-python` metadata field goes too far.


Based on poetry practice I also notice that I could of use single version style packages in metada by removal of upper limit version of python. I also discussed with Piotr and we don't want to artificially limit people with versions. I changed my PR accordingly. Now We have Python 3.8 is min version and anything above should work as well. But in metadata we add  "3.12" since it is released.  We can add 3.13 once we also tested in our test matrix as well.  Now it should support everything 3.8 and above as well. I added docs and CI changes as well. It should be fine.

onuralpszr:
> > > @onuralpszr and @SkalskiP, you use Poetry. The Poetry team advises this:
> > > > That all goes to say, that best practices are likely this:
> > > > 
> > > > 1. do not constrain Requires-Python based on a test matrix, but instead hard incompatibility only (e.g. Python extension ABI incompatibility)
> > > > 2. Poetry must still solve on a limited range of Pythons, but those should not drive Requires-Python
> > > 
> > > 
> > > [python-poetry/poetry#6921 (comment)](https://github.com/python-poetry/poetry/issues/6921#issuecomment-1308973524)
> > > I understand you want to test against new Python versions, but the way you now use this `requires-python` metadata field goes too far.
> > 
> > 
> > Based on poetry practice I also notice that I could of use single version style packages in metada by removal of upper limit version of python. I also discussed with Piotr and we don't want to artificially limit people with versions. I changed my PR accordingly. Now We have Python 3.8 is min version and anything above should work as well. But in metadata we add "3.12" since it is released. We can add 3.13 once we also tested in our test matrix as well. Now it should support everything 3.8 and above as well. I added docs and CI changes as well. It should be fine.
> 
> @onuralpszr Thanks for improving the design!
> 
> With metadata, do you specifically mean something like the [`classifiers` key](https://python-poetry.org/docs/pyproject/#classifiers), e.g.:
> 
> ```toml
> classifiers = [
>   "Programming Language :: Python :: 3.11",
> ]
> ```
> 
> ?
> 
> Regrettably, Poetry seems to derive these classifiers from the [`python` dependency](https://python-poetry.org/docs/pyproject/#dependencies-and-dependency-groups). In Poetry `pyproject.toml`:
> 
> ```
> [tool.poetry.dependencies]
> python = "^3.11"
> ```
> 
> In PEP 621 and PDM, there is the `requires-python` key for that. And you use it as well. So I hope that the classifiers will turn out all right in your PR. Can you inspect the sdist build output to be sure? If you have a PR ready, I could participate in reviewing it.

Sorry for late answer, I had to something in IRL and I did check and pushed yes it does work. I forgat that trick, thank you !! :) It does works nicely.

SkalskiP:
Hi @sanmai-NL and @onuralpszr! 👋🏻 We have collaboratively arrived at an improved solution. I just merged https://github.com/roboflow/supervision/pull/659 and released https://pypi.org/project/supervision/0.17.1rc1. Would be awesome if you could test if that version works for you, and if so, we will make `supervision-0.17.1` release.

PR #665, opened 2023-12-10
feat(xml-parsing): 🔐 enhance security with defusedxml and bandit rules

onuralpszr:
### Description

Based on analysis of Bandit,snyk we have security vulnerability around xml and for that python docs even recommend to use defusedxml 

### Links

https://docs.python.org/3/library/xml.html#module-xml
https://docs.python.org/3/library/xml.html#xml-vulnerabilities
https://docs.python.org/3/library/xml.html#the-defusedxml-package
https://github.com/tiran/defusedxml

You can also view vulnerable state in here  
https://www.codefactor.io/repository/github/roboflow/supervision/issues#

This PR fixes xml vulnerability and adding extra rules into bandit for make sure in future we don't effect by XML attacks in anycases.

fix bandit rule B318: (xml-attribute-parsing) defusedxml package should be used to parse untrusted XML data
fix bandit rule B314: (xml-entity-expansion) XML entity expansion should be disabled
fix bandit rule B413: (xml-etree-parse) XML parsing library should be secure
fix bandit rule B412: (xml-lib-exploits) XML libraries should be vetted for defusedxml or lxml.vulns
fix bandit rule B410: (xml-minidom-xxe) XML minidom can be vulnerable to XML eXternal Entity attacks

### DIFF
--- a/.github/requirements-dev-min.txt
+++ b/.github/requirements-dev-min.txt
@@ -25,3 +25,4 @@ tomli==2.0.1
 tqdm==4.62.3
 typing_extensions==4.8.0
 urllib3==1.26.18
+defusedxml == 0.7.1
--- a/poetry.lock
+++ b/poetry.lock
@@ -518,6 +518,64 @@ test = ["pytest"]
 
 [[package]]
 name = "contourpy"
+version = "1.1.0"
+description = "Python library for calculating contours of 2D quadrilateral grids"
+optional = false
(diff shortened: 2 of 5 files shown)


SkalskiP:
Thanks a lot for thinking about it, @onuralpszr 🙏🏻 Merging!

SkalskiP (approved):

PR #696, opened 2023-12-27
Add Prediction Smoothing

yeldarby:
# Description

Adds a `Smoother` class which averages the last several predictions to decrease noise (like wobbly or flickering boxes).

Before & after smoothing:

https://github.com/roboflow/supervision/assets/870796/160d4d53-51c2-4d82-a6d0-d655fc6eec73

## Type of change

Installed locally & tested on several videos/models with `InferencePipeline`.

No.

## Docs

-   [x] Docs updated? What were the changes: Added new docs page with video & example code.


supervision/detection/smoother.py
```diff
@@ -0,0 +1,172 @@
...
+    to track objects over time and averaging out the predictions over the
+    `length` most recent frames.
+
+    <video controls>
+        <source src="https://media.roboflow.com/supervision/video-examples/smoothed-grocery-example-720.mp4" type="video/mp4">
+    </video>
+    > _On the left are the model's raw predictions,
+    > on the right is the output of Smoother._
+
+    ## Example Usage:
```
SkalskiP, line 20: Can we show a more general example? I assume `Smoother` can work with any object detection model.

mkdocs.yml
```diff
@@ -37,6 +37,7 @@ nav:
         - Core: classification/core.md
     - Detections:
         - Core: detection/core.md
+        - Smoother: detection/smoother.md
```
SkalskiP, line 40: Please move `Smoother` into `Tools`

supervision/detection/smoother.py
```diff
@@ -0,0 +1,172 @@
...
+    )
+    pipeline.start()
+    pipeline.join()
+    ```
+
+    !!! warning
+
+        Smoother utilizes the `tracker_id`. Read
+        [here](https://supervision.roboflow.com/trackers/) to learn how to plug
+        tracking into your inference pipeline.
```
SkalskiP, line 71: We should probably add some information about behavior when running the segmentation model.

supervision/detection/smoother.py
```diff
@@ -0,0 +1,172 @@
...
+
+        self.current_frame = 0
+        self.tracks = NoneDict()
+        self.track_starts = NoneDict()
+        self.track_ends = NoneDict()
+
+    def tracker_length(self, tracker_id):
+        return self.current_frame - self.track_starts[tracker_id]
+
+    def add_frame(self, detections: Detections) -> None:
```
SkalskiP, line 90: I'd love to keep the current naming convention. Let's rename it to `update_with_detections`. We use the same name in [`ByteTrack`](https://supervision.roboflow.com/trackers/#supervision.tracker.byte_tracker.core.ByteTrack.update_with_detections). It is a very similar use case. We mutate the state based on the information stored in the `Detections` object.

supervision/detection/smoother.py
```diff
@@ -0,0 +1,172 @@
...
+    ```
+
+    !!! warning
+
+        Smoother utilizes the `tracker_id`. Read
+        [here](https://supervision.roboflow.com/trackers/) to learn how to plug
+        tracking into your inference pipeline.
+    """
+
+    def __init__(self, length: int = 5) -> None:
```
SkalskiP, line 74: Please provide some method to reset `Smoother` state.

supervision/detection/smoother.py
```diff
@@ -0,0 +1,172 @@
...
+        """
+
+        self.length = length
+
+        self.current_frame = 0
+        self.tracks = NoneDict()
+        self.track_starts = NoneDict()
+        self.track_ends = NoneDict()
+
+    def tracker_length(self, tracker_id):
```
SkalskiP, line 87: Please specify input and output types.

supervision/detection/smoother.py
```diff
@@ -0,0 +1,172 @@
...
+            if self.track_ends[track_id] < self.current_frame:
+                # continue tracking for a few frames after the object has left
+                # (to prevent flickering in case it comes back)
+                track.append(None)
+
+            if len(track) > self.length:
+                # remove the oldest detection from the track it's too long
+                track.pop(0)
+
+    def get_track(self, track_id):
```
SkalskiP, line 126: Please specify input and output types.

supervision/detection/smoother.py
```diff
@@ -0,0 +1,172 @@
...
+            return None
+
+        ret = track[0]
+        # set to an average of all the detection boxes
+        ret.xyxy = np.mean([d.xyxy for d in track], axis=0)
+        ret.confidence = np.mean([d.confidence for d in track], axis=0)
+
+        return ret
+
+    def get_smoothed_detections(self):
```
SkalskiP, line 142: Please specify input and output types.

supervision/detection/smoother.py
```diff
@@ -0,0 +1,172 @@
...
+            self.track_ends[tracker_id] = self.current_frame
+
+        for track_id in self.tracks:
+            track = self.tracks[track_id]
+            if self.track_ends[track_id] < self.current_frame:
+                # continue tracking for a few frames after the object has left
+                # (to prevent flickering in case it comes back)
+                track.append(None)
+
+            if len(track) > self.length:
```
SkalskiP, line 122: We can simplify this code using Python built-in data structures. Specifically [`deque`](https://docs.python.org/3/library/collections.html#collections.deque). This allows you to set the max length and takes care of poping the lat element automatically.

supervision/detection/smoother.py
```diff
@@ -0,0 +1,172 @@
...
+        tracked_detections = []
+        for track_id in self.tracks:
+            track = self.get_track(track_id)
+            if track is not None:
+                tracked_detections.append(track)
+
+        return Detections.merge(tracked_detections)
+
+
+class NoneDict(defaultdict):
```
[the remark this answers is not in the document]
SkalskiP: Update after [this](https://github.com/roboflow/supervision/pull/696/files#diff-cdfd1a9e16bbb56b7d072b59f3151a8b1e7d79ea8189120d1c592a48a62f124bR122) comment. You can also take care of those two problems by using `defaultdict` and `deque` together. This is probably best in your use case.

```python
self.tracks = defaultdict(lambda: deque(maxlen=length))
```

supervision/detection/smoother.py
```diff
@@ -0,0 +1,172 @@
...
+        track = self.tracks[track_id]
+        if track is None:
+            return None
+
+        track = [d for d in track if d is not None]
+        if len(track) == 0:
+            return None
+
+        ret = track[0]
+        # set to an average of all the detection boxes
```
SkalskiP, line 136: Please remove redundant comments.

supervision/detection/smoother.py
SkalskiP: Also, we don't need an extra method here. It can be just returned at the end of the `add_frame` call (hopefully renamed to `update_with_detections`).

SkalskiP (changes requested):

supervision/detection/smoother.py
yeldarby: Sure, what do you have in mind? I figured person detection with COCO using one of the videos from `assets` was about as general as it gets to demonstrate the functionality but we can change it to something else.

Good candidate videos to communicate the functionality would have
* A small number of objects moving fairly subtly (otherwise it's really distracting trying to understand what you're supposed to be comparing side by side)
* Prediction jitter coming from the model

For example, while it **works** with [vehicles-2.mp4](https://media.roboflow.com/supervision/video-examples/vehicles-2.mp4), it's not a good example video because there's too much going on and the things are moving too fast to be able to grok what's going on at the detection by detection level.

supervision/detection/smoother.py
yeldarby: What _should_ it do in that case? Right now it just deals with the xywh. 

Blending masks (averaging pixelwise confidences?) seems like good future functionality but is out of scope for the first release.

supervision/detection/smoother.py
yeldarby: If I annotate that, how do I mark this method as primarily for internal use of the class vs showing up in the API in the docs?

Dunder prefix seems to have other effects vs just hiding from the docs.

SkalskiP: For the method to appear in docs, you must fulfill two requirements: the method must be public (no `_` or `__` prefix, and docstring. I don't want you to add docstring I meant type hints.

supervision/detection/smoother.py
SkalskiP: Just mention it is intended for Detections that don't contain a `mask` field.

SkalskiP:
@yeldarby, please let me know once the PR is ready for the next review round.

yeldarby:
PR is updated in response to your comments here & @capjamesg's comments on Slack.

SkalskiP:
@yeldarby I started the second round of review, but I got confused and decided to ask here:
- Is `tracker_length` still needed? Because it is not used, it is not documented and generally unrelated to the Smoother task (smoothing boxes).
- If `tracker_length` is not needed, do we need `self.track_starts` and `self.track_ends` ? `self.track_starts` is only there to support `tracker_length` and `self.track_ends` is used in two places but the second place can be easily implemented differently.
- If we don't need `self.track_starts` and `self.track_ends` we probably do not need `self.current_frame`.

All around, the whole logic can get a lot easier if we drop `tracker_length`, which, as I said, is not used; it is not documented and, in general, is unrelated to the Smoother task.

yeldarby:
> Is tracker_length still needed?

I'm using it in some client code to create animated `Annotator`s. The frame of the animation for a detection needs to be `tracker_length%num_frames`.

Also planning to use it to let people choose if we should wait to display a detection until it's been seen some number of times so that if something is detected in a frame or two but not again it doesn't flicker onto the screen (but haven't implemented that here just yet).

> Because it is not used, it is not documented

Good call; if we keep it here I'll document it.

> and generally unrelated to the Smoother task (smoothing boxes).

The entry delay will be related to this (though that's easier since you should be able to just take the length of the non-`None` detections from the tracker; it'd be a tiny bit slower to sum vs reference a count but that's negligible given `length` will typically be small).

But yeah I think this probably makes a bit more structural sense at the `Tracker` level (even in my primary use-case you should be able to have an animated `Annotator` w/o using `Smoother`); want me to move it there?

yeldarby:
Updated to remove `tracker_length` and `track_starts`.

supervision/detection/tools/smoother.py
```diff
@@ -0,0 +1,161 @@
...
+                # skip detections without a tracker id
+                continue
+
+            if self.tracks.get(tracker_id, None) is None:
+                self.tracks[tracker_id] = deque(maxlen=self.length)
+
+            self.tracks[tracker_id].append(detections[detection_idx])
+            self.track_ends[tracker_id] = self.current_frame
+
+        for track_id in self.tracks:
```
SkalskiP, line 123: We have no state-clearing mechanism. After some time, most of the tracks we store will be dead.

supervision/detection/tools/smoother.py
```diff
@@ -0,0 +1,161 @@
+from collections import deque
+from typing import Optional
+
+import numpy as np
+
+from supervision.detection.core import Detections
+
+
+class Smoother:
```
SkalskiP, line 9: What do you think about changing the name to `DetectionsSmoother`?

SkalskiP (changes requested):

supervision/detection/tools/smoother.py
capjamesg: @SkalskiP How would you recommend we clear state?

supervision/detection/tools/smoother.py
capjamesg: Updated in my latest commit.

supervision/detection/tools/smoother.py
```diff
@@ -0,0 +1,166 @@
...
+    def get_track(self, track_id: int) -> Optional[dict]:
+        track = self.tracks.get(track_id, None)
+        if track is None:
+            return None
+
+        track = [d for d in track if d is not None]
+        if len(track) == 0:
+            return None
+
+        ret = track[0]
```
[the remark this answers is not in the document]
capjamesg: Updated in my latest commit.

supervision/detection/tools/smoother.py
```diff
@@ -0,0 +1,166 @@
...
+        for track_id in self.tracks:
+            track = self.tracks[track_id]
+            if self.track_ends[track_id] < self.current_frame:
+                # continue tracking for a few frames after the object has left
+                # (to prevent flickering in case it comes back)
+                track.append(None)
+
+        return self.get_smoothed_detections()
+
+    def get_track(self, track_id: int) -> Optional[dict]:
```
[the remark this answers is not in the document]
capjamesg: Updated in my latest commit.

supervision/detection/tools/smoother.py
capjamesg: Should we have an explicit method `DetectionsSmoother.clear()`, or some logic here?

capjamesg:
@SkalskiP In my latest change, I add logic to purge unused tracker IDs:

```python
used_tracker_ids = {}

for detection_idx in range(len(detections)):
    tracker_id = detections.tracker_id[detection_idx]
    if tracker_id is None:
        # skip detections without a tracker id
        continue

    if self.tracks.get(tracker_id, None) is None:
        self.tracks[tracker_id] = deque(maxlen=self.length)

    self.tracks[tracker_id].append(detections[detection_idx])
    self.track_ends[tracker_id] = self.current_frame

    used_tracker_ids[tracker_id] = True

for track_id in list(self.tracks.keys()):
    if track_id not in used_tracker_ids:
        del self.tracks[track_id]
        del self.track_ends[track_id]
```

Here is an example of the smoothing running on a video:

https://github.com/roboflow/supervision/assets/37276661/7f514588-b2c9-4ef0-be91-920156ace99c

SkalskiP (approved):

PR #720, opened 2024-01-11
Add `sv.PercentageBarAnnotator()` for confidence level bar annotation in specified position 

### LINKED ISSUE #695: Add annotator `sv.PercentageBarAnnotator` which shows percentage bar of confidence level in specified position
### Search before asking

### Description

I have created a new annotator class named `PercentageBarAnnotator` which draws a percentage bar based on the confidence for each prediction, in a specific position around the box.

An issue already exists, issue #459 , but it seems to be inactive since it's 2+ months old. I had posted there a demo video from a previous try to implement that but it was not pretty and uncompleted.

`PercentageBarAnnotator` accepts the following attributes:

```
-   height: int = 14,
-   width: int = 80,
-   color: Union[Color, ColorPalette] = ColorPalette.default(),
-   border_color: Color = Color.black(),
-   position: Position = Position.TOP_CENTER,
-   color_lookup: ColorLookup = ColorLookup.CLASS,
```

`percentage_bar_annotator = sv.PercentageBarAnnotator()`

Below is a video demo of the annotator. Let me know your thoughts about it, for any possible changes, and if its good I can submit a PR.

https://github.com/roboflow/supervision/assets/65918758/e214fab1-ba12-4479-95cd-33cfcd42fcd6

Some more examples regarding the height attribute of the percentage bar:
`
sv.PercentageBarAnnotator(height=12)`

percentage_bar_height_12

`sv.PercentageBarAnnotator(height=22)`

percentage_bar_height_22

### Use case

_No response_

### Additional

_No response_

xaristeidou:
# Description

This PR Implements (Fix #695) 

This annotator will draw a percentage bar with borders, based on the confidence level for each prediction.
Annotator has the following attributes:

- `height`
- `width`
- `color`
- `border_color`
- `position`
- `color_lookup`

Execution time of annotation process (tested on Ryzen 7 5800H):
- Exec time for 40 objects: 0.0011143684387207031
- Mean exec time for each object: 2.7859210968017577e-05

**No new dependencies required for this change**

I tested using walking people asset video, prediction providing by YOLOv8. An example is presented below using `height = 16`, `width = 80`.
Screenshot from 2024-01-12 01-29-43

- Add new annotator class in Documentation
- Add image example link reference in class Description in annotators/core.py

## Docs

-   [ ] Docs updated?  **NO**


onuralpszr:
Google Collab link: https://colab.research.google.com/drive/1SQUkBXcvCB-FGeMb9F6xLH1U30N878jm?usp=sharing

onuralpszr (approved):
I did some gitignore changes and docs typo fixes as well. I also added docs for Percentage Bar I only waiting for pics to be upload so It will render properly.

onuralpszr:
cc @SkalskiP

supervision/annotators/core.py
```diff
@@ -1496,3 +1496,138 @@ def annotate(
...
+        """
+        for detection_idx in range(len(detections)):
+            cx, cy = detections.get_anchors_coordinates(anchor=self.position)[
+                detection_idx
+            ].astype(int)
+
+            conf = detections.confidence[detection_idx]
+            border_thickness = int(0.15 * self.height)
+
+            border_mapping = {
```
SkalskiP, line 1578: please, let's move border mapping logic to separate function / method

supervision/annotators/core.py
```diff
@@ -1496,3 +1496,138 @@ def annotate(
...
+                Position.CENTER_RIGHT: (
+                    (cx, cy - self.height // 2),
+                    (cx + self.width, cy + self.height // 2),
+                ),
+                Position.BOTTOM_LEFT: ((cx - self.width, cy), (cx, cy + self.height)),
+                Position.BOTTOM_CENTER: (
+                    (cx - self.width // 2, cy),
+                    (cx + self.width // 2, cy + self.height),
+                ),
+                Position.BOTTOM_RIGHT: ((cx, cy), (cx + self.width, cy + self.height)),
```
SkalskiP, line 1602: There is also `CENTER_OF_MASS`. It should behave as `CENTER` in this case.

supervision/annotators/core.py
```diff
@@ -1496,3 +1496,138 @@ def annotate(
...
+                Options are `INDEX`, `CLASS`, `TRACK`.
+        """
+        self.height: int = height
+        self.width: int = width
+        self.color: Union[Color, ColorPalette] = color
+        self.border_color: Color = border_color
+        self.position: Position = position
+        self.color_lookup: ColorLookup = color_lookup
+
+    def annotate(
```
SkalskiP, line 1534: One thing we really should add to make it a killer annotator is allowing users to pass a custom NumPy array with values between 0 and 1. This way users would be able to plot confidence by default, but also any other percentage value.

SkalskiP (changes requested):

supervision/annotators/core.py
onuralpszr: I added CENTER_OF_MASS

supervision/annotators/core.py
onuralpszr: I made custom_value acceptability instead of using conf values and I added couple of different types and make sure values are between 0 and 1 as well. Please check it out.

SkalskiP (approved):

SkalskiP:
Hi @xaristeidou and @onuralpszr, I just made the final changes. We are merging!

Issue #722, opened 2024-01-12
speed estimation issue


yeongnamtan:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar bug report.

### Bug

I run the speed estimate code and get the error below:

Code:

```console
!python ultralytics_example.py \
  --source_video_path data/vehicles.mp4 \
  --target_video_path data/vehicles-result.mp4 \
  --confidence_threshold 0.3 \
  --iou_threshold 0.5
```

Error:

```console
0: 384x640 (no detections), 65.5ms
Speed: 3.5ms preprocess, 65.5ms inference, 0.7ms postprocess per image at shape (1, 3, 384, 640)
Traceback (most recent call last):
  File "ultralytics_example.py", line 114, in <module>
    points = view_transformer.transform_points(points=points).astype(int)
  File "ultralytics_example.py", line 34, in transform_points
    return transformed_points.reshape(-1, 2)
AttributeError: 'NoneType' object has no attribute 'reshape'
```

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

onuralpszr:
Hello, @yeongnamtan first did you download the video and did you use the same code or did you change something If yes what is it ?

yeongnamtan:
@onuralpszr I downloaded the video and use the same code. I didn't change anything.

onuralpszr:
> @onuralpszr I downloaded the video and use the same code. I didn't change anything.

What is your Python env, and also share the pip list please, too. Also, what is your operation system ?

yeongnamtan:
@onuralpszr  I used sagemaker studio python 3.8.   

Package                     Version
--------------------------- --------------------
apex                        0.1
appdirs                     1.4.4
asttokens                   2.4.1
attrs                       22.2.0
awscli                      1.27.70
awsio                       0.0.1
backcall                    0.2.0
bcrypt                      4.0.1
bokeh                       2.4.3
boto3                       1.26.70
botocore                    1.29.70
brotlipy                    0.7.0
cached-property             1.5.2
certifi                     2022.12.7
cffi                        1.15.1
charset-normalizer          2.1.1
click                       8.1.3
cloudpickle                 2.2.1
cmake                       3.18.2.post1
colorama                    0.4.4
conda                       22.11.1
conda-content-trust         0.1.3
conda-package-handling      2.0.2
conda_package_streaming     0.7.0
[...]

PR #733, opened 2024-01-16
`fix/speed-estimation-issue-when-detections-empty`

SkalskiP:
# Description

Fix https://github.com/roboflow/supervision/issues/722.

## Type of change

### DIFF
--- a/examples/speed_estimation/ultralytics_example.py
+++ b/examples/speed_estimation/ultralytics_example.py
@@ -23,12 +23,16 @@ TARGET = np.array(
 
 
 class ViewTransformer:
+
     def __init__(self, source: np.ndarray, target: np.ndarray) -> None:
         source = source.astype(np.float32)
         target = target.astype(np.float32)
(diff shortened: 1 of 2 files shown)


examples/speed_estimation/ultralytics_example.py
```diff
@@ -29,6 +29,9 @@ def __init__(self, source: np.ndarray, target: np.ndarray) -> None:
         self.m = cv2.getPerspectiveTransform(source, target)
 
     def transform_points(self, points: np.ndarray) -> np.ndarray:
+        if points.size == 0:
+            return points
```
onuralpszr, line 33: AttributeError: 'NoneType' object has no attribute 'reshape'

Error was NoneType so ".size" can also cause same because NoneType doesn't have ".size" so you should check type instead of size first.

SkalskiP: That None is generated later. It is generated if point list is empty.

PR #756, opened 2024-01-18
feat(color.py): ✨ add color constants

onuralpszr:
# Description

This commit adds the WHITE, BLACK, RED, GREEN, and BLUE color constants to the Color class in color.py. These constants represent the corresponding colors in the RGB color space.

## Type of change

-   [X] Update existing feature (breaking change which adds functionality)


supervision/draw/color.py
```diff
@@ -116,25 +122,40 @@ def as_bgr(self) -> Tuple[int, int, int]:
...
+    def RED(cls):
+        if cls._RED is None:
+            cls._RED = cls.from_hex("#ff0000")
+        return cls._RED
 
+    @property
     @classmethod
-    def green(cls) -> Color:
-        return Color.from_hex(color_hex="#00ff00")
+    def GREEN(cls):
```
SkalskiP, line 148: Do we need `cls._GREEN`? Can't we just return `cls.from_hex("#00ff00")`?

SkalskiP: The same question goes for all other colors.

supervision/draw/color.py
```diff
@@ -116,25 +122,40 @@ def as_bgr(self) -> Tuple[int, int, int]:
...
-    def black(cls) -> Color:
-        return Color.from_hex(color_hex="#000000")
+    def BLACK(cls):
+        if cls._BLACK is None:
+            cls._BLACK = cls.from_hex("#000000")
+        return cls._BLACK
 
+    @property
     @classmethod
-    def red(cls) -> Color:
```
SkalskiP, line 128: Please mark those methods as deprecated and announce dropping them in `supervision-0.20.0`.

SkalskiP: The same request goes for all other colors.

SkalskiP (changes requested):

onuralpszr:
@SkalskiP  changes are done.

SkalskiP (approved):

Issue #792, opened 2024-01-26
[LineZoneAnnotator] - make the display of in/out values optional


SkalskiP:
### Description

Make the display of [`in_count`](https://github.com/roboflow/supervision/blob/3024ddca83ad837651e59d040e2a5ac5b2b4f00f/supervision/detection/line_counter.py#L53) and [`out_count`](https://github.com/roboflow/supervision/blob/3024ddca83ad837651e59d040e2a5ac5b2b4f00f/supervision/detection/line_counter.py#L54) values optional. This may require adding additional `display_in_count` and `display_out_count` arguments to [`LineZoneAnnotator`](https://github.com/roboflow/supervision/blob/3024ddca83ad837651e59d040e2a5ac5b2b4f00f/supervision/detection/line_counter.py#L165).

### API

```python
class LineZoneAnnotator:
    def __init__(
        self,
        thickness: float = 2,
        color: Color = Color.WHITE,
        text_thickness: float = 2,
        text_color: Color = Color.BLACK,
        text_scale: float = 0.5,
        text_offset: float = 1.5,
        text_padding: int = 10,
        custom_in_text: Optional[str] = None,
        custom_out_text: Optional[str] = None,
        display_in_count: bool = True,
        display_out_count: bool = True,
    ):
        # Existing initialization code...

        self.display_in_count: bool = display_in_count
        self.display_in_count: bool = display_in_count

    def annotate(self, frame: np.ndarray, line_counter: LineZone) -> np.ndarray:
        # Required logic changes...
```

### Additional

- Note: Please share a Google Colab with minimal code to test the new feature. We know it's additional work, but it will definitely speed up the review process. Each change must be tested by the reviewer. Setting up a local environment to do this is time-consuming. Please ensure that Google Colab can be accessed without any issues (make it public). Thank you! 🙏🏻

Kadermiyanyedi:
Hi, I am interesting with this task. Can you assign me?

onuralpszr:
@Kadermiyanyedi  hello and task is your good luck :)

PR #802, opened 2024-01-28
LineZoneAnnotator make display of int/out values optional

Kadermiyanyedi:
## Description

LineZoneAnnotator has two new boolean parameters, display_in_count and display_out_count, to the functions display_in_count and display_on_count. When set to True, these parameters display the input/output values. (Fix: #792)

To optimize performance and avoid unnecessary mathematical calculations when the values are not displayed, I encapsulated these calculations within a separate function.

For testing purposes, please refer to the provided [Colab link](https://colab.research.google.com/drive/1ChKBEtA3ij-Cezf3DKOBoX5rs2VhXGdf?usp=sharing).

### Type of change
-   [x ] New feature (non-breaking change which adds functionality)

### Changes Made:

Added two boolean parameters, display_in_count and display_out_count, to display_in_count and display_on_count functions.
Implemented conditional logic to display input/output values based on the boolean parameters.
Moved unnecessary mathematical calculations to a separate function to optimize performance.

Example Output: [Google Drive Link](https://drive.google.com/file/d/111yZslXcmxgkbFzzMQZ6yn6nshPbyxC0/view?usp=sharing)

Please review thank you :)

### DIFF
--- a/supervision/detection/line_counter.py
+++ b/supervision/detection/line_counter.py
@@ -218,7 +218,7 @@ class LineZoneAnnotator:
             frame (np.ndarray): The image on which the text will be drawn.
             center_text_anchor: The center point that the text will be drawn.
             text (str): The text that will be drawn.
-            is_in_count (bool): Whether to display the in count or not.
+            is_in_count (bool): Whether to display the in count or out count.
         """
         _, text_height = cv2.getTextSize(
             text, cv2.FONT_HERSHEY_SIMPLEX, self.text_scale, self.text_thickness
--- a/supervision/detection/line_counter.py
+++ b/supervision/detection/line_counter.py
@@ -5,7 +5,8 @@ import numpy as np
 
 from supervision.detection.core import Detections
 from supervision.draw.color import Color
-from supervision.geometry.core import Point, Position, Rect, Vector
+from supervision.draw.utils import draw_text
+from supervision.geometry.core import Point, Position, Vector
 
 
 class LineZone:
@@ -209,7 +210,7 @@ class LineZoneAnnotator:
         frame: np.ndarray,
         line_counter: LineZone,
         text: str,
-        in_count: bool,
+        count_pos: bool,
     ) -> None:
         """This method is drawing the text on the frame.
 
@@ -218,51 +219,32 @@ class LineZoneAnnotator:
             line_counter (LineCounter): The line counter
                 that will be used to draw the line.
             text (str): The text that will be drawn.
-            in_count (bool): Whether to display the in count or not.
+            count_pos (bool): Whether to display the in count or not.
         """
-        (text_width, text_height), _ = cv2.getTextSize(
+        text_width, text_height = cv2.getTextSize(
             text, cv2.FONT_HERSHEY_SIMPLEX, self.text_scale, self.text_thickness
-        )
+        )[0]
 
-        text_x = (
-            line_counter.vector.end.x + line_counter.vector.start.x - text_width
-        ) / 2
+        center_text_anchor = Vector(
+            start=line_counter.vector.start, end=line_counter.vector.end
+        ).center
 
-        text_y = (
-            line_counter.vector.end.y + line_counter.vector.start.y + text_height
-        ) / 2
-
-        if in_count:
-            # if the text is an in count, we want to display it above the line
-            text_y -= self.text_offset * text_height
+        if count_pos:
+            center_text_anchor.y -= int(self.text_offset * text_height)
         else:
-            # if the text is an out count, we want to display it below the line
-            text_y += self.text_offset * text_height
+            center_text_anchor.y += int(self.text_offset * text_height)
 
-        text_background_rect = Rect(
-            x=text_x,
-            y=text_y - text_height,
-            width=text_width,
-            height=text_height,
-        ).pad(padding=self.text_padding)
-
-        cv2.rectangle(
-            frame,
-            text_background_rect.top_left.as_xy_int_tuple(),
-            text_background_rect.bottom_right.as_xy_int_tuple(),
-            self.color.as_bgr(),
-            -1,
-        )
-
-        cv2.putText(
-            frame,
-            text,
-            (int(text_x), int(text_y)),
-            cv2.FONT_HERSHEY_SIMPLEX,
-            self.text_scale,
-            self.text_color.as_bgr(),
-            self.text_thickness,
-            cv2.LINE_AA,
+        draw_text(
+            scene=frame,
+            text=text,
+            text_anchor=center_text_anchor,
+            text_color=self.text_color,
+            text_scale=self.text_scale,
+            text_thickness=self.text_thickness,
+            text_padding=self.text_padding,
+            background_color=self.color,
+            text_width=text_width,
+            text_height=text_height,
         )
 
     def annotate(self, frame: np.ndarray, line_counter: LineZone) -> np.ndarray:
@@ -311,7 +293,10 @@ class LineZoneAnnotator:
                 else f"in: {line_counter.in_count}"
             )
             self._annotate_count(
-                line_counter=line_counter, text=in_text, in_count=True, frame=frame
+                frame=frame,
+                line_counter=line_counter,
+                text=in_text,
+                count_pos=True,
             )
 
         if self.display_out_count:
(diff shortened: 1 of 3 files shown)


SkalskiP:
Hi @Kadermiyanyedi 👋🏻! The logic looks good. 

I like that you decided to extract the drawing of the text using a separate method. Could we go a step further? We already have [`draw_text`](
https://github.com/roboflow/supervision/blob/9deea666798b25eebca80f6661324a2a842e4e96/supervision/draw/utils.py#L104) function, and we could use it here.

supervision/detection/line_counter.py
```diff
@@ -197,6 +201,69 @@ def __init__(
...
+            line_counter (LineCounter): The line counter
+                that will be used to draw the line.
+            text (str): The text that will be drawn.
+            in_count (bool): Whether to display the in count or not.
+        """
+        (text_width, text_height), _ = cv2.getTextSize(
+            text, cv2.FONT_HERSHEY_SIMPLEX, self.text_scale, self.text_thickness
+        )
+
+        text_x = (
```
SkalskiP, line 227: I think we could add vector `center` property and hide this logic there.

```python
@dataclass
class Vector:
    start: Point
    end: Point

    @property
    def center(self) -> Point:
        pass
```

supervision/detection/line_counter.py
```diff
@@ -197,6 +201,69 @@ def __init__(
...
+            text_y += self.text_offset * text_height
+
+        text_background_rect = Rect(
+            x=text_x,
+            y=text_y - text_height,
+            width=text_width,
+            height=text_height,
+        ).pad(padding=self.text_padding)
+
+        cv2.rectangle(
```
SkalskiP, line 249: We could use `draw_text` here and remove the logic for drawing text and background. We already use it to draw the `PolygonZone` label.

supervision/detection/line_counter.py
```diff
@@ -197,6 +201,69 @@ def __init__(
...
+        text_x = (
+            line_counter.vector.end.x + line_counter.vector.start.x - text_width
+        ) / 2
+
+        text_y = (
+            line_counter.vector.end.y + line_counter.vector.start.y + text_height
+        ) / 2
+
+        if in_count:
+            # if the text is an in count, we want to display it above the line
```
SkalskiP, line 236: Please remove the comments. I think we do not need it. :)

SkalskiP (changes requested):

supervision/detection/line_counter.py
onuralpszr: @SkalskiP I think we should changed in_count to pos_count technically we determine pos based bool valueso using pos is better 

```python
       if count_pos:
            # if count_pos is true, we want to display text above the line
            text_y -= self.text_offset * text_height
        else:
            # if count_pos is false, we want to display text below the line
            text_y += self.text_offset * text_height
```

onuralpszr: Actually we can remove all together entirely... with draw_text

SkalskiP: Let's go iteratively. I do not want to drop to many comment at once :)

supervision/detection/line_counter.py
Kadermiyanyedi: Thanks for your advice. I updated my PR.

supervision/detection/line_counter.py
Kadermiyanyedi: Thanks for your advice. I updated my PR.  I needed to make some minor updates to the draw_text function. I'm open to different suggestions

supervision/detection/line_counter.py
Kadermiyanyedi: Thanks for your advicese. I updated my PR @onuralpszr @SkalskiP

supervision/detection/line_counter.py
```diff
@@ -197,6 +202,50 @@ def __init__(
...
+            line_counter (LineCounter): The line counter
+                that will be used to draw the line.
+            text (str): The text that will be drawn.
+            count_pos (bool): Whether to display the in count or not.
+        """
+        text_width, text_height = cv2.getTextSize(
+            text, cv2.FONT_HERSHEY_SIMPLEX, self.text_scale, self.text_thickness
+        )[0]
+
+        center_text_anchor = Vector(
```
SkalskiP, line 228: I'd refactor the `_annotate_count`, and instead of passing the whole `line_counter: LineZone` object as an argument here, I'd pass the result of this code - `text_anchor: Point.`

supervision/detection/line_counter.py
```diff
@@ -197,6 +202,50 @@ def __init__(
...
         self.custom_out_text: str = custom_out_text
+        self.display_in_count: bool = display_in_count
+        self.display_out_count: bool = display_out_count
+
+    def _annotate_count(
+        self,
+        frame: np.ndarray,
+        line_counter: LineZone,
+        text: str,
+        count_pos: bool,
```
SkalskiP, line 213: I'd rename this argument to something more descriptive. For example, `is_in_count`?

SkalskiP (changes requested):

SkalskiP:
Awesome job @Kadermiyanyedi! 🔥 I left a few more comments, but we are close. I think we should be done after this.

supervision/geometry/core.py
```diff
@@ -56,6 +56,19 @@ def magnitude(self) -> float:
...
+    @property
+    def center(self) -> Point:
+        """
+        Calculate the center point of the vector.
+
+        Returns:
+            Point: The center point of the vector.
+        """
+        return Point(
+            x=(self.start.x + self.end.x) // 2,
```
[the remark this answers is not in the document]
Kadermiyanyedi: `org=(text_anchor.x - text_width // 2, text_anchor.y + text_height // 2),
`
This code block sometimes returns a float value. If it is appropriate, I would like to apply int type change here.

An example of the error encountered and the printed version of the points:

image

supervision/draw/utils.py
```diff
@@ -111,6 +111,8 @@ def draw_text(
     text_padding: int = 10,
     text_font: int = cv2.FONT_HERSHEY_SIMPLEX,
     background_color: Optional[Color] = None,
+    text_width: Optional[int] = None,
```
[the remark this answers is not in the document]
Kadermiyanyedi: The input/output value to draw above and below the line zone, I had to update the y value of the text_anchor with text_offset and text_height. I made this change to avoid recalculation. I reverted my changes. Thanks for your advice

Kadermiyanyedi:
@SkalskiP @onuralpszr I resolved all comments, can you review again please :)

SkalskiP:
@Kadermiyanyedi, could you make your Google Colab public?

Kadermiyanyedi:
> @Kadermiyanyedi, could you make your Google Colab public? The code looks good I just want to confirm everything works :)

I updated Colab link, now you can access it :)

onuralpszr:
Tests are passed that looks good.

supervision/detection/line_counter.py
```diff
@@ -197,6 +202,43 @@ def __init__(
...
+        text: str,
+        is_in_count: bool,
+    ) -> None:
+        """This method is drawing the text on the frame.
+
+        Args:
+            frame (np.ndarray): The image on which the text will be drawn.
+            center_text_anchor: The center point that the text will be drawn.
+            text (str): The text that will be drawn.
+            is_in_count (bool): Whether to display the in count or not.
```
onuralpszr, line 221:
```suggestion
            is_in_count (bool): Whether to display the in count or out count.
```

onuralpszr: @Kadermiyanyedi  could you please update that small part please as well.

Kadermiyanyedi: Done

SkalskiP (approved):

SkalskiP:
Awesome! @Kadermiyanyedi, thanks a lot for your help! 🙏🏻 We are merging! 

Are you on LinkedIn?

Kadermiyanyedi:
@onuralpszr Thank you, you've been very helpful in assigning the relevant task and aiding my understanding of the project

@SkalskiP Thank you very much for taking the time to provide feedback and review. Your comments are highly valuable

SkalskiP:
My pleasure!

Issue #790, opened 2024-01-26
[LineZone] - allow per class counting  


SkalskiP:
### Description

Currently, [sv.LineZone](https://github.com/roboflow/supervision/blob/3024ddca83ad837651e59d040e2a5ac5b2b4f00f/supervision/detection/line_counter.py#L11) provides only aggregated counts - all classes are thrown into one bucket. In the past, many users have asked us to provide more granular - per class count. This can be achieved by adding `class_in_count` and `class_out_count` dictionaries that will store per-class counts.

### API

```python
class LineZone:

    def __init__(
        self,
        start: Point,
        end: Point,
        triggering_anchors: Iterable[Position] = (
            Position.TOP_LEFT,
            Position.TOP_RIGHT,
            Position.BOTTOM_LEFT,
            Position.BOTTOM_RIGHT,
        ),
    ):
        # Existing initialization code...

        self.class_in_count: Dict[int, int] = {}
        self.class_out_count: Dict[int, int] = {}

    def trigger(self, detections: Detections) -> Tuple[np.ndarray, np.ndarray]:
        crossed_in = np.full(len(detections), False)
        crossed_out = np.full(len(detections), False)

        # Required logic changes...
```

### Additional

- Note: Please share a Google Colab with minimal code to test the new feature. We know it's additional work, but it will definitely speed up the review process. Each change must be tested by the reviewer. Setting up a local environment to do this is time-consuming. Please ensure that Google Colab can be accessed without any issues (make it public). Thank you! 🙏🏻

RaghavvGupta:
Hey @SkalskiP, I was wondering to work on this. I have this logic in mind, pls confirm if this approach is what you guys want?

Dictionaries to keep track of counts for each class

`self.class_in_count: Dict[str, int] = {}
self.class_out_count: Dict[str, int] = {}
`

Updating per class-in and class-out count

```
class_label = detections.class_labels[i]
                if class_label not in self.class_in_count:
                    self.class_in_count[class_label] = 0
                self.class_in_count[class_label] += 1
            else:
                self.out_count += 1
                crossed_out[i] = True

                # Updating per-class out count
                class_label = detections.class_labels[i]
                if class_label not in self.class_out_count:
                    self.class_out_count[class_label] = 0
                self.class_out_count[class_label] += 1
```

RaghavvGupta:
Hey @SkalskiP, I was wondering to work on this. I have this logic in mind, pls confirm if this approach is what you guys want?

[...]

PR #836, opened 2024-02-01
Adjusting LineZoneAnnotator to display per-class counts in annotations

RaghavvGupta:
Adjusting LineZoneAnnotator to display per-class counts in annotations #790

# Description

Introduced per-class counts for objects crossing the line. Updated the trigger method to handle per-class counts.

## Type of change

class_in_count (Dict[int, int]): A dictionary to store per-class in counts.
class_out_count (Dict[int, int]): A dictionary to store per-class out counts.
These dictionaries will store counts for each class separately.

### DIFF
--- a/supervision/detection/line_counter.py
+++ b/supervision/detection/line_counter.py
@@ -51,8 +51,10 @@ class LineZone:
         self.vector = Vector(start=start, end=end)
         self.limits = self.calculate_region_of_interest_limits(vector=self.vector)
         self.tracker_state: Dict[str, bool] = {}
-        self.in_count: int = 0
-        self.out_count: int = 0
+        #self.in_count: int = 0
+        #self.out_count: int = 0
+        self.class_in_count: Dict[int, int] = {}  # Per-class in count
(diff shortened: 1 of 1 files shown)


LinasKo:
Hi @RaghavvGupta 👋

Thank you for your contribution, and apologies for the long wait.

Your solution fits the original specification (except for tests on Colab), but there were a few flaws and we had no time to give feedback. Briefly:
1. This removed the prior `in_count` and `out_count` variables which might break existing code
2. This creates new behaviour for the existing annotator, without providing a way to access old annotation style.
3. There's no Colab with tests.

It turned out a more complicated change than we expected, but the implementation is on #1555.

I'll close this one, but you have my sincerest thanks for the effort put in! 🙏

RaghavvGupta:
Hi @LinasKo, Thanks for your feedback! I have understood the implementation and sincerely apologise for not being able to deliver the requisite. However, I learned a lot from this!
Thanks again for giving me this opportunity!

Issue #678, opened 2023-12-15
[InferenceSlicer] - add segmentation models support


SkalskiP:
### Description

Currently, [`sv.InferenceSlicer`](https://supervision.roboflow.com/detection/tools/inference_slicer/) supports only object detection models. Adding support for instance segmentation would require the following changes:

- The `sv.InferenceSlicer` uses Non-Max Suppression (NMS) to sift out duplicate detections at the tile intersection. At the moment, Supervision only has a box-based [NMS](https://github.com/roboflow/supervision/blob/5b7e0bac5afa1c4aaf47d75fa58d60bf92d032a2/supervision/detection/utils.py#L59). A segmentation-based NMS would be almost ideal, the only change would be to replace the [`box_iou_batch`](https://github.com/roboflow/supervision/blob/5b7e0bac5afa1c4aaf47d75fa58d60bf92d032a2/supervision/detection/utils.py#L28) with a new `mask_iou_batch`.
- A segmentation-based NMS must be plugged into `sv.InferenceSlicer`. At [this](https://github.com/roboflow/supervision/blob/5b7e0bac5afa1c4aaf47d75fa58d60bf92d032a2/supervision/detection/tools/inference_slicer.py#L112) point, we would have to check whether the detections have masks. And if so, use the new NMS.

### API

```python
# create
def mask_iou_batch(boxes_true: np.ndarray, boxes_detection: np.ndarray) -> np.ndarray:
    pass

# rename non_max_suppression -> box_non_max_suppression

# create
def mask_non_max_suppression(predictions: np.ndarray, iou_threshold: float = 0.5) -> np.ndarray:
    pass

# change InferenceSlicer
```

### Usage example

```python
import cv2
import supervision as sv
from ultralytics import YOLO

image = cv2.image = cv2.imread(<SOURCE_IMAGEPATH>)
model = YOLO("yolov8x-seg.pt")

def callback(image_slice: np.ndarray) -> sv.Detections:
    result = model(image_slice)[0]
    return sv.Detections.from_ultralytics(result)

slicer = sv.InferenceSlicer(
    callback=callback,
    slice_wh=(512, 512),
    iou_threshold=0.5,
)

detections = slicer(image)
```

### Additional

[...]

PR #847, opened 2024-02-03
[NMS] - add segmentation models support

AdonaiVera:
# Description

This PR introduces Non-Maximum Suppression (NMS) algorithm focused on segmentation, enhancing our object detection capabilities, particularly in segmentation tasks. We have renamed the traditional `non_max_suppression` function to `box_non_max_suppression` for better clarity regarding its application to bounding boxes. Furthermore, we've integrated a conditional mechanism within the `with_nms` function that utilizes segmentation masks for NMS when such masks are present in the predictions. This optimization leverages the spatial context provided by segmentation masks to improve the suppression process.

This enhancement is part of a task segmented into two parts for more focused development and review. This PR addresses the first part, as discussed in the related issue [here](https://github.com/roboflow/supervision/issues/678). Splitting the task ensures thorough implementation and testing of each component.

In addition, this PR encompasses comprehensive unit tests for the new NMS functionality and a demo that demonstrates the algorithm's application and effectiveness in real-world scenarios.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

I created a demo to showcase this functionality [here](https://colab.research.google.com/drive/1lpJKXryY59FTrgZU85a3r0KK7gm1OTNh?usp=sharing)

## Docs

- [x] Docs updated? What were the changes:

### DIFF
--- a/supervision/__init__.py
+++ b/supervision/__init__.py
@@ -42,12 +42,13 @@ from supervision.detection.tools.polygon_zone import PolygonZone, PolygonZoneAnn
 from supervision.detection.tools.smoother import DetectionsSmoother
 from supervision.detection.utils import (
     box_iou_batch,
+    box_non_max_suppression,
     calculate_masks_centroids,
     filter_polygons_by_area,
+    mask_non_max_suppression,
     mask_to_polygons,
     mask_to_xyxy,
     move_boxes,
-    non_max_suppression,
     polygon_to_mask,
     polygon_to_xyxy,
     scale_boxes,
--- a/supervision/__init__.py
+++ b/supervision/__init__.py
@@ -45,6 +45,7 @@ from supervision.detection.utils import (
     box_non_max_suppression,
     calculate_masks_centroids,
     filter_polygons_by_area,
+    mask_iou_batch,
     mask_non_max_suppression,
     mask_to_polygons,
     mask_to_xyxy,
--- a/supervision/detection/core.py
+++ b/supervision/detection/core.py
@@ -1002,7 +1002,7 @@ class Detections:
         self, threshold: float = 0.5, class_agnostic: bool = False
     ) -> Detections:
         """
-        Performs non-max suppression on detection set. If the detections result 
+        Performs non-max suppression on detection set. If the detections result
         from a segmentation model, the IoU mask is applied. Otherwise, box IoU is used.
 
         Args:
--- a/supervision/detection/core.py
+++ b/supervision/detection/core.py
@@ -1033,14 +1033,22 @@ class Detections:
             assert self.class_id is not None, (
                 "Detections class_id must be given for NMS to be executed. If you intended"
                 " to perform class agnostic NMS set class_agnostic=True."
-            )   
+            )
             predictions = np.hstack(
-                (self.xyxy, self.confidence.reshape(-1, 1), self.class_id.reshape(-1, 1))
+                (
+                    self.xyxy,
+                    self.confidence.reshape(-1, 1),
+                    self.class_id.reshape(-1, 1),
+                )
             )
 
-        if hasattr(self, 'mask') and self.mask is not None:
-            indices = mask_non_max_suppression(predictions=predictions, masks=self.mask, iou_threshold=threshold)
+        if hasattr(self, "mask") and self.mask is not None:
+            indices = mask_non_max_suppression(
+                predictions=predictions, masks=self.mask, iou_threshold=threshold
+            )
         else:
-            indices = box_non_max_suppression(predictions=predictions, iou_threshold=threshold)
+            indices = box_non_max_suppression(
+                predictions=predictions, iou_threshold=threshold
+            )
 
         return self[indices]
--- a/supervision/detection/core.py
+++ b/supervision/detection/core.py
@@ -1002,7 +1002,7 @@ class Detections:
         self, threshold: float = 0.5, class_agnostic: bool = False
     ) -> Detections:
         """
-        Perform non-maximum suppression on the current set 
+        Perform non-maximum suppression on the current set
         of object detections or segmentation prediction.
 
         Args:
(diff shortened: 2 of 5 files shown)


supervision/__init__.py
```diff
@@ -42,12 +42,13 @@
 from supervision.detection.tools.smoother import DetectionsSmoother
 from supervision.detection.utils import (
     box_iou_batch,
```
SkalskiP, line 44: Let's import `mask_iou_batch` as well.

supervision/detection/core.py
```diff
@@ -1028,18 +1030,26 @@ def with_nms(
...
+                    self.class_id.reshape(-1, 1),
+                )
             )
-            return self[indices]
 
-        assert self.class_id is not None, (
-            "Detections class_id must be given for NMS to be executed. If you intended"
-            " to perform class agnostic NMS set class_agnostic=True."
-        )
+        if hasattr(self, "mask") and self.mask is not None:
```
SkalskiP, line 1046: `Detections` always has an attribute `mask`, but it can be equal to `None` or `np.ndarray`. So we can simplify this condition to `self.mask is not None`.

supervision/detection/core.py
```diff
@@ -1001,7 +1002,8 @@ def with_nms(
         self, threshold: float = 0.5, class_agnostic: bool = False
     ) -> Detections:
         """
-        Perform non-maximum suppression on the current set of object detections.
+        Perform non-maximum suppression on the current set
```
SkalskiP, line 1005: Thanks for noticing the lack of description. I would replace it with something more precise. 

Performs non-max suppression on detection set. If the detections result from a segmentation model, the IoU mask is applied. Otherwise, box IoU is used.

SkalskiP (changes requested):

SkalskiP:
Hi @AdonaiVera 👋🏻 ! I left a few comments on the code, but for the moment, these are secondary issues. The main problem I see is speed. Using your Colab example, I measured that it takes 13 seconds to calculate the `mask_iou_batch` for your image. We need to make it 1-2 orders of magnitude faster.

A quick query in ChatGPT suggested the code below, but I didn't check if the calculation result was the same or if the code was faster. I leave it here simply as an idea.

```python
import numpy as np

def mask_iou_batch(masks_true: np.ndarray, masks_detection: np.ndarray) -> np.ndarray:
    """
    Compute Intersection over Union (IoU) of two sets of masks -
        `masks_true` and `masks_detection`.

    Args:
        masks_true (np.ndarray): 3D `np.ndarray` representing ground-truth masks.
        masks_detection (np.ndarray): 3D `np.ndarray` representing detection masks.

    Returns:
        np.ndarray: Pairwise IoU of masks from `masks_true` and `masks_detection`.
    """
    # No need to cast to bool here if the masks are already boolean

    # Calculate intersection and union areas
    intersection_area = np.logical_and(masks_true[:, None], masks_detection).sum(axis=(2, 3))
    masks_true_area = masks_true.sum(axis=(1, 2))[:, None]
    masks_detection_area = masks_detection.sum(axis=(1, 2))

    # Calculate union area using the inclusion-exclusion principle
    union_area = masks_true_area + masks_detection_area - intersection_area

    # Avoid division by zero and calculate IoU
    iou = np.divide(intersection_area, union_area, out=np.zeros_like(intersection_area, dtype=float), where=union_area != 0)

    return iou
```

The second idea I have is to resize the masks. Maybe we don't need to count IoU at full resolution. I'm curious, if we did rescaling on the input to `mask_iou_batch` and scaled the masks to the maximum dimension of 640 would the IoU result differ much?

AdonaiVera:
Hi @SkalskiP 👋 

I made the changes to the algorithm based on your feedback. I executed the test on my local machine and found that the execution time for the `mask_iou_batch` function was 11 seconds without any improvements. 
`mask_iou_batch execution time: 11.263449907302856 seconds`

Then, I implemented your suggestion of using the inclusion-exclusion principle to improve the `mask_iou_batch` function. After this change, the execution time reduced from 11 seconds to almost 4 seconds. 
`mask_iou_batch execution time: 3.967336893081665 seconds`

After that, I applied your resize function to the input images before applying the `mask_iou_batch` function, and this further decreased the execution time to almost 0.5 seconds.
`mask_iou_batch execution time: 0.5139651298522949 seconds`

Thank you for your improvements, and I believe that these changes have significantly improved the algorithm's speed. I also incorporated the other small changes that you suggested and pushed the changes.

I think the idea of resizing the images worked really well because, from the graph you provided, the absolute value of the difference between the mask IoU calculated on full-size masks and on scaled masks is not significant, and the processing time improved a lot.

I executed all the unit test and the algorithm is working fine 👍 

PDT: I left the print time message to keep making more tests if needed. Thank you for your valuable feedback. 🚀 

Thank you 🥷

SkalskiP:
Hi @AdonaiVera 👋🏻 ! I made a few small changes in the second part of `mask_non_max_suppression`. Converted those for loops into a more vectorized solution. 

```python
keep = np.ones(rows, dtype=bool)
for i in range(rows):
    if keep[i]:
        condition = (ious[i] > iou_threshold) & (categories[i] == categories)
        keep[i + 1 :] = np.where(condition[i + 1 :], False, keep[i + 1 :])

return keep[sort_index.argsort()]
```

I also added some additional unit tests. 

In general, we could make it even faster. But taking it from 13 to 0.4 seconds is still a good start. I'm merging! Thanks a lot for your help! 🔥

SkalskiP (approved):

AdonaiVera:
Amazing, thank you @SkalskiP 🚀  
Now, I'm going to move on to the second part of the [issue](https://github.com/roboflow/supervision/issues/678) 
```
- load image
- slice the image into NxN tiles
- surround smaller size slices (the ones close to the edges) with a letterbox so that all tiles are NxN
- loop over slices
   - run inference
- update box coordinate values to match the image coordinate system, not the slice coordinate system
- pad masks to match the image coordinate system, not the slice coordinate system
- merge detections
```

I'll update a PR soon 💪

SkalskiP:
@AdonaiVera, thanks a lot! 🙏🏻 You know where to find me!

PR #871, opened 2024-02-09
docs(notebooks): 📝 add badges to all notebooks for support multiple platforms and download button

onuralpszr:
### Description

I added all badges from quickstart and adjusted based on notebook for making sure we cover more options for people to work for it. 

## Docs

-   [X] Docs updated? What were the changes:


SkalskiP:
Hi @onuralpszr 👋🏻, I would prefer those badges to be justified to left, not to center.

onuralpszr:
> Hi @onuralpszr 👋🏻, I would prefer those badges to be justified to left, not to center.

All done and fixed.

LinasKo:
As discussed, let's still think about adding badges in the future.

PR #894, opened 2024-02-13
add gdown to  pyproject.toml 

Ali-Fayzi:


onuralpszr:
I am against this package to add, supervision not depend on this and without proper reason I am closing this PR. Supervision is try to be as minimal as possible avoiding extra packages we don't need it.

PR #895, opened 2024-02-13
Add deepsort tracker

rzamarefat:
# Description

The deepsort algorithm is added to supervision/tracker directory and custimzed in that it works with the required supervision format. The base code can be found here [https://github.com/ModelBunker/Deep-SORT-PyTorch]. The main motivation is that I really like supervision and use it a lot. So, in order to have different tracking algorithms in a single source I decided to have this contribution. I believe all good and famous tracking algorithms should be integrated to supervision to make this tool perfect.

- gdown

## Type of change

-   [-] New feature (non-breaking change which adds functionality)

```python
import numpy as np
from ultralytics import YOLO

import supervision as sv

model = YOLO("yolov8n.pt")
tracker = sv.DeepSort()
bounding_box_annotator = sv.BoundingBoxAnnotator()


def callback(frame: np.ndarray, index: int) -> np.ndarray:
    results = model(frame)[0]
    detections = sv.Detections.from_ultralytics(results)
    detections = tracker.update_with_detections(detections, frame)

    labels = [f"#{tracker_id}" for tracker_id in detections.tracker_id]

    annotated_frame = bounding_box_annotator.annotate(
        scene=frame.copy(), detections=detections
    )
    annotated_frame = label_annotator.annotate(
        scene=annotated_frame, detections=detections, labels=labels
    )
    return annotated_frame


sv.process_video(
    source_path="path/to/video",
    target_path="path/to/output",
    callback=callback,
)
```


onuralpszr:
@Ali-Fayzi  thank you for your contribution but this implementation is basically related to pytorch and supervision nature shouldn't have pytorch and torchvision at all. We make sure we only implement mathematical logic and other requirement I also stated previously it needs to be pytorch free as well. So for that I am closing it. 

Thank you.

PR #907, opened 2024-02-15
Add a cookbook about alignment evaluation

iamhatesz:
# Description

This PR introduces a new cookbook related to detecting and visualizing objects in artificially generated images.

## Type of change

A new cookbook was introduced.

The notebook itself was well-tested manually. The documentation changes were tested using `mkdocs`.

pre-commit hook doesn't allow to files larger than 500 KB, but somehow other cookbooks were successfully committed. I had to skip this check, but looking forward to hearing a better solution.

## Docs

-   [X] Docs updated? What were the changes: added a link to the new cookbook as described [here](https://supervision.roboflow.com/develop/contributing/#cookbooks).


SkalskiP:
Hi @iamhatesz 👋🏻 ! 


Unfortunately, GitHub does not allow users to leave comments inside notebooks. For this reason, I've gathered my thoughts in this comment.

- Firstly, you surprised me a bit. I didn't expect that we would ever have a cookbook related to generative AI. Knowing your background, I could have anticipated it, but it's a great idea. 🔥 
- It seems that `¶` appeared at the end of the title. I would also remove `and` from the last bullet point and add links leading to the repo of all libraries, so people can easily jump there. Additionally, I think it would be great if we switched from `ultralytics` to `inference`. You can find an example of how to do this [here](https://supervision.roboflow.com/develop/notebooks/zero-shot-object-detection-with-yolo-world/).

  ```
  ## Install required packages¶
  
  In this cookbook, we'll leverage the following Python packages:
  * `diffusers` for image generation pipelines,
  * `ultralytics` for running object detection,
  * and `supervision` for visualizing detections.
  ```

- Would it make sense to update the code to make it work not only on GPU but also on CPU?

  ```
  pipeline = StableDiffusionXLPipeline.from_pretrained(
      "stabilityai/stable-diffusion-xl-base-1.0",
      torch_dtype=torch.float16,
      variant="fp16",
      use_safetensors=True,
  ).to("cuda")
  ```

- Generally, I like to keep magic numbers and strings in constants. I would recommend moving `"a black cat playing with a blue ball next to a parked white car, wide angle, photorealistic"`, `"low quality, blurred, text, illustration"`, and so on, to constants. This way, the cookbook will be consistent with the others and (it seems to me) cleaner.

  ```
  image = pipeline(
      prompt="a black cat playing with a blue ball next to a parked white car, wide angle, photorealistic",
      negative_prompt="low quality, blurred, text, illustration",
      generator=torch.manual_seed(9213799),
      num_inference_steps=50,
      width=1024,
      height=768,
  ).images[0]
  image
  ```

- FYI, This will soon be necessary. We already have an open [PR](https://github.com/roboflow/supervision/pull/875) that will add the logic to perform this conversion inside `supervision`. In the meantime, I just ask you to ensure consistent indentation.

  ```
  def pil_to_bgr_array(image: Image.Image) -> np.array:
    rgb_array = np.array(image)
    bgr_array = cv2.cvtColor(rgb_array, cv2.COLOR_RGB2BGR)
    return bgr_array
  ```

- Please, if possible, do not use abbreviations.

  ```
  gt = {"black cat", "blue ball", "white car"}
  pred = {results.names[class_id] for class_id in valid_detections.class_id}
  
  pred.issubset(gt)
  ```

iamhatesz:
Hey @SkalskiP !

Thanks for the review. I fixed most of the minor issues but didn't have time to look at the `inference` package. I can do it, but it will probably be in a few days.

Regarding running this with CPU only - it doesn't make sense for Stable Diffusion pipelines, as they take a lot of RAM due to the lack of support for FP16 types in PyTorch's CPU backend. Also, it would take an enormous amount of time to generate a single image.

Lastly, I wasn't quite convinced about:

> Please, if possible, do not use abbreviations.

because `gt` and `pred` are well-known abbreviations in the ML world, but if that's a rule, then I am fine :) However, in such a case, we shouldn't be using RGB or BGR abbreviations as well.

onuralpszr:
With @SkalskiP guidance, we bump supervision version can use new ImageType can accept pillow image and I removed helper function from notebook since we have our plot function also now support ImageType so we can use more easily. I also swap ultralytics to inference and I change couple of sections to make sure it works and rest of them all same. Thank you for your contribution @iamhatesz to cookbooks. 

I am merging in 🚀

onuralpszr (approved):

Issue #844, opened 2024-02-02
[PolygonZone] - allow `triggering_position` to be `Iterable[Position]`


SkalskiP:
### Description

Update [`PolygonZone`](https://github.com/roboflow/supervision/blob/87a4927d03b6d8ec57208e8e3d01094135f9c829/supervision/detection/tools/polygon_zone.py#L15) logic, triggering the zone when not one anchor but multiple anchors are inside the zone. This type of logic is already implemented in [`LineZone`](https://github.com/roboflow/supervision/blob/87a4927d03b6d8ec57208e8e3d01094135f9c829/supervision/detection/line_counter.py#L12).

- Rename `triggering_position` to `triggering_anchors` to be consistent with the `LineZone` naming convention.
- `Update type of argument from `Position` to `Iterable[Position]
- Maintain the default behavior. The zone should, by default, be triggered by `Position.BOTTOM_CENTER`.

### API

```python
class PolygonZone:

    def __init__(
        self,
        polygon: np.ndarray,
        frame_resolution_wh: Tuple[int, int],
        triggering_anchors: Iterable[Position] = (Position.BOTTOM_CENTER, )
    ):
        pass

    def trigger(self, detections: Detections) -> np.ndarray:
        pass
```

### Additional

- Note: Please share a Google Colab with minimal code to test the new feature. We know it's additional work, but it will speed up the review process. The reviewer must test each change. Setting up a local environment to do this is time-consuming. Please ensure that Google Colab can be accessed without any issues (make it public). Thank you! 🙏🏻

LeviVasconcelos:
Hi, I would like to work on this.

SkalskiP:
Hi @LeviVasconcelos 👋🏻 Go for it! Let us know if you need help.

LeviVasconcelos:
Hi @SkalskiP,

PR is ready for review, let me know what you think when you find time :).

Best,

onuralpszr:
This issue completed via PR #910 closing as complete

LinasKo:
Hi @JShengP :wave:

By any chance, could you attach the image you used, before annotation? Are you using a custom model, or one of our or ultralytics' pretrained models?

PetrosKataras:
Hey, also faced an issue with false positives in the case of `Position.CENTER`

I tracked it down in the clipping happening inside `PolygonZone` and specifically the `trigger` function - the way it is right now it actually modifies the bounding box coordinates before getting the anchor points which can lead to wrong calculations.

[...]

PR #910, opened 2024-02-15
Polygonzone add multiple anchors support

LeviVasconcelos:
# Description

This PR addresses [Issue #844](https://github.com/roboflow/supervision/issues/844).

- Rename `triggering_position` to `triggering_anchors` to be consistent with the `LineZone` naming convention.
- Update type of argument from `Position` to `Iterable[Position]`.
- Maintain the default behavior. The zone should, by default, be triggered by `Position.BOTTOM_CENTER`.
- Adds unit tests for `PolygonZone`.
- Updates documentation.

## Type of change

-   [ x ] New feature (non-breaking change which adds functionality)
-   [ x ] This change requires a documentation update

For a minimalist setup, I'm doing the following:

1- Create mock detections simulating an object moving in a straight line.
2- Instantiate a `PolygonZone` object with polygon partially intersecting the line.
3- Compute the `PolygonZone.trigger(mock_detections)` and annotate the detections triggering the `PolygonZone` object as green, and red otherwise.

The setup used here was also used as base for the unit tests.

Google colab: [https://colab.research.google.com/drive/1fcsVrprMuf7hm_bxitSqmeL5L7KBVGHh?usp=sharing](https://colab.research.google.com/drive/1fcsVrprMuf7hm_bxitSqmeL5L7KBVGHh?usp=sharing)

#### Option 1: Maintaining Deprecation Logic Within the Function

To ensure backward compatibility after renaming a parameter in this PR, one approach is to maintain deprecation logic within the function itself. The implementation could look something like this:

```python
class PolygonZone:
    def __init__(self, polygon, frame_resolution_wh, triggering_anchors, **kwargs):
        if 'triggering_position' in kwargs:
            warning.warn('deprecated argument...')
            ''' logic to handle deprecated parameter'''
```

While effective for specific cases, this approach has some drawbacks:

1. **Documentation Readability**: Inserting `**kwargs` might make the documentation less clear.
2. **Code Visibility**: Deprecated code may become challenging to locate, especially with scale.
3. **Lack of Precedence**: I couldn't find any examples of this pattern in the codebase.

#### Option 2: Adding a `@deprecated_parameter` Annotator (Adopted Solution)

The chosen solution involves creating a `@deprecated_parameter` annotator using pseudocode like this:

```python
def deprecated_parameter(old_param: str, new_param: str, map_func: Callable = lambda x: x):
    def decorator(func):
        @functools.wraps(func)
        def wrapper(*args, **kwargs):
            if old_param in kwargs:
                print('warning message...')
                kwargs[new_param] = map_func(kwargs.pop(old_param))
            return func(*args, **kwargs)
        return wrapper
    return decorator
```

While this solution may not be suitable for highly complex parameter changes, it aligns with existing practices, such as the use of the `@deprecated` annotator, making it easier to track deprecated code references.

Please share your thoughts on this proposed solution.

### Deprecated Examples

Some examples, like `examples/traffic_analysis`, explicitly set `triggering_position` for the `PolygonZone` class. This PR renders it deprecated.

Should we address these changes within this PR, or would it be more appropriate to handle them in a separate one? I am open to incorporating the necessary modifications here if that is the preferred approach.

Let me know your preferences regarding these considerations. 
## Docs

-   [ x ] Docs updated? What were the changes:

[Changed the documentation for class `PolygonZone`](https://github.com/roboflow/supervision/pull/910/files#diff-738bb6b6510e46b8633ab062a4c980fa338f7f3ccf7c9792b91a1bd589e8bf01L23-R27).

### DIFF
--- a/supervision/detection/tools/polygon_zone.py
+++ b/supervision/detection/tools/polygon_zone.py
@@ -23,7 +23,8 @@ class PolygonZone:
         frame_resolution_wh (Tuple[int, int]): The frame resolution (width, height)
         triggering_anchors (Iterable[sv.Position]): A list of positions specifying
             which anchors of the detections bounding box to consider when deciding on
-            whether the detection fits within the PolygonZone (default: (sv.Position.BOTTOM_CENTER,)).
+            whether the detection fits within the PolygonZone
+            (default: (sv.Position.BOTTOM_CENTER,)).
         current_count (int): The current count of detected objects within the zone
         mask (np.ndarray): The 2D bool mask for the polygon zone
     """
@@ -46,7 +47,9 @@ class PolygonZone:
         self.current_count = 0
 
         width, height = frame_resolution_wh
-        self.mask = polygon_to_mask(polygon=polygon, resolution_wh=(width + 1, height + 1))
+        self.mask = polygon_to_mask(
+            polygon=polygon, resolution_wh=(width + 1, height + 1)
+        )
 
     def trigger(self, detections: Detections) -> np.ndarray:
         """
@@ -61,7 +64,9 @@ class PolygonZone:
                 if each detection is within the polygon zone
         """
 
-        clipped_xyxy = clip_boxes(xyxy=detections.xyxy, resolution_wh=self.frame_resolution_wh)
+        clipped_xyxy = clip_boxes(
+            xyxy=detections.xyxy, resolution_wh=self.frame_resolution_wh
+        )
         clipped_detections = replace(detections, xyxy=clipped_xyxy)
         all_clipped_anchors = np.array(
             [
@@ -70,7 +75,11 @@ class PolygonZone:
             ]
         )
 
-        is_in_zone = self.mask[all_clipped_anchors[:, :, 1], all_clipped_anchors[:, :, 0]].transpose().astype(bool)
+        is_in_zone = (
+            self.mask[all_clipped_anchors[:, :, 1], all_clipped_anchors[:, :, 0]]
+            .transpose()
+            .astype(bool)
+        )
         is_in_zone = np.all(is_in_zone, axis=1)
 
         self.current_count = int(np.sum(is_in_zone))
--- a/supervision/detection/tools/polygon_zone.py
+++ b/supervision/detection/tools/polygon_zone.py
@@ -1,5 +1,5 @@
 from dataclasses import replace
-from typing import Optional, Tuple
+from typing import Iterable, Optional, Tuple
 
 import cv2
 import numpy as np
@@ -30,17 +30,15 @@ class PolygonZone:
         self,
         polygon: np.ndarray,
         frame_resolution_wh: Tuple[int, int],
-        triggering_position: Position = Position.BOTTOM_CENTER,
+        triggering_anchors: Iterable[Position] = (Position.BOTTOM_CENTER,),
     ):
         self.polygon = polygon.astype(int)
         self.frame_resolution_wh = frame_resolution_wh
-        self.triggering_position = triggering_position
+        self.triggering_anchors = triggering_anchors
         self.current_count = 0
 
         width, height = frame_resolution_wh
-        self.mask = polygon_to_mask(
-            polygon=polygon, resolution_wh=(width + 1, height + 1)
-        )
+        self.mask = polygon_to_mask(polygon=polygon, resolution_wh=(width + 1, height + 1))
 
     def trigger(self, detections: Detections) -> np.ndarray:
         """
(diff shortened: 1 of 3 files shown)


supervision/utils/internal.py
```diff
@@ -1,5 +1,37 @@
 import functools
 import warnings
+from typing import Callable
+
+
+def deprecated_parameter(
+    old_param: str, new_param: str, map_func: Callable = lambda x: x
```
SkalskiP, line 7: I like the idea. When would be the example use case when I'd use the non-default `map_func` value?

test/detection/test_polygonzone.py
```diff
@@ -0,0 +1,102 @@
...
+        [135.0, 135.0, 165.0, 165.0],
+        [160.0, 160.0, 190.0, 190.0],
+        [185.0, 185.0, 215.0, 215.0],
+        [210.0, 210.0, 240.0, 240.0],
+        [235.0, 235.0, 265.0, 265.0],
+    ],
+    dtype=np.float32,
+)
+
+DETECTIONS = sv.Detections(
```
SkalskiP, line 23: Consider using the `mock_detections` function here. You can find it in `test/test_utils.py`.

test/detection/test_polygonzone.py
```diff
@@ -0,0 +1,102 @@
...
+    polygon_zone: sv.PolygonZone,
+    expected_results: np.ndarray,
+    exception: Exception,
+) -> None:
+    with exception:
+        in_zone = polygon_zone.trigger(detections)
+        assert np.all(in_zone == expected_results)
+
+
+def test_polygon_zone_deprecated() -> None:
```
SkalskiP, line 94: I appreciate that you also tested this case, but personally, I would remove this test case.

supervision/detection/tools/polygon_zone.py
```diff
@@ -20,21 +21,29 @@ class PolygonZone:
...
-            box that triggers the zone (default: Position.BOTTOM_CENTER)
+        triggering_anchors (Iterable[sv.Position]): A list of positions specifying
+            which anchors of the detections bounding box to consider when deciding on
+            whether the detection fits within the PolygonZone
+            (default: (sv.Position.BOTTOM_CENTER,)).
         current_count (int): The current count of detected objects within the zone
         mask (np.ndarray): The 2D bool mask for the polygon zone
     """
 
+    @deprecated_parameter(
```
SkalskiP, line 32: I'd love to have a way to attach a custom message, for example, mentioning the version of the `supervision` package when support for deprecated parameter names will be removed.

supervision/utils/internal.py
SkalskiP: Will this work with MyPy?

SkalskiP (changes requested):

SkalskiP:
Hi @LeviVasconcelos 👋🏻 ! Awesome PR. I liked your ideas. I left some comments. I'd appreciate it if you consider my feedback.

supervision/utils/internal.py
LeviVasconcelos: Hi!
> I like the idea. When would be the example use case when I'd use the non-default map_func value?

The function is intended to map between types, for example, we use it to map from `Position` to `Iterable[Position]` [here](Iterable[Position]). It could also be used in more complex parameter changes like bounding_box formats e.g. [x1, y1, x2, y2] -> [x_center, y_center, w, h].

> Will this work with MyPy?

Good question. To further analyze, please consider this MyPy [playground sample](https://mypy-play.net/?mypy=latest&python=3.12&gist=b6cc4856cd0a2a0a2270c0e69a2bb80b).
MyPy finds no issues with the code. However, It is still possible to use the `map_function` to cast the old parameter to a inconsistent type: [example](https://mypy-play.net/?mypy=latest&python=3.12&gist=d486f2a2f2420c82f84df9e85ea889cd). It also does not find any issues in the second case.

test/detection/test_polygonzone.py
LeviVasconcelos: Thanks, Done!

supervision/detection/tools/polygon_zone.py
LeviVasconcelos: Thansk, done. Please have a look: https://github.com/roboflow/supervision/pull/910/files#diff-b15cd2cdcbbed3ddb5617e22208f7e084b583346acaf59ca8b9dcbbf7e85a49bR10

Let me know what you think about this solution.

LeviVasconcelos:
Thank you for your comments! I've addressed all of them. 

Please let me know what you think

supervision/detection/tools/polygon_zone.py
SkalskiP: We usually add a message like this: `Color.white()` is deprecated and will be removed in `supervision-0.22.0`. Use `Color.WHITE` instead.` Could we add deprecation drop version as well?

supervision/utils/internal.py
```diff
@@ -1,5 +1,78 @@
 import functools
 import warnings
+from typing import Callable
+
+
+def deprecated_parameter(
+    old_parameter: str,
+    new_parameter: str,
+    map_function: Callable = lambda x: x,
+    warn_message: str = "Warning: '{old_parameter}' in '{function_name}' "
```
SkalskiP, line 10: Please do change `warn_message` to `warning_message`. I try to avoid using abbreviated names.

SkalskiP (changes requested):

supervision/detection/tools/polygon_zone.py
```diff
@@ -20,21 +21,29 @@ class PolygonZone:
...
+        old_parameter="triggering_position",
+        new_parameter="triggering_anchors",
+        map_function=lambda x: [x],
+    )
```
SkalskiP, line 36: Add information about the version when we will drop support. You can say we will drop in `supervision-0.23.0`.

supervision/utils/internal.py
SkalskiP: Understood! Awesome!

onuralpszr:
@LeviVasconcelos  thank you for your contribution I am merging in. For examples I did not touch it. There is little bit more work I see in example let's do them outside.

onuralpszr (approved):

onuralpszr (approved):

SkalskiP:
Thanks for your help @LeviVasconcelos 👋🏻 ! I really like your `deprecated_parameter`. I plan to use it a lot!

Issue #928, opened 2024-02-20
[Smoother] - `tracker_id` is `None` when there is no detection for an extended period


vhuard:
### Bug

I have code that performs detections on video, which operates smoothly without the smoother function. However, upon integrating the smoother function, I encounter the following error:

```
Traceback (most recent call last):
  File "/content/script_detection.py", line 165, in <module>
    processor.process_video()
  File "/content/script_detection.py", line 67, in process_video
    annotated_frame = self.process_frame(frame)
  File "/content/script_detection.py", line 121, in process_frame
    return self.annotate_frame(frame, detections)
  File "/content/script_detection.py", line 86, in annotate_frame
    in zip(detections.tracker_id, detections.confidence, detections.class_id)
TypeError: 'NoneType' object is not iterable
```

### Environment

- Supervision: 0.18.0

### Minimal Reproducible Example

```python
def annotate_frame(
    self, frame: np.ndarray, detections: sv.Detections
) -> np.ndarray:
    annotated_frame = frame.copy()

    labels = [
            f"#{tracker_id} {class_id}"
            for tracker_id, confidence, class_id
            in zip(detections.tracker_id, detections.confidence, detections.class_id)
    ]

    annotated_frame = self.trace_annotator.annotate(
        scene=annotated_frame, detections=detections
    )
    annotated_frame = self.bounding_box_annotator.annotate(
        scene=annotated_frame, detections=detections
    )
    annotated_frame = self.label_annotator.annotate(
        scene=annotated_frame, detections=detections, labels=labels
     )
    annotated_frame = self.percentage_bar_annotator.annotate(
        scene=annotated_frame, detections=detections
    )


    return annotated_frame

def process_frame(self, frame: np.ndarray) -> np.ndarray:
    results = self.model(
        frame, verbose=False, conf=self.conf_threshold, iou=self.iou_threshold
    )[0]
    detections = sv.Detections.from_ultralytics(results)
    detections = detections[detections.confidence > args.confidence_threshold]
    detections = detections.with_nms(threshold=args.iou_threshold)
    detections = self.tracker.update_with_detections(detections=detections)
    detections = self.smoother.update_with_detections(detections=detections)
   

    return self.annotate_frame(frame, detections)
```

LinasKo:
This will happen if there's no detections or all detections go out of view for long enough. `Detections.empty()` only initialized `xyxy`, `confidence` and `class_id` - not `tracker_id`.

vhuard:
[...]

Issue #943, opened 2024-02-25
`Detections.empty()` invalidates detections, causes crashes when `Detections.merge()` is called.


LinasKo:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar bug report.

### Bug

This is a more general version of #928. I believe it is important enough to deserve a separate issue.

Proof found in [this Colab](https://colab.research.google.com/drive/1ktj_CIlM9mcmboo8LJUCsYgKlaPN8bJM?usp=sharing).

### Explanation
`Detections` contains these variables:
```python
xyxy: np.ndarray
mask: Optional[np.ndarray] = None
confidence: Optional[np.ndarray] = None
class_id: Optional[np.ndarray] = None
tracker_id: Optional[np.ndarray] = None
data: Dict[str, Union[np.ndarray, List]] = field(default_factory=dict)
```

Suppose we call `Detections.merge([detections_1, detections_2])`. An error is rightfully raised when the same variable is defined differently - it is `None` in `detections_1` and `np.array([])` in `detections_2`. It makes sense as we don't want to merge detections where, for example "no mask was detected" and "masks should never be detected" - that's the meaning of "None."

However, when calling `Detections.empty()`, frequently used to initialize the no-detections case, it sets a selective subset of the fields - only the `xyxy`, `condifence` and `class_id` are set.
Many models use other fields, some use trackers. If you merge with an empty detections set at some point, you're very likely to raise the aforementioned error.

Ultimately, `Detections.empty()` needs to distinguish between "fields that shouldn't exist" and "fields that are empty" and adapt that based on the type of model used / operations applied (e.g. tracking).

### Minimal Reproducible Example

Again, the Colab [the Colab](https://colab.research.google.com/drive/1ktj_CIlM9mcmboo8LJUCsYgKlaPN8bJM?usp=sharing) shows examples.

### Additional

I can help submit a PR, but this is a non-trivial bug, affecting at least 6 files, with effects on others.

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

LinasKo:
# Affected Code:
The good news is - with the current feature set, it mostly happens when `Detections.merge` is called on results from _unusual models_  - where `confidence` is missing, where `data` is set, where there's a `mask` variable but no detections.

Affected functions:
* `supervision/detection/core.py -> from_inference()`
* `supervision/detection/core.py -> from_roboflow()`
[...]

PR #944, opened 2024-02-25
Fix smoother removing tracking_id from detections

LinasKo:
Addresses issue #928 matching implementation in ByteTrack, but sidestepping the underlying cause outlined in #943.

# Description

Smoother takes as input Detections with a defined `tracker_id`, but when no detections are left after smoothing, it would return an object where `tracker_id` is `None`.

When accessing such detections, for example, using `zip(detections.class_id, detections.tracker_id)`, a `NoneType is not iterable` error would be raised.

Similar to implementation of `ByteTrack.update_with_detections()` does it, the `tracker_id` is added explicitly in cases where it's not present already (in fact - only when no detections are left).

Fixes issue: #928
Related Issue: #943

## Type of change

Input Video:
https://github.com/roboflow/supervision/assets/6500785/d61b7e9d-2992-46d0-9cd1-6a02388dbe72

Test Code:
<details>
  <summary>Test code</summary>
  
    import supervision as sv
    from pathlib import Path
    from ultralytics import YOLO
    
    VIDEO_FILE = Path('data', 'test-traffic.mp4')
    VIDEO_OUT = Path('data', 'test-traffic-annotated.mp4')
    
    
    # Only a regression test. PIL will never be used to open video files.
    trace_annotator = sv.LabelAnnotator()
    
    model_det = YOLO("yolov8n")
    video_info = sv.VideoInfo.from_video_path(video_path=str(VIDEO_FILE))
    frames_generator = sv.get_video_frames_generator(source_path=str(VIDEO_FILE))
    tracker = sv.ByteTrack()
    smoother = sv.DetectionsSmoother()
    
    with sv.VideoSink(target_path=str(VIDEO_OUT), video_info=video_info) as sink:
        for frame in frames_generator:
            result = model_det(frame)[0]
            smoother.update_with_detections(detections)
            detections = smoother.get_smoothed_detections()
    
            # print(detections.tracker_id, detections.class_id)
            labels = [
                f"#{tracker_id} {class_id}"
                for tracker_id, class_id
                in zip(detections.tracker_id, detections.class_id)
            ]
    
                scene=frame.copy(),
                detections=detections,
                labels=labels
            )

</details>

None.

## Docs

No updates.

### DIFF
--- a/supervision/detection/tools/smoother.py
+++ b/supervision/detection/tools/smoother.py
@@ -115,4 +115,8 @@ class DetectionsSmoother:
             if track is not None:
                 tracked_detections.append(track)
 
-        return Detections.merge(tracked_detections)
+        detections = Detections.merge(tracked_detections)
+        if len(tracked_detections) == 0:
+            detections.tracker_id = np.array([], dtype=int)
+
+        return detections
--- a/supervision/detection/tools/smoother.py
+++ b/supervision/detection/tools/smoother.py
@@ -116,7 +116,7 @@ class DetectionsSmoother:
                 tracked_detections.append(track)
 
         detections = Detections.merge(tracked_detections)
-        if len(tracked_detections) == 0:
+        if len(detections) == 0:
             detections.tracker_id = np.array([], dtype=int)
 
         return detections


LinasKo:
Ready for review.

supervision/detection/tools/smoother.py
```diff
@@ -115,4 +115,8 @@ def get_smoothed_detections(self) -> Detections:
             if track is not None:
                 tracked_detections.append(track)
 
-        return Detections.merge(tracked_detections)
+        detections = Detections.merge(tracked_detections)
+        if len(tracked_detections) == 0:
```
SkalskiP, line 119:
```suggestion
        if len(detections) == 0:
```

SkalskiP: Hi @LinasKo 👋🏻 ! Great find! That's probably the only change I'd make.

SkalskiP (approved):

supervision/detection/tools/smoother.py
LinasKo: Done :+1:

Also verified the `__len__` implementation just in case. Seems to be alright - can't imagine changes to `Detections.empty` that would break this.

SkalskiP: I'm considering modifying the `Detections.empty` API to include arguments such as `with_confidence`, `with_tracker_id`, etc. So that in the future, in a similar situation, we would only need to do `Detections.empty(with_tracker_id = True)`.

LinasKo: The problem with that (#943, Option 4) is that intermediary functions that would also need `with_confidence` and `with_tracker_id`, or you need to forbid them from ever accepting zero-length iterables.

We're actually at the best place to show this.

If nothing is detected, `detections = Detections.merge(tracked_detections)` takes `[]`. It then tries to call `Detections.empty` - but what params should it set?..

PR #981, opened 2024-03-07
Add pattern matching

AlexandreKempf:
# Description

:wave: Hello here

In many real-world scenarios, training a detection model, deploying it, and getting predictions is only one half of the job. Processing of the predictions is the other half of data scientists' work, and Supervision is the only tool to help do that in object detection. 
However, thinking about my previous jobs and what I needed, I realized that there was a part of post-processing that was not covered by Supervision: looking for patterns. We can think of it as a regex for object detection: Given a set of rules, give me the set of boxes that respect these rules.

For instance in this image with predictions:
g15

That we obtained with:
```python
    import cv2
    import supervision as sv
    from ultralytics import YOLO

    image = cv2.imread("men_on_bicycle.jpg")
    model = YOLO('yolov8s.pt')

    classes = [person, bicycle, ...]

    result = model(image)[0]
```

We could add a pattern to find an individual with their bike. The rule to identify this pattern (the regex) would be:
 - there should be a person
 - there should be a bike
 - the person center and the bike should be aligned vertically
 
 ```python
     pattern = sv.MatchPattern(
        [
            (lambda id: id == classes.index("bicycle") , ["Bike.class_id"]),
            (lambda id: id == classes.index("person") , ["Rider.class_id"]),
            (lambda bike_center, rider_center: abs(bike_center[0] -  rider_center[0]) <= 50),
                ["Bike.CENTER", "Rider.CENTER"]
            ),
        ]
    )
    matches = pattern.match(detections)
```

`matches` is a list of detections (`List[sv.Detections]`) that match the pattern. Once displayed, the 4 first matches from the list look like this:

g30

For each match, an additional `data` attribute named "match_name" is added to the `sv.Detection` to specify to which detection corresponds each object of the pattern ("Bike" and "Rider" in our case). 

### Additional notes

1 - I added a little sugar on top of the feature. It is possible to define a value instead of a function. In that case, it will ensure that the arguments match exactly the value you provide.
 ```python
     pattern = sv.MatchPattern(
        [
            (1 , ["Bike.class_id"]),
            (0 , ["Rider.class_id"]),
            (lambda bike_center, rider_center: abs(bike_center[0] -  rider_center[0]) <= 50),
                ["Bike.CENTER", "Rider.CENTER"]
            ),
        ]
    )
    matches = pattern.match(detections)
```

2 - The arguments you can give to the function directly come from `sv.Detections`. The complete list is:
 - `xyxy`, `mask`, `class_id`, `confidence`, or `tracker_id`
 - one of the `Position` enum strings
 - a field from the `data` attribute of your detections for more custom use cases

3 - It we want to continue this feature, it would be nice to provide a small bank of function that the user can use out of the box for bounding boxes (`iou`, `included_in`, etc.) and segmentation (`next_to`, `area_are_touching`, etc.)

## Type of change

I added functional tests.

## Docs

-   [ ] Docs updated? What were the changes? I wanted to update the document, but I couldn't find a way to host images in the long run. Your documentation images are hosted on https://media.roboflow.com, but I can't access them. How are you usually doing it for people who don't have access to internal resources?

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2360,13 +2360,13 @@ test = ["flaky", "ipykernel (>=6.19.3)", "ipython", "ipywidgets", "nbconvert (>=
 
 [[package]]
 name = "nbconvert"
-version = "7.16.1"
+version = "7.16.2"
 description = "Converting Jupyter Notebooks (.ipynb files) to other formats.  Output formats include asciidoc, html, latex, markdown, pdf, py, rst, script.  nbconvert can be used both as a Python library (`import nbconvert`) or as a command line tool (invoked as `jupyter nbconvert ...`)."
 optional = false
 python-versions = ">=3.8"
 files = [
-    {file = "nbconvert-7.16.1-py3-none-any.whl", hash = "sha256:3188727dffadfdc9c6a1c7250729063d7bc78b355ad7aa023138afa030d1cd07"},
-    {file = "nbconvert-7.16.1.tar.gz", hash = "sha256:e79e6a074f49ba3ed29428ed86487bf51509d9aab613bd8522ac08f6d28fd7fd"},
+    {file = "nbconvert-7.16.2-py3-none-any.whl", hash = "sha256:0c01c23981a8de0220255706822c40b751438e32467d6a686e26be08ba784382"},
+    {file = "nbconvert-7.16.2.tar.gz", hash = "sha256:8310edd41e1c43947e4ecf16614c61469ebc024898eb808cce0999860fc9fb16"},
 ]
 
 [package.dependencies]


SkalskiP:
Hi @AlexandreKempf 👋🏻 Thanks for your interest in supervision. Is there any reason you decided to close your PR?

AlexandreKempf:
Hi @SkalskiP,
Sorry it was a problem on m'y end. To fix it,
I created a very similar PR that is open and waiting for review ;)

AlexandreKempf:
Hi @SkalskiP,
Sorry it was a problem on my end. To fix it,
I created a very similar PR that is open and waiting for review ;)

PR #1007, opened 2024-03-15
[ImageMaskAnnotator]: Adds an annotator to overlay an image over specified classes of detections

Log45:
# Description

ImageMaskAnnotator is a new annotator that overlays a user-specified image on detections included in a user-specified list of classes which defaults to class 0 (person for MS-COCO). Specified images can be PNGs, but with clear backgrounds, they are not guaranteed to cover the entire detection. Another barrier is using images with pure-black pixels which often gets turned clear after the overlay. 

Motivation for this change was to allow users to redact certain objects or add anonymity to their products in a more customizable way than simply blurring specific detections.

One issue (that is non-breaking) that I would appreciate feedback on is that an IndexError occurs if any point of an object's `xyxy` attribute is outside of the frame, which I mitigated with a try/except loop, but the object is no longer covered in that exception. (Maybe I could add variables to have access to previous positions, but I'd like feedback on how to cover the entire rest of the detection that is inside of the frame). 

This change does not require any additional dependencies.

## Type of change

-   [x] This change requires a documentation update (I updated annotators.md)

There are no unit tests since the change is an additional annotator, but I've tested on my local machine and on colab and it works as intended. Attached is a colab notebook with an example of the change (inspired by @dfreilich 's notebook example): 

https://colab.research.google.com/drive/1FkIwwDaSPVvSk2eKRbTeqHT89oZIbCFk?usp=sharing 

None that I can think of.

## Docs

annotators.md was updated to include the new `ImageMaskAnnotator`

### DIFF
--- a/docs/annotators.md
+++ b/docs/annotators.md
@@ -395,6 +395,21 @@ status: new
 
     </div>
 
+=== "ImageMask"
+
+    ```python
+    import supervision as sv
+
+    frame = ...
+    detections = sv.Detections(...)
+
+    image_annotator = sv.ImageMaskAnnotator("image.png")
+    annotated_frame = image_annotator.annotate(
+        scene=frame.copy(),
+        detections=detections
+    )
+    ```
+
 <div class="md-typeset">
     <h2><a href="#supervision.annotators.core.BoundingBoxAnnotator">BoundingBoxAnnotator</a></h2>
 </div>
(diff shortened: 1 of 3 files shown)


SkalskiP:
Hi, @Log45 👋🏻 Thank you for your interest in `supervision`. I apologize for the late reply. I've been a bit busy with responsibilities unrelated to `supervision`. I have a few comments. Would you be willing to make the corrections?

Log45:
@SkalskiP bumping because I'm not sure if you were notified when I last replied, I'm open to any feedback!

LinasKo:
Hey @Log45 👋 

Apologies for the delay. I found your annotator proposal really interesting, and I appreciate that you included a specific use case—it gives us a better sense of the challenges our users are tackling and the projects they’re working on.

That said, I haven’t come across any requests to fully cover images by overlaying another. Especially since the last release, where `IconAnnotator` was introduced.

I’ll have to close this one for now, but thank you for your efforts in making our library better! 🤝

Issue #754, opened 2024-01-18
[ByteTrack] - redesign `update_with_detections` to use IoU to match input `Detections` with `tracker_id`


SkalskiP:
### Description

- Currently, `update_with_detections` returns the predicted position of boxes, not their actual coordinates received from the detector. Many users have complained about the deterioration of box quality when using ByteTrack. ([#743](https://github.com/roboflow/supervision/issues/743))
- `ByteTrack` does not work with segmentation models because masks are not transferred to the `update_with_detections` output.
- The `Detections.data field` is lost after passing through `update_with_detections`.

All these issues can be resolved by changing the logic in `update_with_detections`. Instead of mapping values obtained from `update_with_tensors` to new `Detections` objects, we should use IoU to map the results of `update_with_tensors` to input `Detections` objects. This way, the input `xyxy` coordinates and the input state of the `mask` and `data` fields will be preserved.

For this purpose, we can utilize the already existing function [`box_iou_batch`](https://github.com/roboflow/supervision/blob/3024ddca83ad837651e59d040e2a5ac5b2b4f00f/supervision/detection/utils.py#L31). The matching procedure has been demonstrated in [one](https://www.youtube.com/watch?v=OS5qI9YBkfk) of our videos on YouTube.

### Additional

- Note: Please share a Google Colab with minimal code to test the new feature. We know it's additional work, but it will definitely speed up the review process. Each change must be tested by the reviewer. Setting up a local environment to do this is time-consuming. Please ensure that Google Colab can be accessed without any issues (make it public). Thank you! 🙏🏻

rolson24:
Hi @SkalskiP!
Can I work on this issue? I think I can implement this in the way you want without modifying update_with_tensors and in a pretty simple way.

SkalskiP:
Hi @rolson24 👋🏻! Feel free to submit PR. You will make me very happy.

SkalskiP:
The solution for this issue is implemented via https://github.com/roboflow/supervision/pull/1035. I am closing the issue. Thanks @rolson24 ! 🔥

PR #1034, opened 2024-03-23
Redesign update with detections to match detections with tracker

rolson24:
# Description

This change addresses this issue: (https://github.com/roboflow/supervision/issues/754)
`update_with_detections` now returns the original detections that were tracked by ByteTrack with their original bboxes, masks, class_ids, data fields and with updated track_ids. Detections that were not tracked this frame are not included in the returned detections.

## Type of change

-   [x] New feature (non-breaking change which adds functionality) - Segmentation tracking

I have tested this change by running object tracking for both object detection and segmentation in this [colab notebook](https://colab.research.google.com/drive/1azgc_A-divM3Z5l0TyHW2SRHnw6EUlbO?usp=sharing)

segmentation_tracking

Tracking docs should be updated to reflect that ByteTrack can now do segmentation tracking.

## Docs

### DIFF
--- a/supervision/tracker/byte_tracker/core.py
+++ b/supervision/tracker/byte_tracker/core.py
@@ -275,22 +275,32 @@ class ByteTrack:
 
         tracks = self.update_with_tensors(tensors=tensors)
 
+        final_detections = Detections.empty()
+
         if len(tracks) > 0:
             det_tlbrs = np.asarray([det[:4] for det in tensors])
+            print(f"  det_tlbrs: {det_tlbrs}")
             track_tlbrs = np.asarray([track.tlbr for track in tracks])
+            print(f"track_tlbrs: {track_tlbrs}")
 
             ious = box_iou_batch(det_tlbrs, track_tlbrs)
 
             iou_costs = 1 - ious
 
-            matches, _, _ = matching.linear_assignment(iou_costs, 0.25)
+            matches, _, _ = matching.linear_assignment(iou_costs, 0.5)
 
             for idet, itrack in matches:
-                detections.tracker_id[idet] = int(tracks[itrack].track_id)
-        else:
-            detections.tracker_id = np.array([], dtype=int)
+                print(f"det xyxy: {detections.xyxy[idet]}")
+                print(f"track xyxy: {tracks[itrack].tlbr}")
+                print(" ")
+                current_detection = detections[idet]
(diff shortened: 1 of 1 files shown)


rolson24:
Sorry about the number of commits. I tried to squash some together but messed up a bit. I can delete and resubmit if you need me to.

supervision/tracker/byte_tracker/core.py
```diff
@@ -269,28 +270,39 @@ def callback(frame: np.ndarray, index: int) -> np.ndarray:
...
-            detections.class_id = np.array(
-                [int(t.class_ids) for t in tracks], dtype=int
-            )
-            detections.tracker_id = np.array(
-                [int(t.track_id) for t in tracks], dtype=int
-            )
-            detections.confidence = np.array(
-                [t.score for t in tracks], dtype=np.float32
-            )
+            det_tlbrs = np.asarray([det[:4] for det in tensors])
```
SkalskiP, line 281: Let's not use abbreviations. They make the code harder to read. To be honest I don't know what `det_tlbrs` or `track_tlbrs` means :)

SkalskiP (changes requested):

SkalskiP:
Hi, @rolson24 👋🏻 ! Thanks a lot for this PR. I left a few comments (nothing serious). I would also like you to open a new PR with just one commit (if possible). I can use `Squash and merge` option but I'm not sure if GitHub will assign this commit to you :/ And I would want to give you credit for this work.

rolson24:
Great! I made those changes and can submit a new PR.
Before I do though, I am wondering if we should add a method to the STrack class to get the actual bounding box from the detection that was used to update the track. Right now STrack.tlbr returns the predicted value of the bounding box. This means that when we do the matching in the update_with_detections from the original detections to the tracks, the wrong track might be matched to the original bounding box when there are two detections that are really close together. This won't matter for tracking performance, but it technically might return the wrong original detection in that frame. This also will allow us to not use the linear assignment function (in-efficient) and instead just search for the ious that are 1.0 because the original detections will match perfectly with the detections we get from the tracks.

SkalskiP:
Hi @rolson24 👋🏻 ! Awesome! In this case, I'll close this PR, and we will focus on https://github.com/roboflow/supervision/pull/1035. Thanks for being so cooperative 🙏🏻

Issue #1044, opened 2024-03-25
[ByteTrack] add `minimum_consecutive_frames` to limit the number of falsely assigned tracker IDs


SkalskiP:
### Description

Expand the `ByteTrack` API by adding the `minimum_consecutive_frames` argument. 
It will specify how many consecutive frames an object must be detected to be assigned a tracker ID. This will help prevent the creation of accidental tracker IDs in cases of false detection or double detection. Until detection reaches `minimum_consecutive_frames`, it should be assigned `-1`.

### API

```python
class ByteTrack:
    def __init__(
        self,
        track_activation_threshold: float = 0.25,
        lost_track_buffer: int = 30,
        minimum_matching_threshold: float = 0.8,
        frame_rate: int = 30,
        minimum_consecutive_frames: int = 1
    ):
        pass
```

### Additional

- Note: Please share a Google Colab with minimal code to test the new feature. We know it's additional work, but it will speed up the review process. The reviewer must test each change. Setting up a local environment to do this is time-consuming. Please ensure that Google Colab can be accessed without any issues (make it public). Thank you! 🙏🏻

rolson24:
Hi @SkalskiP,

I would be interested in working on this. I have a question about possible implementation, though.
Would we want to keep the tracking algorithm the same and just change the tracker ID (meaning possibly returning multiple tracks with a track_id of -1), or should we change the algorithm to only return tracks that have valid tracker ID's?

rolson24:
Nvm, I think the best option is to only return tracks that have valid tracker ID's so we don't have to change how the annotators work.

SkalskiP:
Yup. Let's drop them for now.

PR #1050, opened 2024-03-26
Add minimum_consecutive_frames argument to ByteTrack

rolson24:
# Description

This PR fixes https://github.com/roboflow/supervision/issues/1044. I decided to implement this by keeping the base ByteTrack algorithm the same, but changing when a track gets assigned a new track_id, and only returning tracks that have valid track_ids (>= 0). This change also resets the track_id counter when a new ByteTrack object is initialized.

fix bug from PR #1035 

change update_with_tensors to not return tracks with track_id < 0

reset the track_id counter when initializing a new ByteTrack object

## Type of change

I tested this by running ByteTrack with multiple different values of minimum_consecutive_frames and looking at the output video to see if the short, low-quality tracks were removed. In this [colab notebook](https://colab.research.google.com/drive/1hM1trTUQfFO07znnI77wJPcziiNotCBK?usp=sharing), you can run the video at both minimum_consecutive_frames = 1 and 5 and compare the two output videos. You can also observe the maximum track_id in every frame to see that the tracker with a minimum_consecutive_frames value of 5 has much lower values indicating that there are fewer valid tracks. The number of valid tracks is also more stable.

This might require updating the documentation for ByteTrack, which I am willing to do if this is approved.

## Docs

-   [ ] Docs updated?


SkalskiP:
Hi @rolson24 👋🏻 Looks like there are some unresolved conflicts. Could you take care of this?

rolson24:
This most recent commit fixes this issue: (https://github.com/roboflow/supervision/pull/1049#issuecomment-2026165713)

If all of the detections in a frame have low confidence, then the tracker will still return these detections, but with an empty tracker_id array. This causes issues when you try to get the track_id of one of these detections (for example, using the trace_annotator). I have fixed this by simply returning empty detections with a tracker_id array of length 0. This is the behavior of the original update_with_detections() function.

I have tested this change by creating unit tests that cover all of the possible cases for detection inputs. See this [colab notebook](https://colab.research.google.com/drive/1CdoOVn7Y0VVlykDQJzWV8vvWapn77VC9?usp=sharing).

SkalskiP:
Hi @rolson24 👋🏻 is this ready for testing and review?

rolson24:
After further testing, the whole idea of using track_ids to signify whether a track is valid or not will not work. This is because joint_stracks and sub_stracks rely on track_ids to prevent duplicate tracks from being in the lists. Right now this issue is causing only one track to be initialized in each frame. We would either have to modify those functions, or instead (probably better) only mark tracks that have existed for at least `minimum_consecutive_frames` as being "activated". This will allow them to keep unique track_id's and also allow us to easily filter out tracks that are valid. I can implement this and test it, before submitting a new PR

PR #1057, opened 2024-03-27
Added homography mappings to heatmap annotator

tobieabel:
# Description

Hi - first time contributor (to any open source project) and a self taught amateur when it comes to coding so be kind!  Avid user of supervision and Roboflow tools so hopefully this is useful.
Added some cv2 code for creating a homography mapping and then applying it to the heatmap annotator.  Motivation for this is that many clients I work with want to see the heatmap annotations on an overlay of a floorplan rather than from the point of view of the camera.  This is a standard 'footfall' use case when tracking people in retail or warehouse environments.

## Type of change

Sorry, didn't really understand how I am meany to provide a test case, in your documentation it simply says 'we use pytest'.  I have tested this in my own streamlit app which I use with clients and have a simple unit test I use which I am happy to submit if told how, but I don't think it would work with your setup.

YOUR_ANSWER

None that I know of

## Docs

Added to the doc strings for the heatmap annnotator in a way that made sense to me but may be to wordy for you guys based on your other examples

### DIFF
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -1274,7 +1274,9 @@ class HeatMapAnnotator:
     """
     A class for drawing heatmaps on an image based on provided detections.
     Heat accumulates over time and is drawn as a semi-transparent overlay
-    of blurred circles.
+    of blurred circles.  If homography mappings are provided the annotations
+    will be drawn on the transformed coordinates, typically used to transform
+    the image angle into a top-down view used in floor plans
     """
 
     def __init__(
@@ -1285,6 +1287,7 @@ class HeatMapAnnotator:
         kernel_size: int = 25,
         top_hue: int = 0,
         low_hue: int = 125,
+        mappings: list = None,
     ):
         """
         Args:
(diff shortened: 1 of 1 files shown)


onuralpszr:
@tobieabel  could you please create google collab for us to see it looks as well ? visual result would be great

tobieabel:
Hi @onuralpszr I'll give it a go, not used colab much before and not sure how to use the video and images I've created mappings for within the colab environment, they're currently on my hard-drive.  I'll let you know if its successful

vignesh1507 (approved):
Appreciate the effort & the code also works well. thanks for adding comments in the code to understand the overall code easily.

LinasKo:
Hi @tobieabel 👋

I appreciate it's been a long time since you made the contribution. Thank you for taking the time! I like it as it reveals a new way to use the annotators while also attaching a tangible real-world use case to it.

Unfortunately, we won't be including it this time.

1. The annotators have exactly one responsibility - annotating an image. Doing perspective transform to change the image itself is too complicated and should be done outside of an annotator. I can see this residing in `utils/image.py`, together with good documentation about how to use it (there we could mention the annotators).

2. A Colab with visualisations, for example - based off [Starter Template](https://colab.research.google.com/drive/1rin7WrS-UvVIe-_Gfxmu-yVslGphOq89?usp=sharing) should also be provided.

I believe this is a bit far from a PR we'd include, so will be closing it this time. Still, I'm grateful for revealing a very cool use case and a feature we ought to add sooner or later 😉

PR #1059, opened 2024-03-28
fixed a bug in get_data_item for np.array index

MoHoss007:
# Description

fixed a bug in get_data_item for np.array index, there was a bug when trying to mask detections using np.array and the list values in data field were not getting updated properly.

## Type of change

## How has this change been tested, please provide a testcase or example of how you tested the change.

It was tested as part of my personal project. example:

```python
# detections was created before:
detections.data["test_data"] = [None for _ in range(len(detections))]
detections = detections[detections.confidence > 0.5]
# this would lead to error in validation function without the current changes, since the size of data["test_data"] will not change.
```

N/A

## Docs

-  N/A

### DIFF
--- a/supervision/detection/utils.py
+++ b/supervision/detection/utils.py
@@ -845,8 +845,10 @@ def get_data_item(
         elif isinstance(value, list):
             if isinstance(index, slice):
                 subset_data[key] = value[index]
-            elif isinstance(index, (list, np.ndarray)):
+            elif isinstance(index, list):
                 subset_data[key] = [value[i] for i in index]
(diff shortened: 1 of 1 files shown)


SkalskiP:
Hi, @MoHoss007 👋🏻 thanks a lot for your interest in supervision and the time you spend working on this PR. It looks like we just merged alternative PR to solve this issue. You can learn more from the links below. 

- https://github.com/roboflow/supervision/issues/1061
- https://github.com/roboflow/supervision/pull/1062
- https://github.com/roboflow/supervision/pull/1063

Issue #1061, opened 2024-03-28
Detections _get_item_() does not handle index arrays of booleans correctly for data dictionary in Detections object.


rolson24:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar bug report.

### Bug

When I tried to use boolean array indexing to get specific detections from a Detections object, the data field sometimes returns the wrong values (everything is the second value) and sometimes it just doesn't work. See output of code:
```
valid detections data: {'random': array([0.69156142, 0.43520246, 0.08080701, 0.32277943]), 'Dataset': ['Roboflow 100', 'Roboflow 100', 'Roboflow 100', 'Roboflow 100']}
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
[<ipython-input-6-6cae6894f883>](https://localhost:8080/#) in <cell line: 31>()
     29 # This does not run
     30 detection_inds = np.asarray([True, False, True, False], dtype = bool)
---> 31 print(f"valid detections data: {detections[detection_inds].data}")

4 frames
[/usr/local/lib/python3.10/dist-packages/supervision/detection/core.py](https://localhost:8080/#) in __getitem__(self, index)
    928         if isinstance(index, int):
    929             index = [index]
--> 930         return Detections(
    931             xyxy=self.xyxy[index],
    932             mask=self.mask[index] if self.mask is not None else None,

/usr/local/lib/python3.10/dist-packages/supervision/detection/core.py in __init__(self, xyxy, mask, confidence, class_id, tracker_id, data)

[/usr/local/lib/python3.10/dist-packages/supervision/detection/core.py](https://localhost:8080/#) in __post_init__(self)
     88 
     89     def __post_init__(self):
---> 90         validate_detections_fields(
     91             xyxy=self.xyxy,
     92             mask=self.mask,

[/usr/local/lib/python3.10/dist-packages/supervision/detection/utils.py](https://localhost:8080/#) in validate_detections_fields(xyxy, mask, class_id, confidence, tracker_id, data)
    750     validate_confidence(confidence, n)
    751     validate_tracker_id(tracker_id, n)
--> 752     validate_data(data, n)
    753 
    754 

[/usr/local/lib/python3.10/dist-packages/supervision/detection/utils.py](https://localhost:8080/#) in validate_data(data, n)
    724         if isinstance(value, list):
    725             if len(value) != n:
[...]

PR #1062, opened 2024-03-28
fix Detections.get_data_item() bug (#1061)

rolson24:
# Description

When I tried to use boolean array indexing to get specific detections from a Detections object, the data field sometimes returned the wrong values (everything is the second value) and sometimes it just didn't work. (See issue #1061)

This is an issue I ran into while trying to simplify ByteTrack.update_with_detections(). Boolean indexing is a feature that is commonly used in general and it would be good to ensure Detections works the same way np.arrays and lists work.

## Type of change

I tested this with the same test cases from my bug report. [Colab notebook](https://colab.research.google.com/drive/1nRKhgTnlubPNOtEGZV4QMYffDjI1VxnQ?usp=sharing)

## Docs

### DIFF
--- a/supervision/detection/utils.py
+++ b/supervision/detection/utils.py
@@ -849,7 +849,9 @@ def get_data_item(
                 subset_data[key] = [value[i] for i in index]
             elif isinstance(index, np.ndarray):
                 if index.dtype == bool:
-                    subset_data[key] = [value[i] for i, index_value in enumerate(index) if index_value]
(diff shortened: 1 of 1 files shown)


SkalskiP (approved):

SkalskiP:
Hi @rolson24 👋🏻 Thanks a lot for identifying the bug and preparing the fix. I've tested it; I'll add some unit tests as well.

PR #1067, opened 2024-03-28
fix: 🐞 ci trigger error fix

onuralpszr:


onuralpszr:
@SkalskiP  double quote added for prevent python 3.1 and make sure it does use "3.10"(ten) version.

SkalskiP (approved):

PR #1076, opened 2024-03-30
add minimum_consecutive_frames to bytetrack

rolson24:
# Description

This PR fixes https://github.com/roboflow/supervision/issues/1044. It replaces this PR https://github.com/roboflow/supervision/pull/1050 which caused only one track to be initialized per frame (see the comment)  I decided to re-implement this by instead of setting the tracker_id to -1, I used the existing is_activated property of STrack to mark whether the detection is valid or not. This does not affect the functionality of ByteTrack because every unconfirmed track will get added to the self.tracked_stracks list every time it is matched with a new detection.

## Type of change

I tested this by running writing pytests for ByteTrack and using them in a [colab notebook](https://colab.research.google.com/drive/1hM1trTUQfFO07znnI77wJPcziiNotCBK?usp=sharing). I will submit another PR with these tests added to supervision. They cover several cases, including all of the edge cases that I have run into.

This might require updating the documentation for ByteTrack, which I am willing to do if this is approved.

## Docs

-   [ ] Docs updated?

### DIFF
--- a/supervision/tracker/byte_tracker/core.py
+++ b/supervision/tracker/byte_tracker/core.py
@@ -12,6 +12,7 @@ from supervision.utils.internal import deprecated_parameter
 
 class STrack(BaseTrack):
     shared_kalman = KalmanFilter()
+    _external_count = 0
 
     def __init__(self, tlwh, score, class_ids, minimum_consecutive_frames):
         # wait activate
@@ -23,6 +24,8 @@ class STrack(BaseTrack):
         self.score = score
         self.class_ids = class_ids
         self.tracklet_len = 0
+
+        self.external_track_id = -1
 
         self.minimum_consecutive_frames = minimum_consecutive_frames
 
@@ -55,7 +58,7 @@ class STrack(BaseTrack):
     def activate(self, kalman_filter, frame_id):
         """Start a new tracklet"""
         self.kalman_filter = kalman_filter
-        self.track_id = self.next_id()
+        self.internal_track_id = self.next_id()
         self.mean, self.covariance = self.kalman_filter.initiate(
             self.tlwh_to_xyah(self._tlwh)
         )
@@ -76,7 +79,7 @@ class STrack(BaseTrack):
 
         self.frame_id = frame_id
         if new_id:
-            self.track_id = self.next_id()
+            self.internal_track_id = self.next_id()
         self.score = new_track.score
 
     def update(self, new_track, frame_id):
@@ -95,8 +98,9 @@ class STrack(BaseTrack):
             self.mean, self.covariance, self.tlwh_to_xyah(new_tlwh)
         )
         self.state = TrackState.Tracked
-        if self.tracklet_len >= self.minimum_consecutive_frames:
+        if self.tracklet_len == self.minimum_consecutive_frames:
             self.is_activated = True
+            self.external_track_id = self.next_external_id()
 
         self.score = new_track.score
 
@@ -133,6 +137,15 @@ class STrack(BaseTrack):
 
     def to_xyah(self):
         return self.tlwh_to_xyah(self.tlwh)
+    
+    @staticmethod
+    def next_external_id():
+        STrack._external_count += 1
+        return STrack._external_count
+    
+    @staticmethod
+    def reset_external_counter():
+        STrack._external_count = 0
 
     @staticmethod
     def tlbr_to_tlwh(tlbr):
@@ -147,7 +160,7 @@ class STrack(BaseTrack):
         return ret
 
     def __repr__(self):
-        return "OT_{}_({}-{})".format(self.track_id, self.start_frame, self.end_frame)
+        return "OT_{}_({}-{})".format(self.internal_track_id, self.start_frame, self.end_frame)
 
 
 def detections2boxes(detections: Detections) -> np.ndarray:
@@ -294,7 +307,7 @@ class ByteTrack:
             matches, _, _ = matching.linear_assignment(iou_costs, 0.5)
             detections.tracker_id = np.full(len(detections), -1, dtype=int)
             for i_detection, i_track in matches:
-                detections.tracker_id[i_detection] = int(tracks[i_track].track_id)
+                detections.tracker_id[i_detection] = int(tracks[i_track].external_track_id)
 
             return detections[detections.tracker_id != -1]
 
@@ -318,6 +331,7 @@ class ByteTrack:
         self.lost_tracks: List[STrack] = []
         self.removed_tracks: List[STrack] = []
         BaseTrack.reset_counter()
+        STrack.reset_external_counter()
 
     def update_with_tensors(self, tensors: np.ndarray) -> List[STrack]:
         """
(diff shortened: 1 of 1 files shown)


supervision/tracker/byte_tracker/core.py
```diff
@@ -290,6 +299,7 @@ def callback(frame: np.ndarray, index: int) -> np.ndarray:
             return detections[detections.tracker_id != -1]
 
         else:
+            detections = Detections.empty()
```
LinasKo, line 302: Hi @rolson24 :wave:
 
These changes look good, but I wanted to ask about this line.

Can you bring me up-to-speed for why resetting the detections to `empty` is needed here?

rolson24: Hi @LinasKo,

I made a PR last week that introduced a small bug where if the update_with_tensors function doesn't return any valid tracks and we set the tracker_id array to an empty array, then we run into issues with the annotators because they expect the tracker_id array to be the same length as the other arrays in the Detections object. Also we wouldn't want to return the detections that aren't tracked anyways, so the easiest solution I found is to just return an empty Detections object with an empty tracker_id array.

LinasKo: Thanks, that makes sense. I bumped into the unequally-sized array issue myself a few weeks back.

I need to sleep on this a bit, but I believe we do need the untracked detections to come out the other side unharmed.

The tracker tutorials look as follows:
```python
model = YOLO("yolov8n.pt")
tracker = sv.ByteTrack()
box_annotator = sv.BoundingBoxAnnotator()

def callback(frame: np.ndarray, _: int) -> np.ndarray:
    results = model(frame)[0]
    detections = sv.Detections.from_ultralytics(results)

return box_annotator.annotate(frame.copy(), detections=detections) 
```

Suppose there are many similar objects in the scene, lots of movement, or things tend to appear for brief moments only - detectable, but hard to track. Would be a shame if that produced no detections (and no annotations - user sees nothing happening).

I think it's a matter of initializing the tracker_id array. Do you think it's possible? Or would we need a bunch of None values that we can't cram in there?

(Note to self: possible setup of #943 error, possible trigger in update_with_tensors)

rolson24: Interesting. I guess we could change the tracker_id array to also store None values, and put None values in the tracker_id array. Alternatively, we could put -1 for the tracker_id (can only do this after update_with_tensors) and then change the `put()` method in the Trace class to only add a detection to the existing traces if the tracker_id is valid (non-negative).

LinasKo: This is a tricky one to handle. But you're right - it is better to let the tracker eat the detections instead of crashing the annotator. Well done for spotting that!

I'll open an issue about `tracker_id`. When it's used consistently throughout the codebase, then we might revisit this part.

Happy to merge.

LinasKo (approved):

supervision/tracker/byte_tracker/core.py
```diff
@@ -60,7 +62,7 @@ def activate(self, kalman_filter, frame_id):
 
         self.tracklet_len = 0
         self.state = TrackState.Tracked
-        if frame_id == 1:
+        if frame_id == 1 and self.minimum_consecutive_frames == 1:
```
SkalskiP, line 65: We could solve the problem of cycling through tracker IDs too quickly if we actually assigned a tracker_id only when that condition is met.

SkalskiP (changes requested):

supervision/tracker/byte_tracker/core.py
rolson24: Hmm. This condition only handles the first frame of the video. I considered only assigning a `tracker_id` once the condition on line 98 was met, (track has existed for longer than `minimum_consecutive_frames`), but that would break the rest of the tracker because it relies on tracker_id to identify different tracks in joint_stracks and sub_stracks, so tracker_id must have a unique value as soon as it gets activated.

One possible solution is to create an additional variable in the STrack object called something like `internal_tracker_id` that is used for all of the ByteTrack internal logic, and then leave the original `tracker_id` variable to only get assigned when a track has existed for long enough. This would allow the tracker to function as usual, and would make the output detections have reasonable tracker_ids.

I also considered just using the `tracker_id` variable and when a track is first initialized assigning the `tracker_id` to a temporary value with a new function just like `next_id` called `next_temporary_id()`. Then once a track has existed for `minimum_consecutive_frames`, the tracker_id would be reassigned to a valid track_id with `next_id()`. But I believe this solution would be problematic because there is a high likelihood that the `next_temporary_id()` and `next_id()` values would overlap, causing two tracks to share the same tracker_id. This could be solved by making the temporary_id's negative, but I don't know if we want to do that.

The final solution I have thought of is redesigning sub_stracks and joint_stracks to use iou overlap like remove_duplicate_tracks does instead of relying on tracker_id to identify tracks, which would allow more freedom to use tracker_id in other ways. (this would slow down the tracker a bit because of the linear assignment problem)

SkalskiP: Hi @rolson24 👋🏻 

I'm very sorry for the lack of response. I somehow missed the email notification that you had replied.

In that case, I am leaning towards a solution that introduces an internal and external tracker ID. The internal one can have arbitrary values, while the external one should have consecutive tracker IDs.

rolson24: Ok I have implemented this and it seems to be working. 
Here you can see the end of the video with `minimum_consecutive_frames = 1`. The tracker ids have gotten to quite a high number for a 13 second video (463!)
add_min_frames_test_video_1_frame

And here is the video for `minimum_consecutive_frames = 1`:

https://github.com/roboflow/supervision/assets/99894460/a2a32b09-c069-4b11-9767-094626067ac2


And here you can see the same area of the video with `minimum_consecutive_frames = 10`. The tracker ids have now only reached 110, and the person in the bottom middle is not tracked because the track hasn't existed for long enough.
add_min_frames_test_video_15_frames

Here is the video for `minimum_consecutive_frames = 10`:

https://github.com/roboflow/supervision/assets/99894460/c2f24125-bbbf-437b-bb4e-d9efe6fb7054

SkalskiP: Hi @rolson24 👋🏻 `#463` seems high. It is pretty much the same result we got here: https://github.com/roboflow/supervision/pull/1076#issuecomment-2039904362

rolson24:
Yes and I think it should be the same result because the tracker tracked
463 different people with minimum_consecutive_frames = 1. I will double
check with the current release version of supervision, but I believe that
is the expected behavior for minimum_consecutive_frames = 1.

supervision/tracker/byte_tracker/core.py
rolson24: Hmm i just checked with the release version and you seem to be correct. I will take a look.

rolson24:
Ok I believe I have fixed the issue. The problem was that each track was being re-assigned a new id when it was found after being lost. There also was an issue with the default case of `minimum_consective_frames = 1` producing slightly fewer tracks than the current release version because it would only return tracks that had already existed for 1 frame before it got assigned an external id. I fixed this so now it produces the exact same output as the release version.
This is the end of the test video from the **release** version of ByteTrack. You can see that the track_ids are lower than what I originally had:
add_min_frames_test_video_release

And this is the **new version** with `minimum_consecutive_frames = 1`. You can see that the track_ids are exactly the same:
add_min_frames_test_video_1_frame_final

And finally here is the **new version** with `minimum_consecutive_frames = 10`. You can see that the track_ids are much lower because the short tracks have been filtered out. The track_ids are also lower than the previous version due to fixing the re-assigning of track_ids every time a track is found after being lost. Additionally, the bottom middle person is not tracked yet because the track has not reached 10 frames long.

add_min_frames_test_video_15_frames_final

rolson24:
Just wanted to bump this. I think I fixed the issues we were running into.

SkalskiP (approved):

SkalskiP:
Hi @rolson24 👋🏻 I just tested the code, and it looks AWESOME! Thanks a lot for your contribution. 🙏🏻 Unfortunately, we already closed `supervision-0.20.0` release, so this feature will need to wait till May/June to get released.

rolson24:
Good to hear!

PR #1078, opened 2024-03-31
ci: 👷 supervision documentation 📚 release CI added

onuralpszr:


SkalskiP:
Hi @onuralpszr 👋🏻 What's happening here? :)

onuralpszr:
@SkalskiP it is your documentation release CI but, for the moment it is only a manual trigger because based on Github actions, I couldn't find a clean way of running only "when it is merging to main with tag or when I push tag only to main"  

I can suggest a manual run or run only when we push to the main branch. I assume the tag is already existed before merged

SkalskiP:
@onuralpszr, so this will build a new version of docs when we release a new version of `supervision`?

onuralpszr:
> @onuralpszr, so this will build a new version of docs when we release a new version of `supervision`?

Yes , but we have to trigger manually via GitHub action menu. Automatic ways I was thinking of as I mentioned, we can do when push to "main" If you like.

SkalskiP (approved):

Issue #1085, opened 2024-04-02
Bug during `as_coco()` conversion for segmentation (+ solution)


Youho99:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar bug report.

### Bug

Ce problème fait suite au travail fait sur mon précédent issue #1052 

`results.as_coco(annotations_path="test_coco/", min_image_area_percentage=0.01,)`
>IndexError: list index out of range

The problem is due to the fact that in certain cases, the mask returned by GroundedSam is completely inconsistent (less than 3 pixels for example). In this specific case, the function approximate_mask_with_polygons() (as_coco() > detections_to_coco_annotations() > approximate_mask_with_polygons()) returns an empty array.

So, the line 
```py
list(approximate_mask_with_polygons(
             mask=mask,
             min_image_area_percentage=min_image_area_percentage,
             max_image_area_percentage=max_image_area_percentage,
             approximation_percentage=approximation_percentage,
         )[0].flatten()
``` 
causes my error, because there is no item '0' in an empty list.

I will propose a fix following the writing of this issue.

I changed the architecture of the as_coco() function to be the same as the as_yolo() and as_pascal_voc() architecture.

### Environment

Supervision 0.19.0

### Minimal Reproducible Example

```py
from autodistill_grounded_sam import GroundedSAM
from autodistill.detection import CaptionOntology
from pathlib import Path
import supervision as sv
import os
import shutil
```
```py
# Put the cat image on your input directory
# Put your input directory path
input_dir = "/home/ggiret/Téléchargements/chat"
output_dir = "test/"
```
```py
if os.path.exists(output_dir):
    shutil.rmtree(output_dir)

classes = {"cat": "a cat",}

base_model = GroundedSAM(
    ontology=CaptionOntology(
        classes
    ),
    box_threshold = 0.005
)
```
```py
results = base_model.label(
        input_folder=input_dir,
        extension=".png",
        output_folder=output_dir, 
        record_confidence=True)
```
```py
results.as_coco(annotations_path="test_coco/", min_image_area_percentage=0.01,)
```

Image to use : 
![téléchargement (4)](https://github.com/roboflow/supervision/assets/44434482/9a417299-2d41-47ac-8378-61762dc3108e)

### Additional

In the code that I am going to propose, as_coco() will record either the BoundingBoxes coordinates or the Segmentation coordinates, but not both at the same time.

### Are you willing to submit a PR?

[...]

Issue #574, opened 2023-11-04
From binary masks with holes to COCO JSON format


MihaiDavid05:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Description

The code does not seem to support masks with holes, feature that is quite important for instance segmentation for example. By using the RLE format from COCO this would be attainable.

### Use case

It will help improve the binary masks exportation to COCO JSON format, by relying on RLE encoding when needed, otherwise keeping the polygon format. 

For example, if we take the binary mask of a specific class/category, from an image mask, we can find all possible contours in the class-mask (using RETR_TREE in openCV). If there is only 1 contour in the mask that is for sure 1 full polygon, therefore, we keep the polygon encoding. Otherwise, it means that either there are multiple instances in the mask, either there are some holes in some of the instances. In this case we can go from binary to uncompressed RLE, which will encode the possible holes in some instances. (for finding the area and bbox for this cases, it is needed to go from uncompressed RLE to compressed RLE first). 

### Additional

I built such an exporter [here](https://github.com/MihaiDavid05/Mask-to-COCO-JSON) based on other sources, and I could contribute with a PR, if needed.

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

ryouchinsa:
We updated our [general_json2yolo.py](https://github.com/ryouchinsa/Rectlabel-support/blob/master/general_json2yolo.py) script so that the RLE mask with holes can be converted to the YOLO segmentation format correctly.
This script originates from Ultralytics [JSON2YOLO](https://github.com/ultralytics/JSON2YOLO) repository.
We hope this script would help your business.

The RLE mask.

![スクリーンショット 2023-11-22 1 57 52](https://github.com/ryouchinsa/Rectlabel-support/assets/1954306/df2dfad5-a13f-49c6-818a-8aba1d3ce83c)

The converted YOLO segmentation format.

![スクリーンショット 2023-11-22 2 11 14](https://github.com/ryouchinsa/Rectlabel-support/assets/1954306/cc34027a-075f-4b67-8de7-bd2dddddac59)

On [RectLabel](https://rectlabel.com), you have 2 options to label the polygon with holes.

Label the polygon with holes using Segment Anything models.

![hole_sam](https://github.com/ryouchinsa/Rectlabel-support/assets/1954306/8fc6646d-ca00-4fba-89b0-33f1872fd957)

[...]

Issue #373, opened 2023-09-21
Is there a way to save a Dataset natively, without polygon conversions?


thomasf1:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Question

Is there a way to save a Dataset natively, without polygon conversions? Would be great If there would be a way to save the dataset with the masks as a zip, all save formats (coco, yolo, voc) do some processing. Ideally there would be a way to natively save a dataset (maybe as a ZIP file).

My usecase is experimenting to get the polygon conversion when saving as yolo after a run just right and writing some custom code for it, so it would be helpful to have the "raw" state with the masks saved somewhere.

### Additional

I might just have overlooked some obvious way of saving the dataset... Otherwise it is a feature request I guess.

SkalskiP:
Hi, @thomasf1 👋🏻 Sounds interesting. Have you perhaps already thought about the structure of such a dataset? Would each mask be stored as a separate black-and-white photo? How would you store information about the class to which an object with a given mask belongs?

thomasf1:
Well, there is already a solution for coco that´s somehow not being used roboflow:  RLE (_frString)

Also, not supporting masks causes other issues: When converting to Polygons, cutouts in masks seem often to be converted as a separate polygon.

Example: A person having this arms on his hips. The triangle shape inside the arms isn´t part of the person, but surrounded by it. The way supervisions translates that into polygons is as follows:
-  The outer shape of the person as a person polygon (reasonable, expected behaviour)
-  The inner cutout of background as another person polygon (blatantly wrong and counterproductive)

thomasf1:
@SkalskiP  There is some code here that should get you far: https://github.com/waspinator/pycococreator/blob/master/pycococreatortools/pycococreatortools.py

Also, in binary_mask_to_polygon, they seem to add some padding to avoid a problem supervision (and the roboflow app smart polyon) has: Creating masks that reach to the image edges.  Unfortunately, most polygons leave a 1+ px gap from the edge of the image when the masked subject goes beyond the edge of the image.

thomasf1:
In an Ideal world, both roboflow.com and supervision would use both masks and polygons and decide intelligently which one to use.

For instance segmentation:
Masks for:
- Objects with cutouts
- Objects that are overlapped by something else, cutting the mask into separate areas that are not connected (which in polygons results in several polygons, loosing the information that they are part of one Object)
- Maybe very small Objects

[...]

PR #1086, opened 2024-04-02
Fix `detections_to_coco_annotations` function for empty polygons.

Youho99:
# Description

Fix issue #1085 

Allows you not to cause an error if a polygon is empty

Modify the fields of the json file:
- the order of fields is affected
   - the bbox, area, or segmentation field including now the last fields of the annotation block
- from now on, there will be either the bbox and area fields (when working with bboxes), or the segmentation field (when working with segmentation)

## Type of change

Follow `Minimal Reproducible Example` of issue #1085 

Json fields modified (see comment in description)

## Docs

### DIFF
--- a/tests/dataset/formats/test_coco.py
+++ b/tests/dataset/formats/test_coco.py
@@ -916,8 +916,7 @@ def test_detections_to_coco_annotations(
         assert result == expected_result
 
 
-def test_detections_to_coco_annotations_handles_empty_approximated_polygons(
-) -> None:
+def test_detections_to_coco_annotations_handles_empty_approximated_polygons() -> None:
     detections = Detections(
         xyxy=np.array([[0, 0, 4, 4]], dtype=np.float32),
         class_id=np.array([0], dtype=int),
@@ -935,9 +934,7 @@ def test_detections_to_coco_annotations_handles_empty_approximated_polygons(
         ),
     )
 
-    with pytest.warns(
-        Warning, match="mask approximation returned no polygons"
-    ):
+    with pytest.warns(Warning, match="mask approximation returned no polygons"):
         annotations, _ = detections_to_coco_annotations(
             detections=detections,
             image_id=0,


SkalskiP:
Hi @Youho99! Thanks a lot for the time you spend creating this PR. Could you explain what's the problem with having `bbox`, `area` and `segmentation` simultaneously?

Youho99:
> Hi @Youho99! Thanks a lot for the time you spend creating this PR. Could you explain what's the problem with having `bbox`, `area` and `segmentation` simultaneously?

I don't know if it's really a problem to have both `bbox`, `area`, and `segmentation`.

Just, in this case, there are things to modify:
* Be careful not to use the bbox generated by GroundingDino (the one currently used), as it will not be the segmentation bbox. We therefore need to create a polygon_to_xyxy function for coco.
* For the rest, I think it will be automatically managed by my code


For information, I dissociated `bbox`, `area` from `segmentation` for 2 reasons:
* In yolo and pascal_voc formats, we do not have bbox information when using segmentation. In the idea where I replicated how these formats work, I also replicated this.
* During my tests, I exported the annotations to CVAT to visualize them. When I had an empty segmentation, the annotation that was displayed was that of the bbox (GroundingDino). But this scenario should not occur with my code.


Would you like me to also do the `polygon_to_xyxy` function for coco?

PS: the result of the `polygon_to_xyxy` function, for the pascal_vo and yolo formats, is never used for the polygon scenario. In fact, we should just make the xyxy argument optional in the `object_to_pascal_voc` and `object_to_yolo` functions. But that's not part of this problem, it's just something I noticed while I was studying the code ;).

Youho99:
Small modification:

I left the "segmentation" field empty when the technology is detection (bbox), because otherwise, we could not import into CVAT (so I suppose that the "segmentation" field is obligatory in the coco format)

SkalskiP:
@LinasKo could you take a look at this next week?

LinasKo:
Sure.

I've already verified the error happens and is subsequently fixed by this PR, but haven't looked at all into coco's format specification.

LinasKo:
Hi @Youho99 :wave:

I'll be reviewing this tomorrow. So far, I have one question: should `object_to_coco` remove the `bbox` when returning a segmentation? By my intuition, if it's available, should we not include it in the result?

Can you also comment a bit about the `bbox` generated by GroundingDino versus the one we have? Does it mean something else other than the detected object boxes?

Youho99:
Hi @LinasKo !

What do I think of it :
In the case of segmentation, the bbox will necessarily be available. The code provided in my PR filters empty or too small segmentations (I followed the filter rules of the base code, so no worries about the filtering logic).
So in theory it does not pose a problem to leave the additional bbox information in the coco json. Afterwards, it is not impossible that in certain specific use cases, only segmentation data is needed without bbox data (I don't see any cases like this, I'm just hypothesizing).

I would therefore be in favor of leaving the bbox information (and the area, it goes together) in addition to the segmentation information.

Note: In my current PR, I decided to follow the same logic as the yolo and pascal_voc formats, which are formats that do not save bbox information.

For the question of the difference in the calculation of the bbox:
autodistill_grounded_sam works like this:
1/ it invokes autodistill_grounding_dino, which generates a bbox.
2/ autodistill_grounded_sam retrieves the bbox, and segments the object in the bbox (with bbox prompt mode)
3/ the autodistill_grounding_dino bbox information and the autodistill_grounded_sam segmentation information are saved in the json

So here's the problem:
The autodistill_grounding_dino bbox is not the segmentation bbox, but the bbox used to generate the segmentation.
It is thus necessary to recalculate the bbox of the segmentation (taking the min and max x and y of the segmentation), as is already done for yolo and pascal_voc via the `polygon_to_xyxy` function (because these formats do not record the information of bbox of segmentation)

I made a diagram to illustrate my points, admire my artistic talents!

image

LinasKo:
That's a very thorough explanation :)
(A truly majestic depiction :artist:)

Let's add back the `bbox` in coco segmentation, even if it's inconsistent with other `from_X` methods. Who knows; maybe someone out there is relying on it.

I'd be keen to fix the Grounded SAM bbox issue too. Am I right that if we used Grounded SAM, e.g. via `from_sam` in `supervision/detections/core.py`, we're picking the wrong boxes for our `xyxy`?

Youho99:
Yes, I agree to add it too
If a person tells us that for their use, they necessarily have to remove the bbox, then perhaps we will have to add a parameter to a function to filter, depending on the user's wishes.
But the bbox must be that of segmentation (and not that of grounding dino ;D)

For the second part of the message:
I have the impression that you are indeed using the wrong bbox. However, I haven't checked it there. I know that the bbox contained in the Detections object is the grounding dino bbox, so it should not be called (and that seems to be the case here)

I made a code (for another use) allowing to recalculate the SAM bboxes precisely starting from a larger bbox (this simulates our use case). I'll see about making it open-source (I'll have the info in a few hours) if that helps.
But there is nothing very complex, you just need to read the mask, retrieve the coordinates of the extrema (in length and width) to calculate the bbox

SkalskiP:
> Be careful not to use the bbox generated by GroundingDino (the one currently used), as it will not be the segmentation bbox. We therefore need to create a polygon_to_xyxy function for coco.

@Youho99 Nowhere in the Supervision repository do we directly depend on GroundingDINO results. Could you point out where you saw this?

> I have the impression that you are indeed using the wrong bbox. However, I haven't checked it there. I know that the bbox contained in the Detections object is the grounding dino bbox, so it should not be called (and that seems to be the case here)

Once again, I'm not sure why we are talking about GroundingDINO here. :)

supervision/dataset/formats/coco.py
```diff
@@ -97,6 +97,36 @@ def coco_annotations_to_detections(
     return Detections(xyxy=xyxy, class_id=np.asarray(class_ids, dtype=int))
 
 
+def object_to_coco(
```
SkalskiP, line 100: Please always include `box`, `area`, and `segmentation` in the result COCO JSON.

Youho99:
There would therefore be 2 PRs:
A first allowing not to cause an error if a polygon is empty (on Supervision)

A second allowing the bboxes of the segmentation masks to be recorded in the Detections object generated by autodistill_grounded_sam (on autodistill_grounded_sam)

supervision/dataset/formats/coco.py
```diff
@@ -105,31 +135,36 @@ def detections_to_coco_annotations(
...
-            "segmentation": [polygon] if polygon else [],
-            "iscrowd": 0,
-        }
-        coco_annotations.append(coco_annotation)
-        annotation_id += 1
-    return coco_annotations, annotation_id
+            for polygon in polygons:
+                if polygon.any():  # polygon not empty
+                    next_object = object_to_coco(
+                        xyxy=xyxy,
```
SkalskiP, line 150: That's not the correct box; you save each polygon as a separate entry but still use the original box. We need to calculate a new box for each polygon. I recommend using the `polygon_to_xyxy` function.

IMG_6F79B4D94BBA-1

SkalskiP (changes requested):

supervision/dataset/formats/coco.py
SkalskiP: @Youho99 It looks like the COCO format supports disjointed masks - you can have multiple masks representing the same object. I think we should go for this.

https://github.com/roboflow/supervision/issues/373#issuecomment-1831781340

SkalskiP (changes requested):

supervision/dataset/formats/coco.py
Youho99: I agree with that.
I will make the changes.

supervision/dataset/formats/coco.py
Youho99: I agree in theory with this.

Now, doing the implementation the way I did it, I get exactly the same number of polygons as using as_yolo()

So if we define that the method of calculating polygons should change, then it would be better to also look at how this is done in as_yolo() and as_pascal_voc().

Youho99: The question is this:

Do we want the results of a conversion to coco, yolo and pascal voc to have exactly the same result (in the format I mean)?

Because it does not seem to me that it is possible to manage disjointed masks in yolo for example (except by applying a technique allowing the masks to be connected by thin strips)

Youho99: Example : https://github.com/roboflow/supervision/issues/574

Youho99: So I don't know exactly what to start with for this part.

I'm waiting for your feedback on this.

SkalskiP: Welcome back after the weekend, @Youho99. 👋🏻 For now, don't worry about the consistency of the `as_coco`, `as_yolo`, and `as_pascal_voc` methods. If necessary, we'll open a new task to address the differences.

Converting a disjointed mask into several separate detections is an undesirable result of the lack of support for such annotations in YOLO and PASCAL formats. I believe that if it's possible to avoid such conversions in the COCO format, we should do so.

SkalskiP (changes requested):

supervision/dataset/formats/coco.py
Youho99: Ok, I'll look that way then

Note that according to my tests on the yolo conversion, during mask disjoint, it only saves the first mask (for as_yolo())

Youho99:
@SkalskiP @LinasKo 
I have just made the necessary modifications to handle disjointed masks again.

While doing my tests, I noticed that I had an error when I tried to regenerate the DetectionDataset object with `from_coco`, when I activated the `force_masks` parameter.

I tried with supervision 0.19.0, and I get the error `ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part.`

I therefore deduce that I did not introduce this bug, but that it exists...

Youho99:
Hello !

I took the opportunity to correct the bug cited in my previous message as well.

Now, the `force_masks` parameter of the `from_coco()` function no longer raises an error. I made sure to also support disjoint masks.

Youho99:
Any news about this Pull Request? Could you review it? 
@LinasKo @SkalskiP

jschultz299:
Hi @Youho99, not sure if you are still working on this, but I am also dealing with inconsistent and often disjointed masks. I was able to cherry pick from your draft PR into the current develop branch, and converting detections with disjointed masks to COCO polygon annotations using `as_coco` worked well. 

Converting from COCO annotations using `from_coco()` and `force_masks=True`, however, I ran into an issue. Right now, it looks like you are taking every segmentation in the image and merging them together. Instead, I think we should perform this operation on each detection separately, then concatenate the masks together. That way, we only merge masks belonging to the same detection and preserve the number of total masks. Happy to discuss further or submit a PR if others are still interested in this feature.

Proposed solution ([line 85](https://github.com/roboflow/supervision/pull/1086/files#diff-ff4b973cbcba8720bedcddef6c75c8b98935c1e72c73c783ecbbe17ab9c45af4R85)):
```
if with_masks:
    masks = []
    for image_annotation in image_annotations:
        polygons = [
            np.reshape(np.asarray(segmentation, dtype=np.int32), (-1, 2))
            for segmentation in image_annotation["segmentation"]
        ]
        separate_masks = _polygons_to_masks(
            polygons=polygons, resolution_wh=resolution_wh
        )
        mask = (
            np.sum(
                separate_masks, axis=0, keepdims=True
            )  # merge mask parts representing a disjoint mask
            > 0  # ensure that the result is a binary mask
        )
        masks.append(mask)

    mask = np.concatenate(masks, axis=0)
```

LinasKo:
Hi @jschultz299 👋 

It's on us, that this PR is not reviewed yet. I'll have a look at a few hacktoberfest issues, and come back to this. It's high time I gave it the attention it deserves.

Youho99:
Since then, the library code has changed a lot (so I put the PR in Draft).

I haven't retested since, but I'm pretty sure that if the problem hasn't been solved via other PRs, then you have to start the solution from scratch and reimplement it with the current version of supervision.

I'm not working on the project where I needed this feature anymore, so it's a bit far for me right now.

LinasKo:
Thanks @Youho99. I'll take over and get it sorted.

Youho99:
@LinasKo any news about this ?

Issue #824, opened 2024-01-31
[KeyPoints] - create `KeyPoints` class and implement `from_ultralytics` method


SkalskiP:
### Description

- The `KeyPoints` class should be general. Flexible enough to support pose estimation, retina tracking, face landmarks, and other use cases. 
- The `KeyPoints` class should contain the following fields:
    - `xy` (`np.ndarray`): An array of shape `(N, K, 2)` containing the key points coordinates.
    - `confidence` (`Optional[np.ndarray`]): An array of shape `(N, K)` containing the confidence scores.
    - `data` (`Dict[str, Union[np.ndarray, List]]`): A dictionary containing additional data where each key is a string representing the data type, and the value is either a NumPy array or a list of corresponding data.
- The code should be placed in `supervision/keypoints/core.py`.
- Create a new KeyPoints section in the documentation.
- Ensure users know the code is still experimental and subject to change.

### API

```python
@dataclass
class KeyPoints:
    xy: np.ndarray
    confidence: Optional[np.ndarray] = None
    data: Dict[str, Union[np.ndarray, List]] = field(default_factory=dict)

    def __len__(self):
        return len(self.xy)

    def __eq__(self, other: KeyPoints):
        pass

    @classmethod
    def from_ultralytics(cls, ultralytics_results) -> KeyPoints:
        pass
```

### Additional

- Note: Please share a Google Colab with minimal code to test the new feature. We know it's additional work, but it will definitely speed up the review process. Each change must be tested by the reviewer. Setting up a local environment to do this is time-consuming. Please ensure that Google Colab can be accessed without any issues (make it public). Thank you! 🙏🏻

onuralpszr:
I like to work in this one @SkalskiP

SkalskiP:
@onuralpszr I like this idea a lot. Let's start with only the `KeyPoints` class. No annotators for now.

onuralpszr:
> @onuralpszr I like this idea a lot. Let's start with only the `KeyPoints` class. No annotators for now.

Okay lets do it.

adaiale:
I'm willing to help if you guys need an extra set of hands, I would love to have this implemented.

onuralpszr:
> I'm willing to help if you guys need an extra set of hands, I would love to have this implemented.

Thank you, I am sending this week(tomorrow) and add others too. I wanted that to be  done as well

SkalskiP:
[...]

PR #1105, opened 2024-04-09
feat: 🚀 initial sv.Keypoints features added

onuralpszr:
# Description

This pull request introduces the KeyPoints class. This class aims to provide a unified format for results from various keypoint detection models. First support for yolov8 pose added.

Fix: #824 

## Type of change

-   [X] This change requires a documentation update

YOUR_ANSWER

## Docs

-   [X] Docs updated? What were the changes: New section keypoints added

### DIFF
--- a/supervision/__init__.py
+++ b/supervision/__init__.py
@@ -69,6 +69,7 @@ from supervision.draw.utils import (
 )
 from supervision.geometry.core import Point, Position, Rect
 from supervision.geometry.utils import get_polygon_center
+from supervision.keypoints.core import KeyPoints
 from supervision.metrics.detection import ConfusionMatrix, MeanAveragePrecision
 from supervision.tracker.byte_tracker.core import ByteTrack
 from supervision.utils.file import list_files_with_extensions
--- a/supervision/detection/core.py
+++ b/supervision/detection/core.py
@@ -17,11 +17,11 @@ from supervision.detection.utils import (
     mask_to_xyxy,
     merge_data,
     process_roboflow_result,
-    validate_detections_fields,
     xywh_to_xyxy,
 )
 from supervision.geometry.core import Position
 from supervision.utils.internal import deprecated
+from supervision.validators import validate_detections_fields
 
 
 @dataclass
--- a/supervision/detection/utils.py
+++ b/supervision/detection/utils.py
@@ -1,5 +1,5 @@
 from itertools import chain
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 import cv2
 import numpy as np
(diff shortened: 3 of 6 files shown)


onuralpszr:
@SkalskiP before going further can you please check in current state. I would like to hear your opinion.

onuralpszr:
@LinasKo also please check :) (any suggestion would be nice)

LinasKo:
Hey @onuralpszr :wave:

That looks really good. I'm very fond of having validators, especially if they're nice and encapsulated like that. Plus, the structure is pretty much like `Detections` - new contributors would feel right at home. I've seen a spec floating around in one of GitHub issues, and it seems like this matches it exactly.

The issue is, in `supervision`, I keep bumping into problems coming from having this architecture of related multi-arrays. Synchronizing those is painful, and various bits of the codebase are just inconsistent with that. Sometimes easy to fix, sometimes not (especially if there's the `data` field).

By now, I'm gravitating towards an opinion that turning it into a list-of-structs would solve 99% of those - we just can't since it's painful for users to deprecate things. Here there'll be an extra hierarchy layer too - we'll have a list of keypoint-sets, for each person detected in an image, for example.

The most concise example I saw here: https://www.geeksforgeeks.org/numpy-structured-array/

It's nice since it seems to be memory-efficient, and every object is packed together with all variables it'd need. Iteration is easy, nullability wouldn't be a problem. A bit of thought is needed in how we define special values, but otherwise I feel like it'd be an optimal solution.

Stepping out of this PR (I'm not suggesting you change anything :slightly_smiling_face:), what are your thoughts? Do you think we'd run into any issues?

onuralpszr:
> Hey @onuralpszr 👋
> 
> That looks really good. I'm very fond of having validators, especially if they're nice and encapsulated like that. Plus, the structure is pretty much like `Detections` - new contributors would feel right at home. I've seen a spec floating around in one of GitHub issues, and it seems like this matches it exactly.
> 
> The issue is, in `supervision`, I keep bumping into problems coming from having this architecture of related multi-arrays. Synchronizing those is painful, and various bits of the codebase are just inconsistent with that. Sometimes easy to fix, sometimes not (especially if there's the `data` field).
> 
> By now, I'm gravitating towards an opinion that turning it into a list-of-structs would solve 99% of those - we just can't since it's painful for users to deprecate things. Here there'll be an extra hierarchy layer too - we'll have a list of keypoint-sets, for each person detected in an image, for example.
> 
> (Edit: or we need an 'object_id' array where values change for `person 1`, `person_2`, etc.)
> 
> The most concise example I saw here: https://www.geeksforgeeks.org/numpy-structured-array/
> 
> It's nice since it seems to be memory-efficient, and every object is packed together with all variables it'd need. Iteration is easy, nullability wouldn't be a problem. A bit of thought is needed in how we define special values, but otherwise I feel like it'd be an optimal solution.
> 
> Stepping out of this PR (I'm not suggesting you change anything 🙂), what are your thoughts? Do you think we'd run into any issues?

Thank you for your comment.

I will do more in-depth comments but one quick comment that comes to my mind is changing to structural needs change of internal gears and possibly without affecting others but If we change on the front that's also something I don't want right now (maybe we need to keep for old one for 10-20 releases maybe) I need to think bit more around it too.

LinasKo:
In this case, I only mean the new `Keypoints` module. `Detections` would remain as it currently is.

And yes, no rush! :)

onuralpszr:
> In this case, I only mean the new `Keypoints` module. `Detections` would remain as it currently is.
> 
> And yes, no rush! :)

I had that idea too, do only in Keypoints but still thinking :)

SkalskiP:
Hi @onuralpszr 👋🏻, I'll be taking a look at this PR tomorrow! 🔥

onuralpszr:
> Hi @onuralpszr 👋🏻, I'll be taking a look at this PR tomorrow! 🔥 

Sure, waiting your review.

SkalskiP:
Hi @onuralpszr 👋🏻, I'm closing this PR. @LinasKo picked up key points for work here: https://github.com/roboflow/supervision/pull/1128. Your commits are in the git history of this branch, so no contribution is lost. Thanks a lot!

PR #1108, opened 2024-04-10
Feature/batched inference slicer

LinasKo:
# Description

PR for inference slicer with batching. If `max_batch_size` is set, collections of slices are passed to the model.

Still using threads if `worker_threads>=2` (both batches and threads can be used simultaneously)

## Type of change

https://colab.research.google.com/drive/1j85QErM74VCSLADoGliM296q4GFUdnGM?usp=sharing

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2199,13 +2199,13 @@ pygments = ">2.12.0"
 
 [[package]]
 name = "mkdocs-material"
-version = "9.5.24"
+version = "9.5.25"
 description = "Documentation that simply works"
 optional = false
 python-versions = ">=3.8"
 files = [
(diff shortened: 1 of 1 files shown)


LinasKo:
Ready for review.

Tests in aforementioned Colab: https://colab.research.google.com/drive/1j85QErM74VCSLADoGliM296q4GFUdnGM?usp=sharing
It's safe to do 'Run All', though it will ask for Roboflow auth midway and HF token at the end.

The main outcome is the time printed underneath.
You may uncomment visualizers to plot the images.

Caveats:
* Inference examples run on the CPU - I'm speaking with Pawel to check what I missed
* The first Ultralytics run is slower due to memory allocation. Run the cell again and the `ultralytics threads=1 batch=1` case will run slightly faster.

LinasKo:
Ah, force-pushing a revert to `develop` closes the PR.

Makes sense, and is fine.

PR #1126, opened 2024-04-23
Add codec arg to process_video

LinasKo:
# Description

`VideoSink` provides an argument for a `FourCC` codec, allowing encoding the video in different formats. `process_video`, however, does not and defaults to `mp4v`.

This PR adds the same arg, hardcode-defaulting to `mp4v`.
Note: An alternative implementation would make `process_video` `codec` have the type `Optional[str]` and link it with `VideoSink`, but this is more concise and self-documenting.

## Type of change

https://colab.research.google.com/drive/16ggPDIUnk808HE0nUG-H7QCDjTncjldC?usp=sharing

:warning: Turns out `H264`, which I meant to enable, is not trivial to set up.

`cv2.VideoWriter_fourcc` fails silently, without throwing an error, doesn't even produce the file as an output. (tested in a [different Colab](https://colab.research.google.com/drive/1rso8hy0dZkz-K-_z0LjG62lVR5LeQlWl?usp=sharing))

Similar issues plague others - `XVID`, `YVYU`, etc. 

Leaving this as draft until we figure out what the codec behaviour is.

## Docs

### DIFF
--- a/supervision/utils/video.py
+++ b/supervision/utils/video.py
@@ -177,6 +177,7 @@ def process_video(
     source_path: str,
     target_path: str,
     callback: Callable[[np.ndarray, int], np.ndarray],
+    codec: str = "mp4v",
 ) -> None:
     """
     Process a video file by applying a callback function on each frame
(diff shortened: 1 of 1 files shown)


LinasKo:
I am closing for now. We'd like greater certainty surrounding the codecs first.

PR #1138, opened 2024-04-25
Implemented `from_yolo_nas` for keypoints

LinasKo:
# Description

Implementation of `from_yolo_nas` for KeyPoints.

:warning: Missing features:
- Does not include batch image input (I don't think we want that right now).
- `class_id` array entries are hardcoded to `0`; `class_name` entries as `"person"`. Results object from NAS has a `class_names`, but it is `None`. This seems to be set during training, and used in other model types. Let's revisit this after we decide how exactly we want to structure `class_id` and `class_names` in `KeyPoints`.

## Type of change

Prerequisites:
```bash
pip install super-gradients
```

Test code:
```python
import numpy as np
import cv2

import supervision as sv
from supervision.keypoint.skeletons import Skeleton
from supervision.assets import download_assets, VideoAssets

import super_gradients


# Available models: https://github.com/Deci-AI/super-gradients/blob/master/YOLONAS-POSE.md
yolo_nas = super_gradients.training.models.get(
    "yolo_nas_pose_s", pretrained_weights="coco_pose").to("cuda")

download_assets(VideoAssets.PEOPLE_WALKING)

# I found this important to tweak, regardless of model size
CONFIDENCE_THRESHOLD = 0.1

# cap = cv2.VideoCapture(0)
cap = cv2.VideoCapture(VideoAssets.PEOPLE_WALKING.value)
i = 0
while True:
    ret, frame = cap.read()
    if not ret:
        continue

    # TODO: I assume we're not supporting batch-image inputs yet. YOLO NAS does.

    result = yolo_nas.predict(frame, conf=CONFIDENCE_THRESHOLD)
    keypoints = sv.KeyPoints.from_yolo_nas(result)

    ann_point_large = sv.VertexAnnotator(color=sv.Color.ROBOFLOW, radius=5)
    ann_point_small = sv.VertexAnnotator(color=sv.Color.WHITE, radius=3)

    # Option 1: Use a predefined skeleton
    ann_skeleton = sv.EdgeAnnotator(
        color=sv.Color.ROBOFLOW,
        thickness=5,
        edges=Skeleton.YOLO_NAS.value
    )

    # Option 2: No skeleton
    # ann_skeleton = sv.EdgeAnnotator(
    #     color=sv.Color.ROBOFLOW,
    #     thickness=5,
    #     edges=[]
    # )

    # Option 3: Figure out automatically
    # ann_skeleton = sv.EdgeAnnotator(
    #     color=sv.Color.ROBOFLOW,
    #     thickness=5
    # )

    # Option 4: Take a guess (connect sequential points)
    # TODO: remove YOLO_NAS from Skeleton before running
    # ann_skeleton = sv.EdgeAnnotator(
    #     color=sv.Color.ROBOFLOW,
    #     thickness=5
    # )

    # Draw
    try:
        ann_skeleton.annotate(frame, keypoints)
        ann_point_large.annotate(frame, keypoints)
        ann_point_small.annotate(frame, keypoints)
    except Exception as e:
        print("Caught exception while annotating: \n", e)

    frame = cv2.resize(frame, (1024, 680))
    cv2.imshow('frame', frame)

    if cv2.waitKey(1) & 0xFF == ord('q'):
        break

cap.release()
cv2.destroyAllWindows()
```

## Docs


supervision/keypoint/core.py
```diff
@@ -136,6 +136,48 @@ def from_ultralytics(cls, ultralytics_results) -> KeyPoints:
...
+        Example:
+            ```python
+            import cv2
+            import supervision as sv
+            import super_gradients
+
+            image = cv2.imread(<SOURCE_IMAGE_PATH>)
+
+            yolo_nas = super_gradients.training.models.get(
+                "yolo_nas_pose_s", pretrained_weights="coco_pose").to("cuda")
```
SkalskiP, line 160: Using `to("cuda")` seems quite risky. It will crash if someone copies the snippet in an environment without a GPU. I would add `"cuda" if torch.cuda.is_available() else "cpu"` to the example. Just to be safe.

supervision/keypoint/core.py
```diff
@@ -136,6 +136,48 @@ def from_ultralytics(cls, ultralytics_results) -> KeyPoints:
...
+            results = yolo_nas.predict(image, conf=0.1)
+            keypoints = sv.KeyPoints.from_yolo_nas(results)
+            ```
+        """
+        if len(yolo_nas_results.prediction.poses) == 0:
+            return cls.empty()
+
+        xy = yolo_nas_results.prediction.poses[:, :, :2]
+        confidence = yolo_nas_results.prediction.poses[:, :, 2]
+        class_id = [0] * len(xy)
```
SkalskiP, line 171: We should not include any information not included in the result object. If `class_id` and `class_name` are not there, we should not add them. YOLO NAS allows you to train a custom pose estimation model: https://www.youtube.com/watch?v=J83ZvWfxjoA, so we should not assume we will get people or all classes are the same.

supervision/keypoint/core.py
```diff
@@ -136,6 +136,48 @@ def from_ultralytics(cls, ultralytics_results) -> KeyPoints:
...
+        return cls(
+            xy=xy,
+            confidence=confidence,
+            class_id=np.array(class_id),
+            data=data,
+        )
+
     def __getitem__(
         self, index: Union[int, slice, List[int], np.ndarray, str]
     ) -> Union["KeyPoints", List, np.ndarray, None]:
```
SkalskiP, line 183: Please fix typing from `"KeyPoints"` to `KeyPoints`.

supervision/keypoint/skeletons.py
```diff
@@ -25,6 +25,30 @@ class Skeleton(Enum):
         (17, 15),
     ]
 
+    # Hardcoding here, but it also comes within model results
+    # Note: the classes in the model are 0-indexed, while the skeleton is 1-indexed
+    YOLO_NAS = [
```
SkalskiP, line 30: Let's drop this skeleton. The only difference between this and the COCO skeleton we already have is that it has 2 more edges that connect the head to the torso. That's why we should aim to annotate YOLO NAS result in the same way as YOLOv8.

SkalskiP (changes requested):

supervision/keypoint/core.py
LinasKo: After a moderate attempt, I found no way to add `class_names` or labels when training a pose net for Yolo NAS. I've mocked the results by using a detections network.

LinasKo:
Ready for review:
https://colab.research.google.com/drive/1RGmU9lwXaffYWR9a1MegC310SWLA7Rw_?usp=sharing

I've removed hardcoding of `person`, but kept a special value of `-1` and name `""` when no label or class_name is given by the model. Finding no way to add class names when training a pose net, I've verified the structure of label / class_names by calling a detection model.

The special fields for -1 match what we have in `Detections` and conform yo our current validators + which assert that given we have (n, ...) `xy` points, we'll always have `(n, )` of `class_id`. If that's not something we no longer wish to support, let's make a separate PR.

LinasKo:
Done. Tested and `class_id` is now `None` when labels aren't given; `data` is `{}`.

Class names and IDs still appear when input has labels and names.

SkalskiP:
Awesome! I tested as well. Merging!

PR #1161, opened 2024-05-03
Non-max-merging for Inference Slicer

LinasKo:
# Description

This PR builds on top of @mario-dg PR #500, replacing it, fixing issues, adding non-max-merging to `Detections` and connecting to InferenceSlicer.

There main stylistic complexity here is dealing with individual objects inside `Detections`.

Here, the following changes are implemented:
* `Detections` now have `with_nmm`, which merges the detections based on their overlap.
* `detections/core.py` exposes an `OverlapFiltering` enum, which allows other methods to distinguish between the types. This enum is exposed globally in `__init__.py`.
* `InferenceSlicer` is the only class using it right now. `NMS` is performed by default, and users may specify `overlap_filtering` to pick which filtering type to use.
* New method `Detections._set_at_index` mimicks the former `__setitem__`, to be used internally. Only used in one location, so happy to refactor there.
* Removed `Optional` from InferenceSlicer's `iou_threshold` arg.

:warning: So far, only tested with object detections - not segmentation.
:warning: Naming could be made more clear, e.g. in methods like `utils.py::batch_non_max_merge`. Ping me and I'll look into that.

## Type of change

Tests:
* Works with 0 detections
* Works with ~5 detections
* Works with inference slicer, visual results the same as drawn in SAHI example in #500 
* Benchmarking in progress

Colab: https://colab.research.google.com/drive/1iiJ5oCl8dwwcefRgNXRdaFH6G9dVpf4P?usp=sharing

One thing I couldn't figure out was the **_right_** way to access values of individual detected objects.
When a scalar was needed, I went with `self.xyxy[i]`. When an array was expected, I left the `self[i].xyxy` calls. Feels a bit backwards, but most concise - is there a better way?

## Docs

### DIFF
--- a/supervision/__init__.py
+++ b/supervision/__init__.py
@@ -35,7 +35,7 @@ from supervision.dataset.core import (
     DetectionDataset,
 )
 from supervision.detection.annotate import BoxAnnotator
-from supervision.detection.core import Detections
+from supervision.detection.core import Detections, OverlapFiltering
 from supervision.detection.line_zone import LineZone, LineZoneAnnotator
 from supervision.detection.tools.csv_sink import CSVSink
 from supervision.detection.tools.inference_slicer import InferenceSlicer
--- a/supervision/detection/core.py
+++ b/supervision/detection/core.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from contextlib import suppress
 from dataclasses import dataclass, field
+from enum import Enum
 from typing import Any, Dict, Iterator, List, Optional, Tuple, Union
 
 import numpy as np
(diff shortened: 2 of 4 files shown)


LinasKo:
Colab: https://colab.research.google.com/drive/1iiJ5oCl8dwwcefRgNXRdaFH6G9dVpf4P?usp=sharing

LinasKo:
It fails when `Detections.merge` is called on segmentation detections. Frequently `mask=None` is returned, so detections in the Slicer can't be merged.

My guess is that it's caused by my old friend #943, but I have not verified.

LinasKo:
Closing; Moving fixes to https://github.com/roboflow/supervision/pull/500 as discussed

Issue #1114, opened 2024-04-12
[DetectionDataset] - extend `from_coco` and `as_coco` with support for masks in RLE format


LinasKo:
### Description

The COCO dataset format allows for the storage of segmentation masks in two ways:

- Polygon Masks: These masks use a series of vertices on an x-y plane to represent segmented object areas. The vertices are connected by straight lines to form polygons that approximate the shapes of objects.
Run-Length Encoding (RLE): RLE compresses segments of pixels into counts of consecutive pixels (runs). This method efficiently sequences pixels by reporting the number of pixels that are either foreground or background. For instance, starting from the top left of an image, the encoding might record '5 white pixels, 3 black pixels, 6 white pixels', and so on.

Supervision currently only supports Polygon Masks, but we want to expand support for masks in RLE format. To do this, you will need to make changes in [`coco_annotations_to_detections`](https://github.com/roboflow/supervision/blob/9d9acd7e587d117a2faa395580664aeb83be5efb/supervision/dataset/formats/coco.py#L72) and [`detections_to_coco_annotations`](https://github.com/roboflow/supervision/blob/9d9acd7e587d117a2faa395580664aeb83be5efb/supervision/dataset/formats/coco.py#L100).

### Links

- an official [explanation](https://github.com/cocodataset/cocoapi/issues/184) from the COCO dataset repository
- old supervision [issue](https://github.com/roboflow/supervision/issues/373) providing more context

### Additional

- Note: Please share a Google Colab with minimal code to test the new feature. We know it's additional work, but it will speed up the review process. The reviewer must test each change. Setting up a local environment to do this is time-consuming. Please ensure that Google Colab can be accessed without any issues (make it public). Thank you! 🙏🏻

LinasKo:
There's been an extended discussion in the past along with code examples of various use cases provided in issue https://github.com/roboflow/supervision/issues/373

David-rn:
Hi, great job with supervision! I would like to try contributing, but first I have some doubts regarding this issue. According to  [coco data format](https://cocodataset.org/#format-data) the RLE format is used only when `iscrowd=1` but I don't see this attribute in `Detections` class so, does this mean that the idea is to select one type of segmentation storage when loading or saving annotations? Thanks in advance!

SkalskiP:
[...]

PR #1163, opened 2024-05-05
1114 rle support for coco

emSko:
# Description
PR adds support for the RLE format in the COCO dataset. Based on #1114 .

It required implementing RLE encode/decode functions and extending the _coco_annotations_to_detections_/_detections_to_coco_annotations_ functions.  

## Type of change

All functionalities are covered with unit tests + [notebook](https://colab.research.google.com/drive/1xjnGh007xB-RbRAEqkaA4H2e-GtO_dd1?usp=sharing) checks the functionalities on larger images.

None

## Docs

-   [x] Docs updated? What were the changes: `docs/datasets/utils.md`


SkalskiP:
Hi @emSko 👋🏻 Could you accept CLA? Without this, we won't be able to merge your PR.

SkalskiP:
Let's split the work on this PR into three parts:

- [ ] Implementation of `rle_to_mask` and `mask_to_rle`. Please write valid docstrings in Google format for both methods. For example, take a look at other documented functions in our repository [here](https://github.com/roboflow/supervision/blob/447ef41fc45353130ec4dccdc7eeaf68b622fb7e/supervision/dataset/utils.py#L106). The documentation should describe the input and output arguments, as well as provide a simple usage example. Regarding the tests, create several cases: when the image is entirely zeros, when the image is entirely ones, when the image contains a mask with a hole, when the image contains a disjoined mask consisting of multiple elements. Use `pytest.mark.parametrize` as demonstrated in [this](https://github.com/roboflow/supervision/blob/447ef41fc45353130ec4dccdc7eeaf68b622fb7e/test/dataset/test_utils.py#L18) test.
   - [ ] code
   - [ ] documentation
   - [ ] unit tests 
- [ ] Integration of `rle_to_mask` method into `sv.DetectionDataset.from_coco` logic.
   - [ ] code
   - [ ] documentation updates
- [ ] Integration of `mask_to_rle` method into `sv.DetectionDataset.as_coco` logic.
   - [ ] code
   - [ ] documentation updates

supervision/dataset/utils.py
```diff
@@ -129,3 +130,23 @@ def train_test_split(
 
     split_index = int(len(data) * train_ratio)
     return data[:split_index], data[split_index:]
+
+
+def rle_to_mask(rle: np.ndarray, resolution_wh: Tuple[int, int]) -> np.ndarray:
```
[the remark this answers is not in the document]
SkalskiP: Please include a doc string. For example:
```
"""
Convert run-length encoding (RLE) to a binary mask.

Parameters:
    rle (np.ndarray): The run-length encoding array.
    resolution_wh (Tuple[int, int]): The width and height of the desired mask.

Returns:
    np.ndarray: The decoded binary mask.
"""
```

supervision/dataset/utils.py
```diff
@@ -129,3 +130,23 @@ def train_test_split(
...
+    width, height = resolution_wh
+
+    zero_one_values = np.zeros_like(rle)
+    zero_one_values[1::2] = 1
+
+    decoded_rle = np.repeat(zero_one_values, rle)
+    return decoded_rle.reshape((height, width), order="F")
+
+
+def mask_to_rle(binary_mask: np.ndarray) -> list:
```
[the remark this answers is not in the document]
SkalskiP: Please include a doc string. For example:
```
"""
Convert a binary mask to run-length encoding (RLE).

Parameters:
    binary_mask (np.ndarray): The binary mask.

Returns:
    List[int]: The run-length encoded mask.
"""
```

SkalskiP (changes requested):

supervision/dataset/utils.py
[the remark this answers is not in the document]
emSko: resolved with 18a790e9ea4160d6ab1c2f9a08f62e14ab458784

supervision/dataset/formats/coco.py
```diff
@@ -88,10 +109,25 @@ def coco_annotations_to_detections(
...
+            if image_annotation["iscrowd"]
         ]
-        mask = _polygons_to_masks(polygons=polygons, resolution_wh=resolution_wh)
+        mask_rle = _rles_to_masks(rles=rles, resolution_wh=resolution_wh)
+
         return Detections(
-            class_id=np.asarray(class_ids, dtype=int), xyxy=xyxy, mask=mask
+            class_id=np.asarray(class_ids, dtype=int),
+            xyxy=xyxy,
+            mask=_concatenate_annotation_masks(
```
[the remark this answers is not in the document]
emSko: You are right, I have changed the mask generation procedure here 3ade7f060a2db171f89494bc687c6255cca70f87.
Also now I have second test where RLE mask is first in annotation the dictionary.

emSko:
@SkalskiP what do you mean by the documentation? Only docstrings or are there additional files? Where to find them?

SkalskiP:
Hi @emSko 👋🏻 Thank you very much for dedicating time to this task. In the meantime, I did some light research and found these two videos on YouTube:
- https://www.youtube.com/watch?v=tScp754b8iU
- https://www.youtube.com/watch?v=2zCZXZOoQJQ

I compared the results of our implementation of encoding and decoding as well as the execution time. The outcomes are identical; however, the execution time, especially for encoding, is significantly slower. Could we utilize similar tricks to accelerate the encoding process? Here's the Colab I used.

https://colab.research.google.com/drive/1Yl3w0B12htaso_VxZ-S2LvvxftiRk-ta?usp=sharing

emSko:
Hi @SkalskiP  
**Performance:**
I have changed both `rle_to_mask` and `mask_to_rle` functions. 
I was able to develop even faster solution to encode the mask to RLE.
For decoding RLE to the mask, I applied tricks from your code.

You can find the comparison at the bottom of this [notebook](https://colab.research.google.com/drive/1xjnGh007xB-RbRAEqkaA4H2e-GtO_dd1?usp=sharing).

**Documentation:**
In your comment, you have pointed in your [comment](https://github.com/roboflow/supervision/pull/1163#issuecomment-2100366272) to `detection/utils.md`. I placed my functions in the **dataset package** (utils.py). I figured since Detections are operating on masks (images) and the RLE functionality is used only when reading/writing to a dataset then it would be better to place those functions in the dataset package.
Please let me know if you want me to:
a) move functions to the detection package and modify detection/utils.md.
b) modify detection/utils.md without changing existing code.
c) change a different documentation file. 

**Integration to as_coco:**
It is one of the last things to do. I have proposed 2 possible solutions here:  [comment](https://github.com/roboflow/supervision/issues/1114#issuecomment-2094889036) Can you say if any of them sounds ok, and if not what kind of solution do you have in mind?

SkalskiP:
Hi @emSko 👋🏻, I like your work with `rle_to_mask` and `mask_to_rle`. The code, tests, and docks look great. Thank you! 🙏🏻 I just made a few small changes (I didn't want to bother you with those small comments).

I also like the updates you made in `coco_annotations_to_detections`; now it's time to invest a bit of time into the `test_coco_annotations_to_detections` function.
- please update `mock_cock_coco_annotation` so that it will be able to take `Optional[rle]` argument. 
- use `mock_cock_coco_annotation` in your test cases
- please make your tests easy to follow; for example instead of this:
    
    ```python
    Detections(
        xyxy=np.array([[0, 0, 10, 10]], dtype=np.float32),
        class_id=np.array([0], dtype=int),
        mask=np.array(
            [
                0 if i >= 10 or j >= 10 or (i < 5 and j >= 5) else 1
                for i in range(0, 20)
                for j in range(0, 20)
            ]
        ).reshape((1, 20, 20)),
    ),
    ```

    Try using a smaller mask and define it in place:

    ```python
    array([[[1, 1, 1, 1, 1, 0, 0, 0, 0, 0],
            [1, 1, 1, 1, 1, 0, 0, 0, 0, 0],
            [1, 1, 1, 1, 1, 0, 0, 0, 0, 0],
            [1, 1, 1, 1, 1, 0, 0, 0, 0, 0],
            [1, 1, 1, 1, 1, 0, 0, 0, 0, 0],
            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]]])
    ```

SkalskiP:
Hi @emSko 👋🏻 I like the updates you made in `test_coco_annotations_to_detections`. Now is time to `detections_to_coco_annotations` updates. As I said, it would be awesome if we could deliver a smart solution that uses RLE if needed. Here's a result of quick ChatGPT conversation:

```python
import numpy as np
import cv2

def has_holes(mask):
    """
    Check if a 2D boolean mask has any holes (groups of False values completely surrounded by True values).
    
    Args:
        mask (np.ndarray): 2D boolean array.
    
    Returns:
        bool: True if the mask has holes, False otherwise.
    """
    # Convert boolean mask to uint8
    mask_uint8 = mask.astype(np.uint8)
    
    # Invert the mask
    inverted_mask = cv2.bitwise_not(mask_uint8)
    
    # Find contours
    contours, _ = cv2.findContours(inverted_mask, cv2.RETR_CCOMP, cv2.CHAIN_APPROX_SIMPLE)
    
    # Check for any internal contours, which indicate holes
    for i in range(len(contours)):
        if _[0][i][3] != -1:
            return True
    
    return False

# Example usage
mask = np.array([[True, True, True, True],
                 [True, False, False, True],
                 [True, True, True, True]], dtype=bool)
print(has_holes(mask))  # Output: True
```

```python
def has_multiple_unconnected_sections(mask):
    """
    Check if a 2D boolean mask has multiple unconnected sections of True values.
    
    Args:
        mask (np.ndarray): 2D boolean array.
    
    Returns:
        bool: True if the mask has multiple unconnected sections, False otherwise.
    """
    # Convert boolean mask to uint8
    mask_uint8 = mask.astype(np.uint8)
    
    # Find connected components
    num_labels, _ = cv2.connectedComponents(mask_uint8)
    
    # More than one label means multiple unconnected sections
    return num_labels > 2  # 0 is the background label, so we check for > 2

# Example usage
mask = np.array([[True, False, False, True],
                 [False, False, False, False],
                 [True, True, False, False]], dtype=bool)
print(has_multiple_unconnected_sections(mask))  # Output: True
```

If `has_holes` or `has_multiple_unconnected_sections` is `True` we should auto RLE.

LinasKo:
Hi @emSko :wave: 

Do a `git pull` - I've added a line that fixes the segfault.

I couldn't reproduce it as well, but found [this issue](https://github.com/opencv/opencv/issues/21783)

We'll eventually update the opencv to `>=4.6.0.0` which allegedly resolves it, but for now this should be sufficient.

emSko:
Hi @LinasKo, thank you for your help.

@SkalskiP I think the PR is ready for review. I don't think that `from_coco` requires any documentation updates.

supervision/dataset/formats/coco.py
```diff
@@ -83,20 +97,33 @@ def coco_annotations_to_detections(
...
+            image_annotations=image_annotations, resolution_wh=resolution_wh
+        )
         return Detections(
             class_id=np.asarray(class_ids, dtype=int), xyxy=xyxy, mask=mask
         )
 
     return Detections(xyxy=xyxy, class_id=np.asarray(class_ids, dtype=int))
 
 
+def _mask_has_holes(mask: np.ndarray) -> bool:
```
SkalskiP, line 110: Let's move the `_mask_has_holes` and `_mask_has_multiple_segments` functions to [`supervision/detection/utils.py`](https://github.com/roboflow/supervision/blob/develop/supervision/detection/utils.py) and make them publicly accessible. These features are valuable utilities for Supervision users and should be prominently featured.

Additionally, please:

- Add these functions to the main [`__init__.py`](https://github.com/roboflow/supervision/blob/develop/supervision/__init__.py) file to make them available in the public API.
- Include comprehensive docstrings for each function to explain their use and parameters.
- Document these functions in [`utils.md`](https://github.com/roboflow/supervision/blob/develop/docs/detection/utils.md) to ensure they are visible in our documentation.
- Add unit tests for both functions.

supervision/dataset/formats/coco.py
```diff
@@ -108,24 +135,37 @@ def detections_to_coco_annotations(
...
+                segmentation = [
+                    list(
+                        approximate_mask_with_polygons(
+                            mask=mask,
+                            min_image_area_percentage=min_image_area_percentage,
+                            max_image_area_percentage=max_image_area_percentage,
+                            approximation_percentage=approximation_percentage,
+                        )[0].flatten()
+                    )
+                ]  # multicomponent masks supported only for rle format
```
SkalskiP, line 160: We can remove the comment :)

test/dataset/formats/test_coco.py
```diff
@@ -226,6 +232,156 @@ def test_group_coco_annotations_by_image_id(
             ),
             DoesNotRaise(),
         ),  # two image annotations
+        (
+            [
+                mock_cock_coco_annotation(
```
SkalskiP, line 237: Could you please rename `mock_cock_coco_annotation` to `mock_coco_annotation`? :) I just noticed that some refactoring went wrong, and we ended up with this funny name.

SkalskiP (changes requested):

SkalskiP:
Hi @emSko 👋🏻 At the outset, I apologize for the late reply - I was traveling at the end of the week and was not on GitHub. @LinasKo, thanks for your support! 🙏🏻 

@emSko you did an excellent job here! We are already on the last straight.  I left some final comments.

test/dataset/formats/test_coco.py
emSko: resolved with a40bf78bb20dcca49ad19c75b808a51d913da9f4

supervision/dataset/formats/coco.py
emSko: resolved with a40bf78bb20dcca49ad19c75b808a51d913da9f4

supervision/dataset/formats/coco.py
emSko: resolved with commits from 2626263ff3f552634b301501b4824df160f2ee47 to 6dbf3f61e3e764c56c6152f18088f7624354d1dd

SkalskiP:
Hi @emSko 👋🏻 

I just tested your code in [Google Colab](https://colab.research.google.com/drive/11DZMZWutJAwXWro9Ui-rcf1pHYKIBDpv?usp=sharing). Everything seems to be working fine. I also made a few minor changes to the docs, including adding some visualizations. Everything looks great. Many thanks for this work. This is a massive upgrade for COCO dataset support. Working with you was a pleasure. Thank you! 🙏🏻

SkalskiP (approved):

Issue #1164, opened 2024-05-05
ByteTrack memory consumption increases indefinitely


tc360950:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar bug report.

### Bug

ByteTrack stores removed tracklets in a list, which is extended with new removed tracklets after every pass (line 489 in [https://github.com/roboflow/supervision/blob/develop/supervision/tracker/byte_tracker/core.py#L489](byte_tracker/core.py):
```
self.lost_tracks = sub_tracks(self.lost_tracks, self.removed_tracks)
self.removed_tracks.extend(removed_stracks)
```
As a result, size of `removed_tracks` collection grows unbounded and computational cost of 
`sub_tracks(self.lost_tracks, self.removed_tracks)` increases with every pass. 
The fix is simple - we only need to keep latest (i.e. the one from the last pass) removed tracklets in the collection - I've submitted a corresponding PR. 

        
        

### Environment

Not applicable

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

SkalskiP:
Hi, @tc360950! 👋🏻 I completely agree that the problem exists. Why do you think storing values from the last pass is enough?

tc360950:
Hi @SkalskiP, `removed_tracks` collection is only used on line 489:
```
self.lost_tracks = sub_tracks(self.lost_tracks, self.removed_tracks)
```
where it's used to get rid of lost tracks which have been marked for removal (i.e. those which have not been matched to a detection for a predefined time). Since tracks marked for the removal in the latest pass are added to `removed_tracks` collection on line 490:
```
self.removed_tracks.extend(removed_stracks)
```
lost tracks which were removed on pass no. N will actually be removed (i.e. moved to `removed_tracks` collection) after pass no. N  + 1 (which btw is another error, although without any dire consequences). This implies that we must at least do (which is what I propose):
``` 
self.removed_tracks = removed_stracks
```
Now, as to why this is enough. Once a track lands in `removed_tracks` it does not get out anywhere. I.e. once track is removed from lost_tracks on line 489, it will not reappear there. As a result we do not need to keep whole history, only most recently removed tracks are needed to "sanitize" `lost_tracks` collection. 

[...]

PR #1166, opened 2024-05-05
Fix ever increasing memory consumption in ByteTrack

tc360950:
# Description

Fix issue https://github.com/roboflow/supervision/issues/1164

ByteTrack stores all removed tracklets in a single list. This list grows indefinitely increasing tracking computational cost. 
This PR fixes the issue by restricting the list to tracklets removed in the last pass of the tracker.

## Type of change

Currently, there are no tests for tracking. I've used publicly available tracking videos to test for regression.

### DIFF
--- a/supervision/tracker/byte_tracker/core.py
+++ b/supervision/tracker/byte_tracker/core.py
@@ -487,7 +487,7 @@ class ByteTrack:
         self.lost_tracks = sub_tracks(self.lost_tracks, self.tracked_tracks)
         self.lost_tracks.extend(lost_stracks)
         self.lost_tracks = sub_tracks(self.lost_tracks, self.removed_tracks)
-        self.removed_tracks.extend(removed_stracks)
+        self.removed_tracks = removed_stracks
         self.tracked_tracks, self.lost_tracks = remove_duplicate_tracks(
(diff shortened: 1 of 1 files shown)


SkalskiP:
Hi @tc360950 👋🏻 I understand the memory-saving argument. But are you sure that such a change won't affect ByteTrack's behavior? @rolson24, what do you think about it?

rolson24:
Hi @tc360950, @SkalskiP

This won't affect ByteTrack's self-contained behavior, although if people are using `removed_tracks` to keep track of all of the objects that have been tracked, it might break their code. On the other hand, `removed_tracks` is only storing the final location of each removed track, so it's probably not being used by many people for that, and this change would still allow them to see which tracks were removed in the previous frame.

This change looks OK to me.

SkalskiP:
We have never promoted `removed_tracks` as an external API for Supervision users to utilize; therefore, I support this change, especially if it results in time savings and reduced resource consumption.

@rolson24 thanks a lot for confirmation. @tc360950 thanks a lot for your work. 🙏🏻 

Merging! ;)

SkalskiP (approved):

PR #1177, opened 2024-05-08
More flexibility in Detections.merge 

LinasKo:
# Description

* `Detections.merge` now merges detections when some of them are `None` or empty.
* Missing keys in `data` should not be an issue, unless further detections containing new data in those.
* Expanded tests with the new behaviour.

See the tests to see expected behaviour.

## Type of change

Please delete options that are not relevant.

-   [x] Bug fix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] This change requires a documentation update

## How has this change been tested, please provide a testcase or example of how you tested the change?

:notebook:  [Google Colab](https://colab.research.google.com/drive/1PQ3j_5hAbuh4ivHEsQV8sxKvIVDGK7j7?usp=sharing)
* (Slicer with segmentation model still fails, but with a different error)
* Roboflow model (with `data` is now merged with `Detections.empty` successfully).
* Multiple new tests were written - all pass.

## Any specific deployment considerations

None

## Docs

-   [ ] Docs updated? What were the changes:

PR #1178, opened 2024-05-08
Feat/inference slicer segmentation

LinasKo:
# Description

Prerequisites:
* #1177

This allows the slicer to merge segmentation masks.

Previously it would fail when attempting to stack slices of different sizes, as there'd be no attempts to move & pad the slices. The result is that each mask would be of size(slice), rather than size(image).

## Type of change

Same notebook as for prior PR.
:notebook: [Google Colab](https://colab.research.google.com/drive/1PQ3j_5hAbuh4ivHEsQV8sxKvIVDGK7j7#scrollTo=SCTstIHfAhD7)

1. Non-max suppression wrecks the visuals for large images. Best way to test would be to merge the PRs adding Non-max merging & slicer filter selector, and then test with `NONE`.
2. It takes a lot of RAM and time, even for a small model. High-RAM env in Colab is needed.

## Docs

### DIFF
--- a/supervision/detection/tools/inference_slicer.py
+++ b/supervision/detection/tools/inference_slicer.py
@@ -4,20 +4,29 @@ from typing import Callable, Optional, Tuple
 import numpy as np
 
 from supervision.detection.core import Detections
-from supervision.detection.utils import move_boxes
+from supervision.detection.utils import move_boxes, move_masks
 from supervision.utils.image import crop_image
 
 
-def move_detections(detections: Detections, offset: np.array) -> Detections:
+def move_detections(
+    detections: Detections, offset: np.ndarray, image_shape: np.ndarray
+) -> Detections:
     """
     Args:
         detections (sv.Detections): Detections object to be moved.
-        offset (np.array): An array of shape `(2,)` containing offset values in format
+        offset (np.ndarray): An array of shape `(2,)` containing offset values in format
             is `[dx, dy]`.
+        image_size (np.ndarray): An array of shape `(2,)` or `(3,)`, size of the image
+            in format is `[width, height]`.
     Returns:
         (sv.Detections) repositioned Detections object.
     """
     detections.xyxy = move_boxes(xyxy=detections.xyxy, offset=offset)
+    if detections.mask is not None:
+        shape_xy = image_shape[:2][::-1]
+        detections.mask = move_masks(
+            masks=detections.mask, offset=offset, desired_shape=shape_xy
+        )
     return detections
 
 
@@ -126,7 +135,9 @@ class InferenceSlicer:
         """
         image_slice = crop_image(image=image, xyxy=offset)
         detections = self.callback(image_slice)
-        detections = move_detections(detections=detections, offset=offset[:2])
+        detections = move_detections(
+            detections=detections, offset=offset[:2], image_shape=image.shape
+        )
 
         return detections
 
--- a/supervision/detection/utils.py
+++ b/supervision/detection/utils.py
@@ -592,6 +592,46 @@ def move_boxes(xyxy: np.ndarray, offset: np.ndarray) -> np.ndarray:
     return xyxy + np.hstack([offset, offset])
 
 
+def move_masks(
+    masks: np.ndarray, offset: np.ndarray, desired_shape: Optional[np.ndarray] = None
+) -> np.ndarray:
+    """
+    Offset the masks in an array by the specified (x, y) amount.
+
+    Note the axis orders:
+
+    - `masks`: array of shape `(n, y, x)`
+    - `offset`: array of ints: `(x, y)`
+    - `desired_shape`: array of ints `(x, y)`
+
+    Args:
+        masks (np.ndarray): array of bools
+        offset (np.ndarray): An array of shape `(2,)` containing non-negative int values
+            `[dx, dy]`.
+        desired_shape (Tuple[int, int], optional): Final shape of the mask in the format
+            `(width, height)`. If provided, the masks will be padded to match this
+            shape. Note the axis order (x,y)!
+
+    Returns:
+        (np.ndarray) repositioned masks, optionally padded to the specified shape.
+    """
+    if offset[0] < 0 or offset[1] < 0:
+        raise ValueError(f"Offset values must be non-negative integers. Got: {offset}")
+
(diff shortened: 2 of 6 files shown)


supervision/detection/utils.py
```diff
@@ -592,6 +592,46 @@ def move_boxes(xyxy: np.ndarray, offset: np.ndarray) -> np.ndarray:
...
+        (np.ndarray) repositioned masks, optionally padded to the specified shape.
+    """
+    if offset[0] < 0 or offset[1] < 0:
+        raise ValueError(f"Offset values must be non-negative integers. Got: {offset}")
+
+    size_x, size_y = masks.shape[1:] + offset[::-1]
+    if desired_shape is not None:
+        size_y, size_x = desired_shape[:2]
+
+    mask_arr = np.full((masks.shape[0], size_y, size_x), False)
```
SkalskiP, line 625: Please don't use abbreviations, especially when word `array` is only two characters longer than `arr` :)

supervision/detection/utils.py
```diff
@@ -592,6 +592,46 @@ def move_boxes(xyxy: np.ndarray, offset: np.ndarray) -> np.ndarray:
     return xyxy + np.hstack([offset, offset])
 
 
+def move_masks(
+    masks: np.ndarray, offset: np.ndarray, desired_shape: Optional[np.ndarray] = None
```
SkalskiP, line 596: Why:
- `desired_shape` is `Optional`?
- `desired_shape` is `np.ndarray` not tuple?
- `desired_shape` would have more elements than 2?

supervision/detection/tools/inference_slicer.py
```diff
@@ -4,20 +4,28 @@
...
 
 from supervision.detection.core import Detections
-from supervision.detection.utils import move_boxes
+from supervision.detection.utils import move_boxes, move_masks
 from supervision.utils.image import crop_image
 
 
-def move_detections(detections: Detections, offset: np.array) -> Detections:
+def move_detections(
+    detections: Detections, offset: np.ndarray, image_shape: np.ndarray
```
SkalskiP, line 12: Argument `image_shape` is called `resolution_wh` in most places in `supervision`. Let's keep it this way.

SkalskiP: Also, I think it should be `Optional`.

SkalskiP (changes requested):

SkalskiP:
Also please add `move_masks` import to `__init__.py`.

supervision/detection/utils.py
LinasKo: I see I made some rushed mistakes.

1) If no shape is passed, the mask will be moved to the bottom-right corner with no padding. If we want to support unchanged `move_detections` signature, we should add `Optional` there too. Otherwise, it's better if we remove this one.
2) `tuple` makes more sense. I did some refactoring in my final commit and did not revert this.

3) I think, the parameter is called shape, so it makes sense to support passing `np_array.shape` here, which may have shape (height, width, channels) or (height, width).
* Do you think it's complicated? I could fix it to 2 dims. I figured, the pattern of [:2] on image shapes is something I've seen numerous times - seems trivial.

supervision/detection/utils.py
```diff
@@ -592,6 +592,49 @@ def move_boxes(xyxy: np.ndarray, offset: np.ndarray) -> np.ndarray:
     return xyxy + np.hstack([offset, offset])
 
 
+def move_masks(
+    masks: np.ndarray,
+    offset: np.ndarray,
+    desired_shape: Optional[Union[Tuple[int, int, int], Tuple[int, int]]] = None,
```
SkalskiP, line 598: Why not just `resolution_wh`?

SkalskiP (changes requested):

supervision/detection/utils.py
SkalskiP: I think it should be just `resolution_wh` required argument.

supervision/detection/utils.py
LinasKo: Yeah, sounds good, will change it

supervision/detection/utils.py
```diff
@@ -592,6 +592,38 @@ def move_boxes(xyxy: np.ndarray, offset: np.ndarray) -> np.ndarray:
...
+def move_masks(
+    masks: np.ndarray,
+    offset: np.ndarray,
+    resolution_wh: Tuple[int, int] = None,
+) -> np.ndarray:
+    """
+    Offset the masks in an array by the specified (x, y) amount.
+
+    Args:
+        masks (np.ndarray): array of bools
```
SkalskiP, line 604: I'd make that `masks` docstring a bit more precise. Talk about shapes etc.

```
masks (np.ndarray): A 3D array of binary masks corresponding to the predictions.
    Shape: `(N, H, W)`, where N is the number of predictions, and H, W are the
    dimensions of each mask.
```

docs/how_to/detect_small_objects.md
```diff
@@ -264,3 +264,63 @@ objects within each, and aggregating the results.
...
 
 ![detection-with-inference-slicer](https://media.roboflow.com/supervision_detect_small_objects_example_3.png)
+
+## Small Object Segmentation
+
+[`InferenceSlicer`](/latest/detection/tools/inference_slicer/#supervision.detection.tools.inference_slicer.InferenceSlicer) can perform segmentation tasks too.
+
+=== "Inference"
+
+    ```{ .py hl_lines="6 16 19" }
```
SkalskiP, line 274: make it `python` not `{ .py hl_lines="6 16 19" }`

docs/how_to/detect_small_objects.md
```diff
@@ -264,3 +264,63 @@ objects within each, and aggregating the results.
...
+
+    annotated_image = mask_annotator.annotate(
+        scene=image, detections=detections)
+    annotated_image = label_annotator.annotate(
+        scene=annotated_image, detections=detections)
+    ```
+
+=== "Ultralytics"
+
+    ```{ .py hl_lines="6 16 19" }
```
SkalskiP, line 301: make it `python` not `{ .py hl_lines="6 16 19" }`

SkalskiP: `{ .py hl_lines="6 16 19-20" }`

SkalskiP (approved):

Issue #458, opened 2023-10-12
Cost Matrix contains NaN in ByteTrack() code 


tobieabel:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar bug report.

### Bug

I got the following error message when using supervision ByteTrack().  Something in the image is causing a NaN value to appear in the cost matrix in matching.py.  I have got around the issue for now by inserting the below line of code in bold to remove the NaN value from the matrix in line 32 of matching.py:

    cost_matrix[cost_matrix > thresh] = thresh +  1e-4
    **cost_matrix = np.nan_to_num(cost_matrix)**
    row_ind, col_ind = linear_sum_assignment(cost_matrix)
    indices = np.column_stack((row_ind, col_ind))

console output:

/Users/tobieabel/PycharmProjects/Demo_4_Fight_and_Fall/venv/lib/python3.11/site-packages/supervision/detection/utils.py:56: RuntimeWarning: invalid value encountered in divide
  return area_inter / (area_true[:, None] + area_detection - area_inter)
Traceback (most recent call last):
  File "/Users/tobieabel/PycharmProjects/Demo_4_Fight_and_Fall/Main.py", line 192, in <module>
    start()
  File "/Users/tobieabel/PycharmProjects/Demo_4_Fight_and_Fall/Main.py", line 171, in start
    obj_detections, class_results, Incident = start.Run_Models(frame)
                                              ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tobieabel/PycharmProjects/Demo_4_Fight_and_Fall/Main.py", line 76, in Run_Models
    obj_detections = self.tracker.update_with_detections(obj_detections)  # pass the detections through the tracker to add tracker ID as additional field to detections object
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tobieabel/PycharmProjects/Demo_4_Fight_and_Fall/venv/lib/python3.11/site-packages/supervision/tracker/byte_tracker/core.py", line 235, in update_with_detections
    tracks = self.update_with_tensors(
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tobieabel/PycharmProjects/Demo_4_Fight_and_Fall/venv/lib/python3.11/site-packages/supervision/tracker/byte_tracker/core.py", line 367, in update_with_tensors
[...]

PR #1077, opened 2024-03-30
adding ByteTrack unit tests

rolson24:
# Description

This PR adds unit tests for ByteTrack. They do not all pass because of a bug introduced in the last PR, but I have tested them 
on the current release of supervision and they pass correctly. They also pass after making the 1 line bug fix.
I chose the test cases to be the ones that I have encountered to be edge cases so far, as well as adding edge cases that I can imagine being problematic.

## Type of change

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)

## How has this change been tested, please provide a testcase or example of how you tested the change?

I tested this change, by first running these pytests on the 0.18.0 release version of supervision to ensure they all pass. I then used them to validate my other changes to ByteTrack to ensure the changes still function correctly. I also tested the current develop branch, and 1 test failed due to a small bug in how update_with_detections handles no valid tracked detections. (added here)

## Any specific deployment considerations

Need to make sure they tests will autorun on PR.

## Docs

-   [ ] Docs updated? What were the changes:

PR #1180, opened 2024-05-08
Fix Cost Matrix containing NaN Values in ByteTrack

rolson24:
# Description

This change fixes issue #458. The object tracker was not able to handle detections that had a bounding box area of zero. I fixed this by filtering out the bounding boxes that have an area of zero in `update_with_tensors()`. 

## Type of change

I tested this change by deliberately sending boxes with an area of 0 into the tracker. The tracker was able to handle this by simply ignoring those detections. [Colab notebook](https://colab.research.google.com/drive/1qDtagBbfLHVE8G-VhgBCOiK9H-GTP8c4?usp=sharing) with test code.

### DIFF
--- a/supervision/detection/utils.py
+++ b/supervision/detection/utils.py
@@ -56,7 +56,9 @@ def box_iou_batch(boxes_true: np.ndarray, boxes_detection: np.ndarray) -> np.nda
     bottom_right = np.minimum(boxes_true[:, None, 2:], boxes_detection[:, 2:])
 
     area_inter = np.prod(np.clip(bottom_right - top_left, a_min=0, a_max=None), 2)
-    return area_inter / (area_true[:, None] + area_detection - area_inter)
+    ious = area_inter / (area_true[:, None] + area_detection - area_inter)
+    ious = np.nan_to_num(ious)
+    return ious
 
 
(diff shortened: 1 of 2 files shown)


LinasKo:
Hi @rolson24 :wave: 

Well done isolating the error and finding the root cause!

What prompted me to dig for other solutions was the removal of `class_id`, boxes, etc. The algorithm downstream should be able to handle whatever comes its way, and more broadly - we had issues with similar approaches before.

There is a more concise solution, in fact. In `detection/utils.py`, there's a function `box_iou_batch`. Inside it, unsafe division is causing all of these NaN values. I reckon, instead of returning, it should first do `ious = np.nan_to_num(ious)`.

Would you have some time to update it? :wink:

Also, are the test in the Colab the same as in https://github.com/roboflow/supervision/pull/1077?

SkalskiP:
Hi, @rolson24 and @LinasKo! Thanks a lot! 🙏🏻 @rolson24 
The work you did to find the source of the problem was crucial. 
However, I think the solution proposed by @LinasKo is more advantageous. I am surprised that our IoU calculation code did not account for the possibility of a box having zero area.

SkalskiP (approved):

Issue #1092, opened 2024-04-04
sv.Detections has exactly N of each member, except for `tracker_id`.


LinasKo:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar bug report.

### Bug

Found in: #1044, #1076.

This is a sub-issue of #943 and should be solved before starting that one.

sv.Detections has a list of numpy arrays as its members: `xyxy`, `mask`, `confidence`, `class_id`, `trackier_id` When `N` detections are made, all of these except `tracker_id` will have a shape `(N, ...)`.

For example: When there are no tracks, `update_with_detections` will return detections, with `detections.tracker_id` set to `[]`.

BUT, a few lines above, it only returns those detections where `tracker_id` was not `-1`. Many places in the codebase, e.g. the `TraceAnnotator` expect len(tracker_id) to match len(detections). There's many zips in examples, zipping with `tracker_id`.

So there's 2 paradigms:
1) Make the tracker ID length always match the length of other members by having a special value such as `-1`, already used in some places.
2) Make `tracker_id` different than other members of `sv.Detections`, keep the current implementation, but root out incorrect usage.

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

LinasKo:
I'm heavily in favour of 1. The steps for this would be:

- [ ] Find, learn of all places where `tracker_id` is used.
- [ ] Modify the core of `ByteTrack` - if there is a detection but there's no tracker - instead of returning nothing, return `-1`. Make sure the length of `tracker_id` matches the length of `detections` being returned.
- [ ] Think a bit how this affects case where we deliberately hide uncertain trackers - this would be introduced as part of #1076.
- [ ] Propagate the changes in the core of `ByteTrack`, to the uses by various components - annotators, examples
- [ ] Some tests would be good to have.

LinasKo:
Example: line_zone.py:trigger() checks if `tracker_id` array contains `None`

LinasKo:
Example: smoother.py:update_with_detections() checks if iterating over `tracker_id` array produces `None`

LinasKo:
I was wrong about the `-1` case - the only issues that exist are unnecessary `None` checks.

When the related PR is merged, this can be closed.

PR #1205, opened 2024-05-17
Optimize line zone and add unit tests

tc360950:
# Description

My experiences with supervision's tracking indicate that (quite surprisingly) in some cases  [LineZone](https://github.com/roboflow/supervision/blob/develop/supervision/detection/line_zone.py#L12C7-L12C15) may be a bottleneck. 
Think about edge computing scenarios - where detection model is small and one tracks a lot of objects. In such cases every FPS counts. 
The following is a cProfile output for Yolov8m + ByteTrack + LineZone run 10 times on [market square video](https://github.com/roboflow/supervision/blob/develop/supervision/assets/list.py#L30). 

image

As you can see the cost of LineZone is already quite considerable. Now, consider that there are not a lot of objects (average of 35 per frame) and that the easiest, most common approach to improving counting precision is adding more lines and averaging results. 
For these reasons, in my usage of supervision, I've had to optimize [LineZone](https://github.com/roboflow/supervision/blob/develop/supervision/detection/line_zone.py#L12C7-L12C15).
The new implementation takes only 0.74 seconds on market square example compared to 11.1s seconds of the old implementation. To support this change I've added unit tests for LineZone. 

Note that quite unnecessarily internals of LineZone are exposed through two public static methods - [calculate_region_of_interest_limits](https://github.com/roboflow/supervision/blob/develop/supervision/detection/line_zone.py#L86) and [is_point_in_limits](https://github.com/roboflow/supervision/blob/develop/supervision/detection/line_zone.py#L118). I can only guess that it's because unit tests were just written for parts (encapsulated in those two methods) of LineZone code. These methods should not be public because they expose specifics of how line counting is implemented, hindering any attempts at changing the method (as is in this case). I've left those two methods unchanged, because I don't want to break peoples' code (even though those two methods were not documented as a part of API), but I've added docstring with warning that they are not a part of the API. 

## Type of change

-   [x] Feature optimization (non-breaking change which optimizes existing functionality)

New LineZone unit tests.

### DIFF
--- a/supervision/detection/line_zone.py
+++ b/supervision/detection/line_zone.py
@@ -53,17 +53,16 @@ class LineZone:
         # 7, 2
         ```
     """  # noqa: E501 // docs
-
     def __init__(
-        self,
-        start: Point,
-        end: Point,
-        triggering_anchors: Iterable[Position] = (
-            Position.TOP_LEFT,
-            Position.TOP_RIGHT,
-            Position.BOTTOM_LEFT,
-            Position.BOTTOM_RIGHT,
-        ),
+            self,
+            start: Point,
+            end: Point,
+            triggering_anchors: Iterable[Position] = (
+                    Position.TOP_LEFT,
+                    Position.TOP_RIGHT,
+                    Position.BOTTOM_LEFT,
+                    Position.BOTTOM_RIGHT,
+            ),
     ):
         """
         Args:
@@ -75,15 +74,29 @@ class LineZone:
                 has passed the line counter or not. By default, this
                 contains the four corners of the detection's bounding box
         """
-        self.vector = Vector(start=start, end=end)
-        self.limits = self.calculate_region_of_interest_limits(vector=self.vector)
+        self._vector = Vector(start=start, end=end)
+        if self._vector.magnitude == 0:
+            raise ValueError("The magnitude of the line cannot be zero.")
+        vector, rotation, translation = self._transform_to_vertical_at_zero(self._vector)
+        self.limits = self.calculate_region_of_interest_limits(vector=self._vector)
+        self._transformed_vector = vector
+        self._rotation = rotation
+        self._translation = translation
         self.tracker_state: Dict[str, bool] = {}
         self.in_count: int = 0
         self.out_count: int = 0
         self.triggering_anchors = triggering_anchors
 
+    @property
+    def vector(self):
+        return self._vector
+
     @staticmethod
     def calculate_region_of_interest_limits(vector: Vector) -> Tuple[Vector, Vector]:
+        """
+        !!! Warning "This method is not a part of the API and as such it's
+        not stable and may change in the future. Use at your own risk."
+        """
         magnitude = vector.magnitude
 
         if magnitude == 0:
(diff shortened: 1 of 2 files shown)


LinasKo:
Hi @tc360950 :wave: 

Looking purely at the numbers, it's a massive improvement. I'll see if I can review it soon.

Is this in comparison with the `develop` branch (`supervision==0.21.0rc5`), or the main `supervision` package?

LinasKo:
Am I reading the graph right - is `cross_product` the most expensive operation here? Is that the main reason for transforming to a vertical-line system?

tc360950:
Exactly, that's why it's easier to transform to coordinates where the line is vertical - then everything can be vectorized.

LinasKo:
@SkalskiP, when you're back - I'd like your input. I think this is a fascinating case where an issue is revealed, and the problem is solved, but with a method that's too complicated for us. Specifically, I see that:

* LineZone takes too long. Very obviously so.
* We're repeatedly calling the non-vectorized `vector.cross_product`. It would be much faster if vectorized, possibly precomputed before the loop - similar to `in_limit` here.
* I think there's a `tracker_id` inconsistency here - I can't recall when we'd expect a `None` while iterating over `tracker_id`. The None check here is mildly getting in the way of vectorization. Related: https://github.com/roboflow/supervision/issues/1092
* I wholeheartedly agree that `is_point_in_limits` and `calculate_region_of_interest_limits` should be flagged as private to this class.
* A lot of work went into producing the tests - we should keep those, with full credit to @tc360950.

Let's wait for @SkalskiP, but my verdict is that to merge this, I'd need to know it's significantly more efficient than a solution where we use vectorized and possibly precomputed cross products.

Meanwhile, @tc360950, if you have some time, would you be able to split the test into a separate PR? I assume they'd apply just as well for the current solution.

LinasKo:
Thanks! To add more freedom, I'd say, implementations of `trigger` don't have to use the inefficient `calculate_region_of_interest_limits` and `is_point_in_limits` methods.

If there's a significantly faster way to line count, we can deprecate them.

SkalskiP:
Hi @tc360950 👋🏻  Improving LineZone performance by 10-20x sounds fantastic, but like @LinasKo, I am concerned that the proposed changes significantly increase complexity. I strive to keep the Supervision codebase relatively simple to understand, making maintenance easier and lowering the entry barrier for future contributors.

Therefore, I'd like to propose the following course of action:

- Merge the LineZone unit tests.
- Profiling analysis shows that most time is spent on is_point_in_limits and cross_product because these calculations are done in a for loop instead of being vectorized. I propose writing an alternative implementation where we minimize for loops and process as much as possible in parallel using numpy. If you have time, @tc360950, I think this approach has the potential to achieve similar optimization levels. If you're willing, please open a separate PR and conduct similar profiling.
- Depending on the results, we will choose between the option proposed in this PR and the new vectorized approach.

@tc360950, please let me know your thoughts.

tc360950:
@SkalskiP I think it's a good idea. I actually have a working draft of vectorization of the original method. I'll open a separate PR soon.

SkalskiP:
Hi @tc360950 👋🏻 I just wanted to check in and see how it's going. Do you need any help?

tc360950:
@SkalskiP I was waiting for the unit test PR :) Here's the [PR](https://github.com/roboflow/supervision/pull/1228) with vectorized LineZone.

LinasKo:
Both the unit tests and the vectorized LineZone implementation are now merged. Thank you for very much @tc360950 and @SkalskiP!

We even managed to make it in time for `0.21.0` release. 🙂

PR #1206, opened 2024-05-17
Add LineZone unit tests

tc360950:
# Description
Added unit tests for [LineZone](https://github.com/roboflow/supervision/blob/develop/supervision/detection/line_zone.py#L12C7-L12C15).

## Type of change

-   [x] New unit tests

### DIFF
--- a/test/detection/test_line_counter.py
+++ b/test/detection/test_line_counter.py
@@ -1,10 +1,13 @@
 from contextlib import ExitStack as DoesNotRaise
-from typing import Optional, Tuple
+from itertools import chain, combinations
+from test.test_utils import mock_detections
+from typing import Optional, Tuple, List
 
+import numpy as np
 import pytest
 
-from supervision import LineZone
-from supervision.geometry.core import Point, Vector
+from supervision import Detections, LineZone
+from supervision.geometry.core import Point, Position, Vector
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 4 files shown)


test/detection/test_line_counter.py
```diff
@@ -70,3 +73,320 @@ def test_calculate_region_of_interest_limits(
...
+            ],
+        ),
+    ],
+)
+def test_line_zone_does_not_count_detections_without_tracker_id(vector, bbox_sequence):
+    line_zone = LineZone(start=vector.start, end=vector.end)
+    for bbox in bbox_sequence:
+        detections = Detections(
+            xyxy=np.array([bbox]).reshape((-1, 4)),
+            tracker_id=np.array([None for _ in range(0, 1)]),
```
LinasKo, line 388: 1. That's just [0]. No need for a for loop.
2. I don't think `None` can appear in `tracker_id`. But let's pretend it can, and I'll track it down in #1092
    * If you've seen `None` inside `tracker_id` when testing on your videos, do tell :slightly_smiling_face:

test/detection/test_line_counter.py
```diff
@@ -70,3 +73,320 @@ def test_calculate_region_of_interest_limits(
...
+    ):
+        if not anchors:
+            continue
+        line_zone = LineZone(
+            start=vector.start, end=vector.end, triggering_anchors=anchors
+        )
+        for i, bbox in enumerate(bbox_sequence):
+            detections = mock_detections(
+                xyxy=[bbox],
+                tracker_id=[i for i in range(0, 1)],
```
LinasKo, line 279: Let's make it simple to read and change to [0].

test/detection/test_line_counter.py
```diff
@@ -70,3 +73,320 @@ def test_calculate_region_of_interest_limits(
...
+    ],
+)
+def test_line_zone_single_detection(
+    vector, bbox_sequence, expected_count_in: List[bool], expected_count_out: List[bool]
+) -> None:
+    line_zone = LineZone(start=vector.start, end=vector.end)
+    for i, bbox in enumerate(bbox_sequence):
+        detections = mock_detections(
+            xyxy=[bbox],
+            tracker_id=[i for i in range(0, 1)],
```
LinasKo, line 213: Let's make it simple to read and change to [0].

LinasKo (changes requested):

test/detection/test_line_counter.py
tc360950: Good idea, fixed.

test/detection/test_line_counter.py
```diff
@@ -70,3 +73,320 @@ def test_calculate_region_of_interest_limits(
...
+            ],
+        ),
+    ],
+)
+def test_line_zone_does_not_count_detections_without_tracker_id(vector, bbox_sequence):
+    line_zone = LineZone(start=vector.start, end=vector.end)
+    for bbox in bbox_sequence:
+        detections = Detections(
+            xyxy=np.array([bbox]).reshape((-1, 4)),
+            tracker_id=np.array([None]),
```
LinasKo, line 388: This was a loop of `[None for x in range(0,1)]`, so we correctly simplified to `[None]`.

However, is this right? I think the length of `tracker_id` does not match `xyxy`.
Should we change this to `tracker_id=[i for i in range(0, len(bbox))],`?

LinasKo (changes requested):

test/detection/test_line_counter.py
```diff
@@ -70,3 +73,320 @@ def test_calculate_region_of_interest_limits(
     with exception:
         result = LineZone.calculate_region_of_interest_limits(vector=vector)
         assert result == expected_result
+
+
+@pytest.mark.parametrize(
+    "vector, bbox_sequence, expected_count_in, expected_count_out",
```
SkalskiP, line 79: Please rename the parameters `expected_count_in` and `expected_count_out` to `expected_crossed_in` and `expected_crossed_out` respectively. `count_in` and `count_out` have a well defined meaning in context of `LineZone`.

test/detection/test_line_counter.py
```diff
@@ -70,3 +73,320 @@ def test_calculate_region_of_interest_limits(
...
+def test_line_zone_single_detection(
+    vector, bbox_sequence, expected_count_in: List[bool], expected_count_out: List[bool]
+) -> None:
+    line_zone = LineZone(start=vector.start, end=vector.end)
+    for i, bbox in enumerate(bbox_sequence):
+        detections = mock_detections(
+            xyxy=[bbox],
+            tracker_id=[0],
+        )
+        count_in, count_out = line_zone.trigger(detections)
```
SkalskiP, line 215: Once again, please rename `count_in` and `count_out` to `crossed_in` and `crossed_out` respectively.

test/detection/test_line_counter.py
```diff
@@ -70,3 +73,320 @@ def test_calculate_region_of_interest_limits(
...
+                [-50, 70, -40, 50],
+                [-50, -70, -40, -50],
+            ],
+            [False, False, True, False, True, False, True, False],
+            [False, True, False, True, False, True, False, True],
+        ),
+    ],
+)
+def test_line_zone_single_detection(
+    vector, bbox_sequence, expected_count_in: List[bool], expected_count_out: List[bool]
```
SkalskiP, line 207: Once again, please rename `expected_count_in` and `expected_count_out` to `expected_crossed_in` and `expected_crossed_out` respectively.

SkalskiP: Provide type hints for all arguments.

test/detection/test_line_counter.py
SkalskiP: Please do the same for all occurrences.

test/detection/test_line_counter.py
SkalskiP: I'd recommend merging `test_line_zone_does_not_count_detections_without_tracker_id` with `test_line_zone_single_detection`. Case is almost the same.

SkalskiP (changes requested):

test/detection/test_line_counter.py
tc360950: @SkalskiP it's OK - test's input is a sequence of positions of single tracklet, each position described by 4 ints (which are xyxy of the bounding box) .

tc360950: @SkalskiP I have mixed feelings about this. I understand your point that the inputs and code are the same, but this test tests different functionality. LinasKo claims that occurrences of 'None' 'tracker_id' are an unconfirmed anomaly that should be eradicated. Maybe it's better to leave it as a separate test in case some day it becomes settled that 'tracker_id' can't be None (and hence this test becomes obsolete and fails). But I leave this decision up to you.

LinasKo: Hi @tc360950,

It was bothering me so I verified it - values inside `tracker_id` are always integers and never `None`. Please remove the tests testing for `None` 😉

SkalskiP: Agree. As far as I know, per box, `tracker_id` should not be `None`. 

But I meant to merge those two suits and potentially use the `exception` argument to handle cases that throw. You can find an example of this in [`test_calculate_region_of_interest_limits`](https://github.com/roboflow/supervision/blob/1bbe03ca27dda82b8910feb0352dfb3a443f970d/test/detection/test_line_counter.py#L65). We adopted this convention in other unit tests in `supervision` and I would like to maintain it.

test/detection/test_line_counter.py
```diff
@@ -70,3 +73,332 @@ def test_calculate_region_of_interest_limits(
...
+                [-100, 50, -80, 70],
+                [-100, 50, 120, 70],
+            ],
+            [False, True],
+            [False, False],
+            [Position.TOP_RIGHT, Position.BOTTOM_RIGHT],
+        ),
+    ],
+)
+def test_line_zone_single_detection_on_subset_of_anchors(
```
SkalskiP, line 260: @tc360950 or @LinasKo, could you explain a bit more about what we are testing here?

test/detection/test_line_counter.py
tc360950: @SkalskiP since there's an open PR by LinasKo which removes check for 'None' 'tracker_id' I've deleted this test.

test/detection/test_line_counter.py
tc360950: @SkalskiP Other tests assume that the line is crossed with every anchor. This test checks LineZone behaviour when only a subset of anchors cross the line. For instance if only `[Position.BOTTOM_LEFT, Position.TOP_LEFT]` cross the line we check if 
LineZone returns correct results for (indicated as "crossing"):
`[Position.BOTTOM_LEFT, Position.TOP_LEFT]`
`[Position.TOP_LEFT]`
`[Position.BOTTOM_LEFT]`

and correct result for (indicated as not "crossing"):
`[Position.BOTTOM_LEFT, Position.TOP_LEFT, Position.TOP_RIGHT]`
`[Position.TOP_LEFT, Position.BOTTOM_RIGHT]`
`[Position.BOTTOM_RIGHT]`
etc. 

I've added clarifying docstring to every test.

LinasKo:
Thank you for the changes, @tc360950 !

Would you have some time to add a few extra cases? What I'm thinking of is the following:

* object crossing pependicular to the line, but to its side. As if crossing the line extended to infinity, but not the segment defined by start-end.
* diagonal line
* negative coordinates
 * Ensure error is raised when no anchors are defined

Does that sound like something you'd have time for, by any chance?

tc360950:
@LinasKo Ok, I will add the changes. Could you clarify what you mean by "diagonal line" - there are already test cases for lines which are not horizontal or vertical. Do you want more (e.q. with different slope)?

LinasKo:
Ah, indeed non-vertical and non-horizontal is what I meant. If we have it, we don't need any additional tests for it.

tc360950:
@LinasKo I've added changes you had requested

LinasKo:
Hey @tc360950,

Apologies for the delay! To save time, I've pushed a few updates to the branch.
Despite our questions, the function names were descriptive enough so that docstrings aren't needed. We typically go very light on comments if there's code explaining the behaviour nearby.
Also, while not a typical practice, we internally agreed to ship a similar `ValueError` check to `PolygonZone`. I've added some tests for that.

@SkalskiP, This is ready to merge, but your review is outstanding. When you have the time, please resolve it ;)

LinasKo:
I've added additional tests but found a case where LineZone does not work.
Crossing the line while moving from outside to inside the limits, or vice versa, does not count as a crossing.

This is an issue for large jumps or short lines.

I did not test cases with multiple lines present, for I don't see how they would interact.

Happy to merge this, ahead of the vectorized solution.

tc360950:
> I've added additional tests but found a case where LineZone does not work. Crossing the line while moving from outside to inside the limits, or vice versa, does not count as a crossing.
> 
> This is an issue for large jumps or short lines.

Isn't it an intended result? LineZone is "dumb" in the sense that it does not look at trajectories, there's no correlation between subsequent positions of an object.
For instance: 
If an objects appears "in limits" on the left of the line, then goes around it and lands "in limits" on the right hand side it will be counted as crossing.

LinasKo:
@tc360950 potentially intended, potentially not. If we assume small time increments, we're also in many cases assuming direct motion.

I can imagine an algo where we check for whether each anchor displacement line crosses the LineZone (no longer a zone), which would work for motions coming from outside the limits. I think it's more intuitive for the users as well.

However, that's a long-term thought. For now, we're 100% sticking to your vectorized solution.

LinasKo:
Hi @SkalskiP,

I believe this is ready to merge. When you have time, please have a look.

SkalskiP (approved):

PR #1214, opened 2024-05-21
tracker_id contents are never `None`

LinasKo:
# Description

I've verified every instance of `tracker_id` in our repo, and found that its **contents** are never `None`.
The checks we have are not necessary.

In cases where `tracker_id` gets `-1` added, it is indeed filtered out very soon.

## Type of change

Basic tests of smoother and line counter can be found here: https://colab.research.google.com/drive/1_gajR0D_CZ4Z2dZ1y8SdVaWGbDZ0mBpl?usp=sharing

Even with no detections for longer periods of time, no errors (and no None values) are produced.

None

## Docs

### DIFF
--- a/supervision/detection/line_zone.py
+++ b/supervision/detection/line_zone.py
@@ -140,6 +140,13 @@ class LineZone:
         if len(detections) == 0:
             return crossed_in, crossed_out
 
+        if detections.tracker_id is None:
+            print(
+                "Line zone conting skipped. LineZone requires tracker_id. Refer to "
+                "https://supervision.roboflow.com/latest/trackers for more information."
+            )
+            return crossed_in, crossed_out
+
         all_anchors = np.array(
             [
                 detections.get_anchors_coordinates(anchor)
@@ -148,9 +155,6 @@ class LineZone:
         )
 
         for i, tracker_id in enumerate(detections.tracker_id):
-            if tracker_id is None:
-                continue
-
             box_anchors = [Point(x=x, y=y) for x, y in all_anchors[:, i, :]]
 
             in_limits = all(
--- a/supervision/detection/line_zone.py
+++ b/supervision/detection/line_zone.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import Dict, Iterable, Optional, Tuple
 
 import cv2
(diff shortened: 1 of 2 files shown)


supervision/detection/line_zone.py
```diff
@@ -140,6 +140,13 @@ def trigger(self, detections: Detections) -> Tuple[np.ndarray, np.ndarray]:
         if len(detections) == 0:
             return crossed_in, crossed_out
 
+        if detections.tracker_id is None:
+            print(
```
onuralpszr, line 144: Instead of print what about extend or use our warning system 

https://github.com/roboflow/supervision/blob/2bd4348e5d05f386559ea2d526112216692b1ae2/supervision/utils/internal.py#L7

At least people can turn off instead of forceful print

onuralpszr:
fyi, @LinasKo collab access is not public

LinasKo:
@onuralpszr, thank you - I've opened it up

onuralpszr:
@LinasKo how about make another rc release with it ? For better tracker fix ?

LinasKo:
> @LinasKo how about make another rc release with it ? For better tracker fix ?

That's outside my area of responsibility :slightly_smiling_face:

Our main goal, however, is working towards the new release, rather than pushing out more candidate packages.

SkalskiP:
Hi @onuralpszr 👋🏻 why do you think we should release a new rc?

onuralpszr:
> Hi @onuralpszr 👋🏻 why do you think we should release a new rc?

Well, current release has tracker issue and with those fix and this extra addition should be live for others to easy to use. Personally, I wish we should do point release instead of major release.

onuralpszr:
I can also update current doc version without even change version number If you like (like last time we did)

SkalskiP (approved):

SkalskiP:
> Well, current release has tracker issue and with those fix and this extra addition should be live for others to easy to use.

I understand. I'm also not particularly happy that the latest version has had this bug for so long, but we will have the next release soon – in a week. We still have some inconsistencies, such as [XXX](https://github.com/roboflow/supervision/issues/1215). I would prefer to resolve these before the release.

onuralpszr:
> > Well, current release has tracker issue and with those fix and this extra addition should be live for others to easy to use.
> 
> I understand. I'm also not particularly happy that the latest version has had this bug for so long, but we will have the next release soon – in a week. We still have some inconsistencies, such as #1215. I would prefer to resolve these before the release.

If time is close, We should be fine.

PR #1221, opened 2024-05-22
initial commit adding support for `from_lmm` and specifically for Pal…

SkalskiP:
Initial support for `from_lmm` and specifically for `PaliGemma`.

### DIFF
--- a/supervision/detection/core.py
+++ b/supervision/detection/core.py
@@ -7,6 +7,7 @@ from typing import Any, Dict, Iterator, List, Optional, Tuple, Union
 import numpy as np
 
 from supervision.config import CLASS_NAME_DATA_FIELD, ORIENTED_BOX_COORDINATES
+from supervision.detection.lmm import LMM, validate_lmm_and_kwargs, from_paligemma
 from supervision.detection.utils import (
     box_non_max_suppression,
     calculate_masks_centroids,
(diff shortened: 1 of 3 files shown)


supervision/detection/core.py
```diff
@@ -805,6 +806,52 @@ def from_paddledet(cls, paddledet_result) -> Detections:
...
+        Raises:
+            ValueError: If the LMM is invalid, required arguments are missing, or
+                disallowed arguments are provided.
+            ValueError: If the specified LMM is not supported.
+
+        Examples:
+            ```python
+            import supervision as sv
+
+            paligemma_result = "<loc0256><loc0256><loc0768><loc0768> cat"
```
merveenoyan, line 832: I think this would be also useful for models like KOSMOS-2 so best to make it very general (this is a trend with VLMs these days) https://huggingface.co/docs/transformers/en/model_doc/kosmos-2#transformers.Kosmos2ForConditionalGeneration.forward.example

SkalskiP: So what I want to do is one `from_lmm` function, providing separate dedicated parsers for each model. :)

LinasKo:
@SkalskiP, very quick one: I've just noticed that `LLM` is not in `__init__.py`, so the example with `sv.LLM.PALIGEMMA` didn't work.

SkalskiP:
@LinasKo Good catch! I just fixed it.

Issue #1096, opened 2024-04-05
[DetectionDataset] - expand `from_yolo` to include support for OBB (Oriented Bounding Boxes)


pedbrgs:
### Description

In [supervision-0.18.0](https://github.com/roboflow/supervision/releases/tag/0.18.0), we added initial support for OBB; it's time to extend it to include dataset loading.

Make the necessary changes in [sv.DetectionDataset.from_yolo](https://supervision.roboflow.com/0.19.0/datasets/#supervision.dataset.core.DetectionDataset.from_yolo) to enable loading OBB datasets from disk in YOLO format. [Here](https://docs.ultralytics.com/datasets/obb/) you can read more about the YOLO OBB Format. In short, each line of the `.txt` file should have the following format.

```
class_index, x1, y1, x2, y2, x3, y3, x4, y4
```

The [sv.OrientedBoxAnnotator](https://supervision.roboflow.com/0.19.0/annotators/#supervision.annotators.core.OrientedBoxAnnotator.annotate) expects information about oriented bounding boxes to be stored in the `xyxyxyxy` field of [`sv.Detections.data`](https://github.com/roboflow/supervision/blob/4729e20a9408fbe08d342cc4dbae835d808686a5/supervision/detection/core.py#L88). Ensure that the information loaded from the dataset is stored there.

### API

Here's an example of how to use the new API. Roboflow allows for the export of segmentation datasets as OBB. Let's ensure that our support for OBB definitely works with datasets exported from Roboflow.

```python
import random
import roboflow
from roboflow import Roboflow
import supervision as sv

roboflow.login()
rf = Roboflow()

project = rf.workspace("roboflow-jvuqo").project("fashion-assistant")
version = project.version(3)
dataset = version.download("yolov8-obb")

train_ds = sv.DetectionDataset.from_yolo(
    images_directory_path=f"{dataset.location}/train/images",
    annotations_directory_path=f"{dataset.location}/train/labels",
    data_yaml_path=f"{dataset.location}/data.yaml"
)

image_name = random.choice(list(train_ds.images))
image = train_data.images[image_name]
detections = train_data.annotations[image_name]

oriented_box_annotator = sv.OrientedBoxAnnotator()
annotated_frame = oriented_box_annotator.annotate(
    scene=image.copy(),
    detections=detections
)
```

### Additional

[...]

PR #1227, opened 2024-05-24
Adding support for Yolo-OBB (Oriented Bounding Boxes) format

Bhavay-2001:
# Description

PR for issue #1096 

## Type of change

## Docs

Tagging @LinasKo @SkalskiP


Bhavay-2001:
Hi @SkalskiP, any updates on this?

Bhavay-2001:
Hi @SkalskiP, can you please review this once and leave comments?

LinasKo:
Hi @Bhavay-2001,

Apologies for the wait and the lack of responses in the issue page. Submitting a PR with your proposal of the solution is the right way to go. I'll review it asap.

Meanwhile, what would help is having a [Colab Notebook](https://colab.research.google.com/) that shows shows how the changes work. Us reviewers get stuck in tons of different issues and have little time to give to each one. :slightly_smiling_face:

LinasKo:
Converting to draft for now, until we have a working implementation

LinasKo:
Copying your old comment from the issue:

> What I meant was that this [line](https://github.com/roboflow/supervision/blob/d4541e1469b1e89c057d7c5bd6edcfde0ee73aca/supervision/dataset/formats/yolo.py#L89) reshapes the annotations to a 4x2 matrix in case we have 8 numbers.
> And this [line](https://github.com/roboflow/supervision/blob/d4541e1469b1e89c057d7c5bd6edcfde0ee73aca/supervision/dataset/formats/yolo.py#L90) calculates the min and max values from the 4x2 matrix.
> 
> What I am suggesting is that maybe we can add a parameter which when set True can simply multiply the annotations with width and height and return it.

Within the `yolo_annotations_to_detections()` method, the goal is to insert the OBB into `Detections.data` under the key `ORIENTED_BOX_COORDINATES`, as that's where OBB data is set by our [other connects](https://github.com/roboflow/supervision/blob/d4541e1469b1e89c057d7c5bd6edcfde0ee73aca/supervision/detection/core.py#L255).

Looking at the the lines you mentioned, they seem to treat everything above `len==5` as polygon, drawing a bounding box around for `xyxy`. Indeed, we need to clearly distinguish when `OBB` output is required and only set the `Detections.data[ORIENTED_BOX_COORDINATES]` key when we're 100% sure that's what the user intends. Therefore, I believe you're on the right track and `as_obb: bool` is required here too.

supervision/dataset/formats/yolo.py
```diff
@@ -86,30 +87,47 @@ def yolo_annotations_to_detections(
...
+            if is_obb:
+                polygon = _parse_polygon(values=values[1:])
+                relative_xyxyxyxy.append(np.array(values[1:]))
+                if with_masks:
+                    relative_polygon.append(polygon)
+            else:
+                polygon = _parse_polygon(values=values[1:])
+                relative_xyxy.append(polygon_to_xyxy(polygon=polygon))
+                if with_masks:
+                    relative_polygon.append(polygon)
```
LinasKo, line 99: If we're doing the same operations inside both `if` branches, let's move them outside.

```
polygon = _parse_polygon(values=values[1:])
if with_masks:
        relative_polygon.append(polygon)
```

can be moved above `if is_obb`

supervision/dataset/formats/yolo.py
```diff
@@ -86,30 +87,47 @@ def yolo_annotations_to_detections(
...
 
     class_id = np.array(class_id, dtype=int)
     relative_xyxy = np.array(relative_xyxy, dtype=np.float32)
+    relative_xyxyxyxy = np.array(relative_xyxyxyxy, dtype=np.float32)
     xyxy = relative_xyxy * np.array([w, h, w, h], dtype=np.float32)
+    xyxyxyxy = relative_xyxyxyxy * np.array([w, h, w, h], dtype=np.float32)
```
LinasKo, line 105: Does this work? My guess would be that if we wish to expand `xyxyxyxy` to the dimensions of the whole image, you need `[w, h, w, h, w, h, w, h]`.

supervision/dataset/formats/yolo.py
```diff
@@ -86,30 +87,47 @@ def yolo_annotations_to_detections(
...
+
+    data = {ORIENTED_BOX_COORDINATES: xyxyxyxy}
 
     if not with_masks:
+        if is_obb:
+            return Detections(class_id=class_id, data=data)
```
LinasKo, line 111: `xyxy` in Detections is required - it'll fail without it.

supervision/dataset/formats/yolo.py
```diff
@@ -86,30 +87,47 @@ def yolo_annotations_to_detections(
...
         return Detections(class_id=class_id, xyxy=xyxy)
 
     polygons = [
         (polygon * np.array(resolution_wh)).astype(int) for polygon in relative_polygon
     ]
     mask = _polygons_to_masks(polygons=polygons, resolution_wh=resolution_wh)
-    return Detections(class_id=class_id, xyxy=xyxy, mask=mask)
+    return (
+        Detections(class_id=class_id, data=data, mask=mask)
```
LinasKo, line 119: Same as above - you'll need `xyxy`.

LinasKo (changes requested):
Very good attempt!

I left a few review comments. Nothing major, though it will be important to test in the Colab to check if everything works well.

supervision/dataset/formats/yolo.py
Bhavay-2001: Okay, will do.

supervision/dataset/formats/yolo.py
Bhavay-2001: Okay. I will change.

supervision/dataset/formats/yolo.py
Bhavay-2001: Okay. I will add.

Bhavay-2001:
Hi @LinasKo, could you please let me know how can I create a Colab notebook to test these changes. I mean any sample/steps I can follow to check this?
Thanks

LinasKo:
To test, you'd need the following:
1. Create a new notebook in [Google Colab](https://colab.google/)
2. Add a command to install your branch, `!pip install https://github.com/Bhavay-2001/roboflow-supervision.git@OBB`
3. Get the data, by either:
  * Uploading the dataset images and annotations to the colab directly
  * Uploading to Roboflow App and downloading like [in this example](https://supervision.roboflow.com/latest/datasets/core/#supervision.dataset.core.DetectionDataset.from_yolo)
4. Call `from_yolo`, check that OBB dataset is created and has the OBB boxes inside. You can probably print the object and see the values.
5. Call `yolo_annotations_to_detections` and see that the extra field in `data` is added and filled with obb values.

Bhavay-2001:
Hi Linas, can I change the tests in this file only `test\dataset\formats\test_yolo.py`. Will it count as testing the new feature only or creating a Colab notebook will be a better idea?

LinasKo:
Adding tests would be useful, but Colabs are the primary way we verify feature correctness.

P.S. I've updated step 2, as I forgot `git+` in pip install command.

Bhavay-2001:
Soo the checks suggests that some of the tests on `yolo_annotations_to_detections` failed because I didn't made the changes to the function calling in the test file. 

I wanted to discuss a thing for [this](https://github.com/roboflow/supervision/blob/7f78f93407a13a9f73ef547af85dbda29f857b08/test/dataset/formats/test_yolo.py#L173) function. In this function, I should be adding the is_obb parameter too in order for the test to pass, but should it be passed as a single boolean value or list of boolean values, because in the tests above, some test have only 1 annotation while some have 2, so it really depends on the size of the annotation. Could you suggest something here?

LinasKo:
Indeed, `is_obb` should have a default value of `False`, so no one's code fails after the update. This should fix test failures.

Where modifications to tests are needed, you may design the inputs whichever way it feels better.

LinasKo:
Hi @Bhavay-2001 , based on a quick glance, it looks good. Would you be able to share the dataset? The simplest option would be to put it on Google Drive, and create a shareable link. It can be zipped into one file, if that's more convenient for you.

Bhavay-2001:
[dataset.zip](https://github.com/user-attachments/files/15823758/dataset.zip)

LinasKo:
Yes, thank you very much!
I'll make sure to review this as soon as I can.

Bhavay-2001:
Hi @LinasKo, any updates on this?

LinasKo:
Hi @Bhavay-2001,

Apologies for the wait. I've added a bit of code that adapts this to our system.
It turns out, `ultralytics` provides `[[x, y], [x, y], [x, y], [x, y]]`, and that's what our Obb annotator expects.

Here's an upgraded version of your Colab:
https://colab.research.google.com/drive/1ld5aiOGrhaWzfN1HNNnxMuyy9PQhIrVn?usp=sharing

Overall, I believe this can now be merged. And if other reviewers request a change, I'll update the code.

LinasKo (approved):

PR #1228, opened 2024-05-24
Vectorize line zone

tc360950:
# Description

This is a follow-up to [1205](https://github.com/roboflow/supervision/pull/1205).
In this PR I've vectorized LineZone to achieve better performance. 

As previously all three approaches (the one from `develop`, the one from [1205](https://github.com/roboflow/supervision/pull/1205), and the one from this PR)
have been run on Yolov8m + ByteTrack + LineZone run 10 times on [market square video](https://github.com/roboflow/supervision/blob/develop/supervision/assets/list.py#L30). 
This time I've also checked what happens when all detections are duplicated X times (resulting in a larger number of tracked objects). 
The total time each LineZone version takes to compute on all frames is (seconds):

| Version / Number of duplicates  | 1 (no duplicates) | 2 | 4 |8
| ------------- | ------------- | ------| ------| ------|
| `develop`  | 11.1 | 33 | 77 | 164|
| previous PR  | 0.71 | 1.7 | 3.1 | 5.53 |
| This PR (vectorized version of `develop`  | 1.17 | 2.7 | 5 | 8.7|

As you can see the one from previous PR always "wins" but I find the difference small enough, especially if you find this version more readable.

## Type of change

-   [x] Optimization (non-breaking change which optimizes existing feature)


supervision/detection/line_zone.py
```diff
@@ -158,28 +158,29 @@ def trigger(self, detections: Detections) -> Tuple[np.ndarray, np.ndarray]:
...
+        # anchor is in limits if it's on the same side of both limit vectors
+        in_limits = ~np.logical_xor(cross_products_1 > 0, cross_products_2 > 0)
+        # Reduce array to find out if all anchors for a detection are within limits
+        in_limits = np.min(in_limits, axis=0)
+
+        # Calculate which anchors lie to the left of the line
+        triggers = self._cross_product(all_anchors, self.vector) < 0
+        # Reduce to find out if all anchors for a
+        # detection lie to the left (or right) of the line
+        max_triggers = np.max(triggers, axis=0)
```
SkalskiP, line 172: Hi @tc360950 👋🏻 

Could you explain what is the interpretation of:

```python
max_triggers = np.max(triggers, axis=0)
min_triggers = np.min(triggers, axis=0)

...

if min_triggers[i] != max_triggers[i]:
   ...
```

SkalskiP:
Hi @tc360950 👋🏻 
Considering your results and analyzing the increase in code complexity, I definitely lean towards choosing this PR over https://github.com/roboflow/supervision/pull/1205.

supervision/detection/line_zone.py
tc360950: @SkalskiP I don't like the name triggers but that's what the corresponding variable is named on `develop`.
Lines: 
```
triggers = [
                self.vector.cross_product(point=anchor) < 0 for anchor in box_anchors
            ]

if len(set(triggers)) == 2:
    ...
```
Are replaced with: 
```
triggers = self._cross_product(all_anchors, self.vector) < 0
# Reduce to find out if all anchors for a
# detection lie to the left (or right) of the line
max_triggers = np.max(triggers, axis=0)
min_triggers = np.min(triggers, axis=0)
...

if min_triggers[i] != max_triggers[i]:
   ...
```

As you can see "new" `triggers` is just a vectorized version of the "old" `triggers`, i.e. matrix which holds cross product between an anchor and the vector for all `(anchor, detection)` pairs. Now, this cross product tells us whether a given anchor is "to the left"or "to the right" of the line. We update tracklet's state only if all its anchors are on the same side of the line. 
The following line from `develop` checks if all anchors of a detection are on the same side of the line:
```
if len(set(triggers)) == 2:
```
The lines you are referring to are a direct equivalent of this line for the vectorized version:
Triggers is a boolean array of shape `(number of anchors, detections)`.
If we want to know if all anchors for a detection with index `i` lie on the same side of the line we need to check that:
```
[triggers[anchor_1, i] == triggers[anchor_2, i] for (anchor_1, anchor_2) in possible_anchors]
```
Since `triggers` has at most two values (True / False) it's easier and more efficient to just calculate maximum and minimum value for each detection and then check if they are equal. 

What if I replaced: 
```
max_triggers = np.max(triggers, axis=0)
min_triggers = np.min(triggers, axis=0)

...

if min_triggers[i] != max_triggers[i]:
   ...
```

With something like: 
```
ANCHOR_AXIS = 0
max_triggers = np.max(triggers, axis=ANCHOR_AXIS)
min_triggers = np.min(triggers, axis=ANCHOR_AXIS)
all_anchors_on_the_same_side_of_line = max_triggers == min_triggers

...

if all_anchors_on_the_same_side_of_line[i]:
   ...
```
Would you find it more readable?

SkalskiP: @tc360950, thanks a lot for the explanation. I understand everything now.

SkalskiP, line 172: @tc360950 what do you think about calculating this slightly differently, but at the same time naming variables in a more descriptive way:

```python
has_any_left_trigger = np.any(triggers, axis=1)
has_any_right_trigger = np.any(~triggers, axis=1)
is_uniformly_triggered = ~(has_any_left_trigger & has_any_right_trigger)
```

I don't have a particularly strong opinion on this. If you think it's better to stick with your names, we can merge.

tc360950: @SkalskiP I think it makes it much clearer what's going on. I've added this change.

LinasKo:
Hi @tc360950,

What's the best way to test this? Is there a Colab we can use?

supervision/detection/line_zone.py
```diff
@@ -158,28 +158,28 @@ def trigger(self, detections: Detections) -> Tuple[np.ndarray, np.ndarray]:
             ]
         )
 
+        cross_products_1 = self._cross_product(all_anchors, self.limits[0])
+        cross_products_2 = self._cross_product(all_anchors, self.limits[1])
+        # anchor is in limits if it's on the same side of both limit vectors
+        in_limits = ~np.logical_xor(cross_products_1 > 0, cross_products_2 > 0)
+        # Reduce array to find out if all anchors for a detection are within limits
+        in_limits = np.min(in_limits, axis=0)
```
LinasKo, line 166: It'd be much cleaner with `np.all`. Then, the comment above can be removed.

supervision/detection/line_zone.py
```diff
@@ -158,28 +158,28 @@ def trigger(self, detections: Detections) -> Tuple[np.ndarray, np.ndarray]:
             ]
         )
 
+        cross_products_1 = self._cross_product(all_anchors, self.limits[0])
+        cross_products_2 = self._cross_product(all_anchors, self.limits[1])
+        # anchor is in limits if it's on the same side of both limit vectors
+        in_limits = ~np.logical_xor(cross_products_1 > 0, cross_products_2 > 0)
```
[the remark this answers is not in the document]
LinasKo: With our standards, I'd err on the side to remove the comments.

@SkalskiP, vectorized code like this is the perfect example where I'd prefer to keep the comments. It's efficient, splitting into functions makes it messier, but without having seen cross product before another contributor may not know what's happening.

supervision/detection/line_zone.py
```diff
@@ -158,28 +158,28 @@ def trigger(self, detections: Detections) -> Tuple[np.ndarray, np.ndarray]:
...
                 continue
 
-            triggers = [
-                self.vector.cross_product(point=anchor) < 0 for anchor in box_anchors
-            ]
-
-            if len(set(triggers)) == 2:
+            if not is_uniformly_triggered[i]:
+                # One anchor lies to the left of the line
+                # whilst another lies to the right
```
LinasKo, line 179: uniformly_triggered is clear enough; let's remove the comment

supervision/detection/line_zone.py
```diff
@@ -158,28 +158,28 @@ def trigger(self, detections: Detections) -> Tuple[np.ndarray, np.ndarray]:
...
         )
 
+        cross_products_1 = self._cross_product(all_anchors, self.limits[0])
+        cross_products_2 = self._cross_product(all_anchors, self.limits[1])
+        # anchor is in limits if it's on the same side of both limit vectors
+        in_limits = ~np.logical_xor(cross_products_1 > 0, cross_products_2 > 0)
+        # Reduce array to find out if all anchors for a detection are within limits
+        in_limits = np.min(in_limits, axis=0)
+
+        # Calculate which anchors lie to the left of the line
```
LinasKo, line 168: Let's remove the comment; we're being fairly explicit that we care about one side later on.

supervision/detection/line_zone.py
```diff
@@ -197,6 +197,23 @@ def trigger(self, detections: Detections) -> Tuple[np.ndarray, np.ndarray]:
 
         return crossed_in, crossed_out
 
+    @staticmethod
+    def _cross_product(anchors: np.ndarray, vector: Vector) -> np.ndarray:
```
LinasKo, line 201: We prefer methods like this to live in `utils.py`.

Please move it there and remove leading underscore `_`.

LinasKo (changes requested):

LinasKo:
Hey @tc360950,

I've submitted some review comments. The changes are very small - if you can finish by Tuesday, there's a very high chance we'd include this in the next release!

supervision/detection/line_zone.py
tc360950: done

supervision/detection/line_zone.py
tc360950: done

supervision/detection/line_zone.py
tc360950: done

supervision/detection/line_zone.py
tc360950: done

supervision/detection/line_zone.py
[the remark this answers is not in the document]
tc360950: done

tc360950:
> I did some basic verification with this code: https://colab.research.google.com/drive/1rA0wXfaPlGnGKzn6wVA_x01jcX9Dr8Gu?usp=sharing
> 
> I see no regressions, after testing with `vehicles` and `people-walking` videos. I ran the updated tests from #1206 and it behaves identically, passing the current tests and failing on V-shape motions.
> 
> @tc360950, if you have a better test, let us know 😉

I actually went a little further - I've pasted the vectorized version into a notebook and compared results to the old implementation after every frame: 
https://colab.research.google.com/drive/14tPYRcgFo9-WZSppEt7pDB_djRmDQHqG?usp=sharing

LinasKo:
This is awesome. A very high quality contribution.
I've had a look at the colab - it's indeed a very thorough test.

Thank you very much!

LinasKo (approved):

Issue #1209, opened 2024-05-20
ValueError when loading COCO dataset with multiple segmentation masks for one class 


DancinParrot:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar bug report.

### Bug

My current dataset and project requires annotating a subject with more than 1 segmentation masks of the same class. A rough analogy is as follows whereby the hands of a person are segmented and assigned the same class of "hand":

![image](https://github.com/roboflow/supervision/assets/63692048/b2a449b3-7f45-47b4-848d-adf3dd92dae9)

As a result, when the COCO dataset is loaded into my program using supervision, the program crashes with the following error:
```
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (7,) + inhomogeneous part.
```

After some research, I discovered Ultralytics' [JSON2YOLO](https://github.com/ultralytics/JSON2YOLO) repository on GitHub, and adapted the library's `merge_multi_segment()` (seen [here](https://github.com/ultralytics/JSON2YOLO/blob/e6dd7784b77db4f1d8c3aa95ed93de2890b3ac23/general_json2yolo.py#L330)) function in supervision's `coco.py` file which then allows the COCO dataset to be loaded.

### Environment

- Supervision: 0.20.0
- OS: OpenSUSE Tumbleweed 20240423
- Python: 3.12.2

### Minimal Reproducible Example

The following code is used to load the COCO dataset with the annotations_path being the path to a .json file containing the paths and annotations for all images in the dataset:
```Python
ds = sv.DetectionDataset.from_coco(
        images_directory_path=images_directory_path,
        annotations_path=annotations_path,
        force_masks=True,
    )
```

The following is an example of a class/category containing multiple segmentation masks:
```JSON
{
 ...
    {
      "id": 41,
      "image_id": 6,
      "category_id": 0,
      "bbox": [
        694.801517364719,
        278.90263698033465,
        161.52883212628387,
        282.881946369456
      ],
      "segmentation": [
        [
          694,
          560.5,
          764.5,
          407,
          759.5,
          400,
          765.5,
          397,
          765.5,
          393,
          760.5,
          391,
          759.5,
          384,
          754.5,
          381,
          763,
          376.5,
          767.5,
          370,
          764.5,
          363,
          768.5,
          354,
          735,
          278.5,
          741.5,
          284,
          776,
          356.5,
          782,
          359.5,
[...]

PR #1229, opened 2024-05-25
feat: Merge Multiple Polygons in a COCO Dataset Annotation

DancinParrot:
# Description

This PR attempts to address issue #1209. Within an appropriately structured COCO dataset, an annotation may contain a segmentation list with multiple nested lists of vertices. This results in supervision terminating prematurely and raising a ValueError.

The solution as discussed in the issue includes the creation of binary masks for all polygons within the segmentation list and the usage of numpy's logical_or() function to merge all masks which is used in the `coco_annotations_to_masks` function.

## Type of change

The change has been tested using a custom COCO dataset which produces the ValueError mentioned in the issue linked above. You may test out the changes by following the guide on the provided Jupyter Notebook using the following resources:
- Google Colab Jupyter Notebook: https://colab.research.google.com/drive/1P57MlFAypF2pvrwIxPQl2hIU8mxoFe7T?usp=sharing
- Custom COCO dataset: https://drive.google.com/file/d/1GHA4z8QK07efOu9Is3v__Dh4V0D483sB/view?usp=sharing

Documentation may require an update.

## Docs

-   [x] Docs updated? What were the changes: Created an entry under `docs/datasets/utils.md` with description and types of the function and its parameters. An example and type have also been added for its output.


supervision/dataset/utils.py
```diff
@@ -46,6 +47,59 @@ def approximate_mask_with_polygons(
     ]
 
 
+def merge_polygons(
```
SkalskiP, line 50: I think the `merge_polygons` name is a bit misleading and not conststant with outhe util names in supervision. I think we should call it `polygons_to_mask`.

And of course we already have a function called `polygon_to_mask` that does the same but only take one mask. 

It seems a bit silly that we would have 2 functions doeng almost the same. I thin kwe should:
- rename `merge_polygons` to `polygons_to_mask`.
- make sure that `polygons_to_mask` is not using `polygon_to_mask` internally.
- modify places where `polygon_to_mask` was used and use `polygons_to_mask` instead.
- mask `polygon_to_mask` as deprecated and recommend using `polygons_to_mask` instead.

SkalskiP (changes requested):

supervision/dataset/utils.py
SkalskiP: @DancinParrot, here is an example of how we mark functions as deprecated. We use `deprecated` decorator to do it.

```
@deprecated(
    "`Detections.from_roboflow` is deprecated and will be removed in "
    "`supervision-0.22.0`. Use `Detections.from_inference` instead."
)
```

SkalskiP:
Hi @DancinParrot 👋🏻 Thanks a lot for your contribution 🙏🏻 This is a valid change. As I try to keep the code clean, I left a comment proposing a rename to `polygons_to_mask` and deprecation of the `polygon_to_mask` function. What do you think?

supervision/dataset/utils.py
SkalskiP: Please use `supervision-0.27.0` as a deprecation version.

DancinParrot:
Hi @SkalskiP ! Thanks for reviewing and providing feedback on my PR!

> As I try to keep the code clean, I left a comment proposing a rename to `polygons_to_mask` and deprecation of the `polygon_to_mask` function. What do you think?

Yup, I think that's a great idea. It'd certainly be a more accurate representation of the function's intended purpose. Since multiple lists of vertices within the "segmentation" field (which in itself is a list) of a COCO dataset annotation is perfectly acceptable, the function should be capable of handling both one and multiple lists without relying on yet another function that introduces another unnecessary layer of abstraction.

I'll proceed with implementing the suggested changes and deprecating the current `polygon_to_mask()` function in favour of `polygons_to_mask()`.

Thanks again for your feedback, appreciate it!

DancinParrot:
Hi @SkalskiP ! During refactoring, I noticed a very similar function `_polygons_to_mask()` (see [here](https://github.com/roboflow/supervision/blob/2723735b018cf6f8907d4c706b5f759ff641a53d/supervision/dataset/formats/yolo.py#L43)). The function follows a similar methadology in calling `polygon_to_mask()` for each polygon in the input list, though it does not require the merging of masks into one. In this case, given the similarity of the names with a slight difference in terms of functionality, would the introduction of `polygons_to_mask()` cause confusion?

I suppose we could refactor it in such a way but it does seem to be a bit odd:
```Python
def _polygons_to_masks(
    polygons: List[np.ndarray], resolution_wh: Tuple[int, int]
) -> np.ndarray:
    return np.array(
        [
            polygons_to_mask(polygon=[polygon], resolution_wh=resolution_wh)
            for polygon in polygons
        ],
        dtype=bool,
    )
```

DancinParrot:
Hi @SkalskiP ! Any updates on the above?

DancinParrot:
Hi @SkalskiP, I noticed the issue still remains unfixed. Is this PR still valid?

Abdul-Mukit:
I also need this solution. Can't load the valid datasets without it.

Abdul-Mukit:
Hi, @SkalskiP. We would appreciate it if you could comment whether `polygons_to_mask` vs `_polygons_to_mask` creates any confusion, as asked by @DancinParrot.
https://github.com/roboflow/supervision/pull/1229#issuecomment-2133661718

Hi, @DancinParrot. If SkalskiP can't reply, would you have time to implement his original suggestion of `polygons_to_mask` anyway? I personally think that the name is intuitive and clear. Once reviews are addressed, we can request him again.

DancinParrot:
> Hi, @DancinParrot. If SkalskiP can't reply, would you have time to implement his original suggestion of `polygons_to_mask` anyway? I personally think that the name is intuitive and clear. Once reviews are addressed, we can request him again.

Sure thing, I'll implement the requested changes while waiting for his reply. Since this is a rather old issue, do give me some time to work on it. Not sure if they are any conflicts with all the commits made thus far.

DancinParrot:
Hi @SkalskiP, I have addressed the requested changes and updated my branch with the latest commits, please review it again and let me know if there's any issues. I'm not sure what version to use for deprecation so I decided to stick with the initial one given by you as a placeholder, `supervision-0.27.0`. Changes were tested using the Google Colab notebook and unit tests are all green.

@Abdul-Mukit for your reference as well.

Borda:
@DancinParrot thank you for your effort, but this was already shipped in #1086 🦝

PR #1237, opened 2024-05-28
Dynamically select Detections member variables

LinasKo:
# Description

When checking if both fields are `None` or defined in two detection to be merged, we'd hardcode the fields.

This change allows getting the programmatically.
Note that it requires an instance of the class.

## Type of change

Ran locally, added and ran very basic tests.
Colab: https://colab.research.google.com/drive/193L8OHsEsdhQEenDNwKbGdtPyE68cUbW?usp=sharing

None

## Docs

### DIFF
--- a/supervision/utils/internal.py
+++ b/supervision/utils/internal.py
@@ -1,7 +1,8 @@
 import functools
+import inspect
 import os
 import warnings
-from typing import Callable
+from typing import Any, Callable, Set
 
 
 class SupervisionWarnings(Warning):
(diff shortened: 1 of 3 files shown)


supervision/utils/internal.py
```diff
@@ -141,3 +142,42 @@ def __get__(self, owner_self: object, owner_cls: type) -> object:
         The result of calling the function stored in 'fget' with 'owner_cls'.
         """
         return self.fget(owner_cls)
+
+
+def get_instance_variables(cls: Any, include_properties=False) -> Set[str]:
```
LinasKo, line 147: I called it `get_instance_variables` to emphasize that it works better with instance, but currently it can also accept the class itself.

Should we even allow the latter?

test/utils/test_internal.py
LinasKo, line 1: This is likely insufficient.

supervision/utils/internal.py
LinasKo, line 1: This is nicely encapsulated, but is it too complicated for us?
Getting rid of passing in class only removes one `if` check (not the contents of `if`).

LinasKo: P.S.

ugh, the pre-commit hook made the set init ugly

test/utils/test_internal.py
```diff
@@ -0,0 +1,85 @@
...
+import pytest
+
+from supervision.detection.core import Detections
+from supervision.utils.internal import get_instance_variables
+
+
+@pytest.mark.parametrize(
+    "input_obj, include_properties, expected, exception",
+    [
+        (
```
SkalskiP, line 13: Make that case raise Exception.

supervision/utils/internal.py
SkalskiP, line 147: Check if you get class or instance of class. And raise with some descriptive message if it is a class.

SkalskiP, line 147: Add `include_protected` and maaaaaaybe `include_private` arguments.

LinasKo: Okay, I've just learned something. I knew that prepending `__` scopes the var with the class name. I expected to see `__MockClass__private_var`.

No! It's actually `_MockClass__private_var`! With a single underscore prefix!

I won't add neither `include_protected` nor `include_private`, as it adds a bit more complication than I like. If anyone needs it, they can add it themselves, and now there will be mock classes in the tests that will reveal the issue.

LinasKo:
Hey @SkalskiP,

The changes to allow `_protected` variables introduce more complication than I'd like.

I'd like to merge as is. I can't review my own changes, so can't do it myself.

SkalskiP (approved):

PR #1256, opened 2024-06-04
Fix/detections merge only filter out empty

LinasKo:
# Description

`Detections.merge` creates an invalid `Detections` object when the following are true:
1. Detections 1 have `confidence=None` or `class_id=None`, and is of length > 0.
2. Detections 2 are created with `Detections.empty()`, so have `confidence=np.empty` and `class_id=np.empty`. 

This results in a detection of mixed lengths: `len(xyxy) > 0`, `len(confidence) == 0`.

---

This proposed solution:
1. Adds one test case showing the error
2. Adds a filter discarding empty detections at the start of Detections.merge.
3. Does NOT revert the flexible merge rules introduced in https://github.com/roboflow/supervision/pull/1177

## Type of change

1 Test case + Colab: https://colab.research.google.com/drive/1KWsVmbS6pCPdy5yPyHtwiJfGkf7NX5g7?usp=sharing

Alternative solutions:
1. Remove `class_id` and `confidence` from `Detections.empty`
2. `is_empty` can be simplified if we removed the data set in `from_inference`
3. Alternative PR: also revert flexible merge: #1255 
    * If we clean up `Detections.empty`, then reverting flexible merge wouldn't be needed.

### DIFF
--- a/supervision/detection/core.py
+++ b/supervision/detection/core.py
@@ -928,6 +928,12 @@ class Detections:
             array([0.1, 0.2, 0.3])
             ```
         """
+        detections_list = [
+            detections
+            for detections in detections_list
+            if detections != Detections.empty()
+        ]
+
         if len(detections_list) == 0:
             return Detections.empty()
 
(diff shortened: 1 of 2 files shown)


LinasKo:
Following our discussions, we're keeping #1255

PR #1277, opened 2024-06-12
Patch RichLabelAnnotator, fix in-place functionality of other annotators

LinasKo:
# Description

This is a 3-in-1:
1. Addressed faults with RichLabelAnnotator left after #1116 
  * Primarily: it would not convert back to `np.ndarray` after annotation
2. Realized many annotators do nothing when used in-place.
  * Description on [Loom](https://www.loom.com/share/91308392c7de44ce9ad115778bc99c42?sid=8279b66f-4786-42de-9e77-58cd7381f6be)
  * Fixed in [Colab](https://colab.research.google.com/drive/1NvVs28MnSb-_W-zkFa6-FXiY-0G4nx7k#scrollTo=dVZIKNV1uXVl)
3. Color Annotator would fill boxes with solid color if used in-place, with numpy.

## Type of change

Ran every annotator except for TraceAnnotator.
Ran tests.
Colab: https://colab.research.google.com/drive/1NvVs28MnSb-_W-zkFa6-FXiY-0G4nx7k#scrollTo=dVZIKNV1uXVl

## Docs

### DIFF
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -1147,9 +1147,9 @@ class RichLabelAnnotator(BaseAnnotator):
                 self.font = ImageFont.truetype(font_path, font_size)
             except OSError:
                 print(f"Font path '{font_path}' not found. Using PIL's default font.")
-                self.font = ImageFont.load_default(size=font_size)
+                self.font = ImageFont.load_default()
         else:
-            self.font = ImageFont.load_default(size=font_size)
+            self.font = ImageFont.load_default()
 
     @convert_for_rich_text_annotation
     def annotate(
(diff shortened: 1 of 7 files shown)


SkalskiP:
@LinasKo I see a lot of `copy to` and `copy` operations. Have you measured how those impact performance?

SkalskiP (changes requested):

LinasKo:
No. I'll do that, will let you know the results.

Realistically, it should be equivalent to frame.copy().

LinasKo:
1. Renamed to `ensure_cv2_image_...` and `ensure_pil_image_...`. I've seen this before, and we're not necessarily doing a conversion.
2. Removed a few copy operations. I finally understand why `cv2` has the destination parameter in methods like `addWeighted`, despite returning an image :)

The Colab now has tests in the end, comparing the performance on a 10,000x10,000 size empty image.
Results should show performance dips if any copying is expensive.

With the latest version, performance is equivalent or slightly better than in base supervision.

LinasKo:
Hey @onuralpszr ,

Given you've worked on it before, would you mind reviewing these changes too? There's a colab that shows how it works, only requiring that you upload the font file.

onuralpszr:
@LinasKo start reviewing

SkalskiP (approved):

Issue #1285, opened 2024-06-16
Modified Triangle Annotator


dsaha21:
### Search before asking

- [x] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Description

Upgradation of the triangle above head annotators more prominently

![Modified__traiangle_2](https://github.com/roboflow/supervision/assets/73748916/cb7e9cb7-a540-4f43-bb67-1e38d89a5175)
![Modified_Triangle_1](https://github.com/roboflow/supervision/assets/73748916/c6bcf82f-1edc-4c0b-94a8-bc560095b438)

### Use case

The original Triangle Annotator is attached below :-
![Original](https://github.com/roboflow/supervision/assets/73748916/e4c1f08d-7bba-49c0-93a4-4d69596d5200)

The modified one might help us to view for a more clear detection.

### Additional

My request is to please analyse the modified annotator. If its pleasing and if possible please accept for a PR. Thank You

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

LinasKo:
Hi @dsaha21 👋 

I like how this looks! Feel free to open a PR for it. Please also include a Colab we can test it in.

Let's make sure the old behaviour is the default, (e.g. if `outline_thickness=0` then old triangles are shown).

Are the any [other annotators](https://supervision.roboflow.com/annotators/) we should add this to?

LinasKo:
This looks great. Let's make a PR :)

One change I'd like to request is to check if this can also be added to sv.DotAnnotator, as it can be just as hard to see: [Colab](https://colab.research.google.com/drive/1NvVs28MnSb-_W-zkFa6-FXiY-0G4nx7k#scrollTo=VjdCk7cMuhh2).

onuralpszr:
Complete via https://github.com/roboflow/supervision/pull/1294

PR #1294, opened 2024-06-19
Modified triangle and dot in annotators-core 

dsaha21:
# Description

Adding changes as described in the [ISSUE](https://github.com/roboflow/supervision/issues/1285)

The modifications upon the existing code base might result for a more clear visualisation 

# Testing
Also adding the [colab](https://colab.research.google.com/drive/1vHUSwxYtMz_wbp7IH18kFlZBD1UZdeYV#scrollTo=Gk2KhyBkw6h6) file for checking and testing.

Thanks.

### DIFF
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -901,8 +901,13 @@ class DotAnnotator(BaseAnnotator):
                 else custom_color_lookup,
             )
             center = (int(xy[detection_idx, 0]), int(xy[detection_idx, 1]))
-            cv2.circle(scene, center, self.radius, (0, 0, 0), self.outline_thickness)
-            cv2.circle(scene, center, self.radius, color.as_bgr(), -1)
+            
+            if self.outline_thickness:
+                cv2.circle(scene, center, self.radius, (0,0,0), self.outline_thickness)
+                cv2.circle(scene, center, self.radius, color.as_bgr(), -1)
+            else:
+                cv2.circle(scene, center, self.radius, color.as_bgr(), -1)
+        
         return scene
 
 
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -1696,7 +1696,7 @@ class TriangleAnnotator(BaseAnnotator):
 
                 cv2.fillPoly(scene, [vertices], color.as_bgr())
                 cv2.polylines(scene, [vertices], True, (0,0,0), thickness=self.outline_thickness)
-            
+
         return scene
 
 
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -1695,7 +1695,9 @@ class TriangleAnnotator(BaseAnnotator):
             )
 
             cv2.fillPoly(scene, [vertices], color.as_bgr())
-            cv2.polylines(scene, [vertices], True, (0,0,0), thickness=self.outline_thickness)
+            cv2.polylines(
+                scene, [vertices], True, (0, 0, 0), thickness=self.outline_thickness
+            )
 
         return scene
 
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -901,13 +901,15 @@ class DotAnnotator(BaseAnnotator):
                 else custom_color_lookup,
             )
             center = (int(xy[detection_idx, 0]), int(xy[detection_idx, 1]))
-            
+
             if self.outline_thickness:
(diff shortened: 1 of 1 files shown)


dsaha21:
I have read the CLA Document and I sign the CLA

supervision/annotators/core.py
```diff
@@ -830,9 +830,10 @@ class DotAnnotator(BaseAnnotator):
     def __init__(
         self,
         color: Union[Color, ColorPalette] = ColorPalette.DEFAULT,
-        radius: int = 4,
+        radius: int = 5,
```
LinasKo, line 833: Let's leave the default value at `4`, even if it is hard to see.
People may already be using this in production systems.

supervision/annotators/core.py
```diff
@@ -847,6 +848,7 @@ def __init__(
         self.radius: int = radius
         self.position: Position = position
         self.color_lookup: ColorLookup = color_lookup
+        self.outer_thickness = outer_thickness
```
LinasKo, line 851: Let's keep the naming the same as much as possible. `outline_thickness` sounds correct here too.

supervision/annotators/core.py
```diff
@@ -1669,27 +1674,80 @@ def annotate(
...
-            tip_x, tip_y = int(xy[detection_idx, 0]), int(xy[detection_idx, 1])
-            vertices = np.array(
-                [
-                    [tip_x - self.base // 2, tip_y - self.height],
-                    [tip_x + self.base // 2, tip_y - self.height],
-                    [tip_x, tip_y],
-                ],
-                np.int32,
-            )
+        if self.outline_thickness:
```
LinasKo, line 1677: This is a lot of logic for a small change.

A more efficient method involve using `cv2.polylines`.
There you can specify thickness and draw another shape behind the original triangle.

I wouldn't expect to have two very different if-else branches.

supervision/annotators/core.py
```diff
@@ -899,6 +901,7 @@ def annotate(
             )
             center = (int(xy[detection_idx, 0]), int(xy[detection_idx, 1]))
             cv2.circle(scene, center, self.radius, color.as_bgr(), -1)
+            cv2.circle(scene, center, self.radius, (0, 0, 0), self.outer_thickness)
```
LinasKo, line 904: I'd expect the background to be drawn first. I'd swap these 2 `cv2.circle` lines.

LinasKo (changes requested):
Good starting point - I like the testing colab and the resulting annotation, but there's way to improve the code. I left a few comments 😉 

I'll review again after that's done.

supervision/annotators/core.py
dsaha21: Cool, will keep it same.

supervision/annotators/core.py
dsaha21: Actually the original triangles are little small :)
No problem, I will add cv2.polygons and the thickness parameter.

onuralpszr:
@LinasKo I added doc part quickly, and rest looks better but @dsaha21  can you squash your commits and make meaningful comment message, or I can do it too (I am gonna squash couple of them and write something cleaner)

supervision/annotators/core.py
```diff
@@ -898,6 +901,7 @@ def annotate(
                 else custom_color_lookup,
             )
             center = (int(xy[detection_idx, 0]), int(xy[detection_idx, 1]))
+            cv2.circle(scene, center, self.radius, (0, 0, 0), self.outline_thickness)
```
LinasKo, line 904: Let's only do this when `self.outline_thickness` is not 0, just like in the triangle annotator

LinasKo (approved):

supervision/annotators/core.py
dsaha21: Yes @LinasKo, will add a condition statement here. Updating it and submiting.

dsaha21:
Hi @LinasKo, please check the latest [colab notebook](https://colab.research.google.com/drive/1vHUSwxYtMz_wbp7IH18kFlZBD1UZdeYV#scrollTo=Gk2KhyBkw6h6)

These are the final changes. Please let me know about further changes if needed. Thanks

LinasKo:
Thank you, @dsaha21 and @onuralpszr!

It's great to have such a nice contribution in the repo this quick 😉

PR #1300, opened 2024-06-21
detectron2 mask convert to detection

kirilman:
# Description
Copying segmentation masks from detectron2 predictions to the detection class. The resulting masks can be used for displaying with sv.MaskAnnotator(). 

## Type of change

The code is executed:
detections = sv.Detections.from_detectron2(p_det)
bounding_box_annotator = sv.MaskAnnotator()
annotated_frame = bounding_box_annotator.annotate(
    scene=img.copy(),
    detections=detections,
)

### DIFF
--- a/supervision/detection/core.py
+++ b/supervision/detection/core.py
@@ -540,7 +540,9 @@ class Detections:
         return cls(
             xyxy=detectron2_results["instances"].pred_boxes.tensor.cpu().numpy(),
             confidence=detectron2_results["instances"].scores.cpu().numpy(),
-            mask=detectron2_results["instances"].pred_masks.detach().cpu().numpy(),
+            mask=detectron2_results["instances"].pred_masks.cpu().numpy()
+            if hasattr(detectron2_results["instances"], "pred_masks")
+            else None,
             class_id=detectron2_results["instances"]
             .pred_classes.cpu()
             .numpy()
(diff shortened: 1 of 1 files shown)


LinasKo:
Hi @kirilman 👋 

Does Detectron 2 produce masks every time? Are there cases when no mask would be produced?

onuralpszr:
@LinasKo  Test collab with results 

https://colab.research.google.com/drive/1ich6f2UBOYXAV4AuKRt2mDi4trzXfs0w?usp=sharing

With and without mask result and original vis result, If all good I am merging

LinasKo (approved):
Indentation looked peculiar, but it seems the pre-commit formats it this way.

LinasKo:
Hi @kirilman 👋 

This is a great contribution, and together with @onuralpszr's change, we'd really like to use it in our repository!

However, before we do, you need to sign the [Contributor License Agreement](https://cla-assistant.io/roboflow/supervision?pullRequest=1300). Can you please follow the link and complete it?

Thank you!

kirilman:
> Hi @kirilman 👋
> 
> Does Detectron 2 produce masks every time? Are there cases when no mask would be produced?
Yes, you're right. We need to think about the case of no mask

onuralpszr:
> > @LinasKo  👋
> > Does Detectron 2 produce masks every time? Are there cases when no mask would be produced?
> 
> Yes, you're right. We need to think about the case of no mask

I already cover that part in your PR branch could you please accept CLA so we can merge :)

LinasKo:
Indeed, the CLA allows us to include your code, @kirilman

onuralpszr:
@kirilman thank you for validation 

@LinasKo merge complete :)

LinasKo:
Thank you @kirilman , @onuralpszr !

onuralpszr:
> Thank you @kirilman, @onuralpszr!

You're welcome, Pleasure ❤️

PR #1307, opened 2024-06-25
Suppress mkdocs relative link INFO messages

LinasKo:
# Description

`mkdocs serve` would output a wall of `INFO` messages about relative links. However, we explicitly opted to use relative links. This PR adds configs suppressing the messages.

## Type of change

`mkdocs serve` - checked messages during build, then checked the docs site for regressions.

## Docs

Only the build process is affected


onuralpszr (approved):
@LinasKo thank you, merging in.

Issue #1274, opened 2024-06-11
[KeyPoints] - extend `from_mediapipe` with Google MediaPipe FaceMesh


LinasKo:
# Description

Much like #1174 and #1232 adding pose landmark support, we'd also like to add face detection support to the `from_mediapipe` method.

* Add `Skeleton.FACEMESH_TESSELETION` of size `468` to the [Skeleton](https://github.com/roboflow/supervision/blob/447ef41fc45353130ec4dccdc7eeaf68b622fb7e/supervision/keypoint/skeletons.py#L7) enum.
  * The nodes can be found here: https://github.com/google-ai-edge/mediapipe/blob/8cb99f934073572ce73912bb402a94f1875e420a/mediapipe/python/solutions/face_mesh_connections.py#L74
  * Docs can be found here: https://github.com/google-ai-edge/mediapipe/blob/master/docs/solutions/face_mesh.md
* Add the code to the `from_mediapipe` function in [`KeyPoints`](https://github.com/roboflow/supervision/blob/447ef41fc45353130ec4dccdc7eeaf68b622fb7e/supervision/keypoint/core.py#L16) object that is introduced in #1232.
* We'd like to support responses from both legacy and modern way to call the face mesher - see links below.

![facemesh](https://github.com/roboflow/supervision/assets/6500785/59433e66-74e0-448c-b902-4f19947d379e)

# Links:
- Google Mediapipe repository: https://github.com/google/mediapipe
- Google Mediapipe face landmarker: https://ai.google.dev/edge/mediapipe/solutions/vision/face_landmarker
  - Python Guide (Modern): https://ai.google.dev/edge/mediapipe/solutions/vision/face_landmarker/python
  - Legacy: https://colab.research.google.com/github/googlesamples/mediapipe/blob/main/examples/face_landmarker/python/%5BMediaPipe_Python_Tasks%5D_Face_Landmarker.ipynb
- Skeletons: https://github.com/google-ai-edge/mediapipe/blob/8cb99f934073572ce73912bb402a94f1875e420a/mediapipe/python/solutions/face_mesh_connections.py#L74

# Additional

- Note: Please share a Google Colab with minimal code to test the new feature. We know it's additional work, but it will speed up the review process. The reviewer must test each change. Setting up a local environment to do this is time-consuming. Please ensure that Google Colab can be accessed without any issues (make it public). Thank you! 🙏🏻

LinasKo:
Hey @David-rn,

[...]

PR #1316, opened 2024-06-28
Facemesh from mediapipe

David-rn:
# Description

According to issue #1274, this PR extends the usage of `from_mediapipe` function to support FaceMesh.

## Type of change

The following [Google Colab](https://colab.research.google.com/drive/1iltp_wxLdwRBvTd3PXSJZ5-j4jKDq5B5?usp=sharing) describes the process used for testing this new feature. It allows to upload a picture to test it. (This Colab feature only worked in Chrome for me).

The first part of the Colab is the examples from the PoseLandmaker to ensure that this functionality is still working after the new changes.

- As discussed in the issue thread, I created two Skeletons: `FACEMESH_TESSELATION_NO_IRIS` and `FACEMESH_TESSELATION` to support the legacy code with optional refinement of the iris. However, there is a minor difference between the skeletons described in the [mediapipe library](https://github.com/google-ai-edge/mediapipe/blob/8cb99f934073572ce73912bb402a94f1875e420a/mediapipe/python/solutions/face_mesh_connections.py#L47) and the ones in this PR. The skeletons represent the connections between each point on them, but there are 2 points out of the 478 that do not have a connection (the centre of the iris). Since the `EdgeAnnotator` class uses the SKELETONS_BY_VERTEX_COUNT variable for counting unique vertices in the skeleton, I connected each center to itself to have a unique vertex count of 478. This can be appreciate in the Google Colab notebook.

- The code was updated to look for the correct inference result using the `hasattr` method, which makes easier to integrate the `HandLandmarker` results in the future (if it is planned). 

- The docstring was updated to add a `FaceLandmarker` inference example. (The lines are probably too long but the pre-commit won't reformat)

## Docs


LinasKo (changes requested):
Hey @David-rn,

Excellent work as always! There's a few extra tests in the Colab that I'd like to see.

1. I see the section "Modern" at the end, but the code is missing; likely not saved. This is probably the copy of what's in the docstring + download of the landmarker task.
2.  Can we test with an empty image? `np.empty((640,640,3), dtype=np.uint8)` usually works.
3. I wonder, can this detect more than 1 face?

David-rn:
Hi @LinasKo, Thanks for the comments!

My bad!! I don't know why it was not saved. In the same link you can find now the saved version with the extra considerations:

1. Indeed it was not saved.
2. The last example is an empty (noise) image.
3. Yes it can, since it wasn't saved it didn't appear. It can handle more than one face, but `num_faces` attribute needs to be set.

onuralpszr:
@David-rn thanks for fixing collab (I was also curious why is it was missing, I thought you were working)  can you change commit sha in last commit for trigger pre-commit (There was a github server issue at the time but it is fixed so pre-commit should work fine again) 

Without any change you do this

```
git add --amend --no-edit 
git push -f
```

LinasKo (approved):

LinasKo:
I've tested it with multiple images, checked out how the docs look, added a bit of a change. I believe my suggestion in the colab to test on an empty image wasn't entirely correct - I meant `np.zeroes` rather than `np.empty`. Still, the result looks as I'd expect in both cases.

Thank you @David-rn (and @onuralpszr) - another flawless, well-described contribution. It's a pleasure working with you :)

Merging!

(fyi @SkalskiP, [skeletons.py](https://github.com/roboflow/supervision/pull/1316/files#diff-169ce07e6f3187b38ab77311914b2292ab39339b351b942ee10b70835372a0ae) now contains a variable which is a VERY long list of edges. I want the functionality in the repo now, but if it bothers us, we can export it to a json file later).

PR #1317, opened 2024-06-28
fix: 🐞 skip broken docutils package, revert numpy version lock

onuralpszr:
# Description

Poetry installation shows that docutils broken because sdist for 0.21 release is versioned "0.21.post1" but the directory inside it is "0.21". Therefore, it is not installable via poetry and cause issues

- https://github.com/python-poetry/poetry/issues/9293
- https://sourceforge.net/p/docutils/bugs/483/#f648 (fixed in version 0.21.1)

Secondly, I removed numpy version lock since opencv fix via

- https://github.com/opencv/opencv-python/pull/1002

Therefore let's merge this PR, I also tested via tox and manually created venvs before sent


LinasKo:
Tested it, built the docs, installed `python-opencv` explicitly - the numpy no longer causes issues.

I've seen the docutils issue before and failed to find a solution. Thanks a lot for tracking it down!

LinasKo (approved):

Issue #1355, opened 2024-07-13
Tracker IDs Skipping


likith1908:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Question

- While using the tracker, I noticed that the tracker IDs are not assigned sequentially or some numbers are being skipped.

- Videos Used: https://drive.google.com/drive/folders/14o1MStr357HPr0zl1uauMP6k4swOh63D?usp=sharing
This folder has input and output video files

- Below is my code

```
import argparse
import supervision as sv
import cv2
from ultralytics import YOLO
import numpy as np
from collections import defaultdict, deque

SOURCE = np.array([[0, 0], [3072, 0], [3072, 1080], [0, 1080]])

TARGET_WIDTH = 0.7388
TARGET_HEIGHT = 0.2594

TARGET = np.array([
    [0, 0],
    [TARGET_WIDTH, 0],
    [TARGET_WIDTH, TARGET_HEIGHT],
    [0, TARGET_HEIGHT]
])

class ViewTransformer:
    def __init__(self, source=np.ndarray, target=np.ndarray):
        source = source.astype(np.float32)
        target = target.astype(np.float32)
        self.m = cv2.getPerspectiveTransform(source, target)

    def transform_points(self, points: np.ndarray) -> np.ndarray:
        if points.size == 0:
            print("Warning: No points to transform.")
            return np.array([])
        reshaped_points = points.reshape(-1, 1, 2).astype(np.float32)
        transformed_points = cv2.perspectiveTransform(reshaped_points, self.m)
        return transformed_points.reshape(-1, 2)

def parse_arguments() -> argparse.Namespace:
    parser = argparse.ArgumentParser(
        description="Speed Estimation using Ultralytics and Supervision"
    )
    parser.add_argument(
        "--source_video_path",
        required=False,
        default="/input.mp4",
        help="Path to the source video file",
        type=str``
    )
    return parser.parse_args()

def main():
    args = parse_arguments()
    print(f"Source video path: {args.source_video_path}")
    
    video_info = sv.VideoInfo.from_video_path(args.source_video_path)
    print(f"Video info: {video_info}")
    
    model = YOLO("model.pt")
    print("YOLO model loaded.")
    
    byte_track = sv.ByteTrack(frame_rate=video_info.fps)
    print("ByteTrack initialized.")
    
[...]

Issue #1320, opened 2024-07-01
YOLOv8 + ByteTrack integration issues


ddrisco11:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Question

Hello! I'm currently building a program to detect deep sea creatures in submarine video. I am using YOLOv8 to make detections and ByteTrack to assign object IDs to these detections. My output includes both an annotated video (based exclusively on YOLO output) and a csv file of all distinct detections (determined as distinct by ByteTrack). I am having an issue where certain creatures are annotated in the video output, ie. detected by YOLO, but then they are omitted from the csv output ie. not assigned a tracking ID by ByteTrack. Please help! Thanks!

### Additional

def process_video(video_path: str, output_path: str, model_path: str, location_path: str, start_time: str, time_col: int, lat_col: int, lon_col: int, depth_col: int, salinity_col: int, oxygen_col: int, altitude_col: int,
                  confidence_threshold: float, iou_threshold: float, track_activation_threshold: float, minimum_matching_threshold: float, lost_track_buffer: int,
                  frame_rate: int, min_box_area: int, aspect_ratio_thresh: float):
    """Process the video to track objects and save tracking data."""
    model = YOLO(model_path)
    tracker = ByteTrack(
        track_activation_threshold=track_activation_threshold,
        minimum_matching_threshold=minimum_matching_threshold,
        lost_track_buffer=lost_track_buffer
    )
    location_data = get_location_data(location_path, time_col, lat_col, lon_col, depth_col, salinity_col, oxygen_col, altitude_col)
    start_time_seconds = time_to_seconds(start_time)

    cap = cv2.VideoCapture(video_path)
    width = int(cap.get(cv2.CAP_PROP_FRAME_WIDTH))
    height = int(cap.get(cv2.CAP_PROP_FRAME_HEIGHT))
    fps = cap.get(cv2.CAP_PROP_FPS)
    frame_count = int(cap.get(cv2.CAP_PROP_FRAME_COUNT))

    fourcc = cv2.VideoWriter_fourcc(*'mp4v')
    out = cv2.VideoWriter(output_path.replace('.csv', '.mp4'), fourcc, fps, (width, height))

    tracking_info = {}
    pbar = tqdm(total=frame_count, desc='Processing frames', leave=True, mininterval=10)

    frame_index = 0
    cached_boxes = None
    cached_labels = None

    try:
        while cap.isOpened():
[...]

PR #1364, opened 2024-07-15
Fix external track ids

rolson24:
# Description

This fix is for issue #1355. The tracker is skipping ID's for a few reasons. First off when the `minimum_consecutive_frames` is set to 1, each track gets assigned a new external id when it is created in `activate()` is called. This is an issue because not all of these tracks make it through the next frame where they get matched with a new detection and `update()` gets called. That means we are assigning an external track id to tracks that never are real tracks. I fixed this by getting rid of the external track id assignment in `activate()` (except for the first frame of the video) and instead just using the external id assignment in the `update()` function, which is how we do the assignment when `minimum_consecutive_frames > 1`.

The second reason the tracker is skipping ID's is because in the `removed_duplicate_tracks()` function, there is an overlap threshold to determine if two tracks are the same. In the current release this threshold is set at 0.15 which is relatively high because two tracks could be passing by each other and if they overlap more than 85% one would get removed. I lowered the threshold to 0.05 to reduce this.

## Type of change

This fix was tested with this [notebook](https://colab.research.google.com/drive/13NPdsDVagoqw3PdxH2URfyHN7YO-KClJ?usp=sharing)

Here you can see that in the old version external id 4 was skipped, but in the new version its not

Output from old version:
```
Current Frame: 38
New Track ids: set()
Track ids are consecutive: True
Current Frame: 39
New Track ids: set()
Track ids are consecutive: True
Current Frame: 40
New Track ids: set()
Track ids are consecutive: True
Current Frame: 41
New Track ids: {3, 5}
Track ids are consecutive: False
Current Frame: 42
New Track ids: set()
Track ids are consecutive: False
Current Frame: 43
New Track ids: set()
Track ids are consecutive: False
Current Frame: 44
New Track ids: set()
```
Image

Output from fixed version:

```
Current Frame: 38
New Track ids: set()
Track ids are consecutive: True
Current Frame: 39
New Track ids: set()
Track ids are consecutive: True
Current Frame: 40
New Track ids: set()
Track ids are consecutive: True
Current Frame: 41
New Track ids: {3, 4}
Track ids are consecutive: True
Current Frame: 42
New Track ids: set()
Track ids are consecutive: True
Current Frame: 43
New Track ids: set()
Track ids are consecutive: True
Current Frame: 44
New Track ids: set()
Track ids are consecutive: True
```
Image

## Docs


SkalskiP:
Hi @rolson24 👋🏻, we have two issues related to tracking that were submitted recently:
- https://github.com/roboflow/supervision/issues/1355
- https://github.com/roboflow/supervision/issues/1320

IS it related to both of them or only https://github.com/roboflow/supervision/issues/1355?

rolson24:
Yea it is also related to #1320, sorry about that.

supervision/tracker/byte_tracker/core.py
```diff
@@ -67,9 +67,11 @@ def activate(self, kalman_filter, frame_id):
         self.state = TrackState.Tracked
         if frame_id == 1:
             self.is_activated = True
+            if self.minimum_consecutive_frames == 1:
+                self.external_track_id = self.next_external_id()
 
-        if self.minimum_consecutive_frames == 1:
-            self.external_track_id = self.next_external_id()
+        # if self.minimum_consecutive_frames == 1:
```
SkalskiP, line 73: Do we aim to keep that comment here?

supervision/tracker/byte_tracker/core.py
```diff
@@ -206,8 +208,8 @@ class ByteTrack:
             reducing the likelihood of track fragmentation or disappearance caused
             by brief detection gaps.
         minimum_matching_threshold (float, optional): Threshold for matching tracks with detections.
-            Increasing minimum_matching_threshold improves accuracy but risks fragmentation.
-            Decreasing it improves completeness but risks false positives and drift.
+            Decreasing minimum_matching_threshold improves accuracy but risks fragmentation.
+            Increasing it improves completeness but risks false positives and drift.
```
SkalskiP, line 212: I'm not sure if I understand. How is increasing `minimum_matching_threshold` risks false positives?

SkalskiP (changes requested):

supervision/tracker/byte_tracker/core.py
rolson24: I don't think we actually need it. I will remove it.

PR #1385, opened 2024-07-19
Add BackgroundColorAnnotator

capjamesg:
# Description

This PR adds a new `sv.BackgroundColorAnnotator` that highlights all regions outside of a list of bounding boxes or masks.

## Type of change

You can test this feature using the following Colab:

https://colab.research.google.com/drive/18GLw4TR-au9DUmAeWU30M7BFJ_00BlxO?usp=sharing

N/A

## Docs

N/A

### DIFF
--- a/docs/detection/annotators.md
+++ b/docs/detection/annotators.md
@@ -450,7 +450,8 @@ status: new
 
     <div class="result" markdown>
 
-        ![background-overlay-annotator-example](https://media.roboflow.com/supervision-annotator-examples/background-color-annotator-example-purple.png)
+    ![background-overlay-annotator-example](https://media.roboflow.com/supervision-annotator-examples/background-color-annotator-example-purple.png)
+
     </div>
 
 <div class="md-typeset">
--- a/docs/detection/annotators.md
+++ b/docs/detection/annotators.md
@@ -433,6 +433,26 @@ status: new
 
     </div>
 
+=== "Background Color"
+
+    ```python
+    import supervision as sv
+
+    image = ...
+    detections = sv.Detections(...)
+
+    blur_annotator = sv.BackgroundColorAnnotator()
+    annotated_frame = blur_annotator.annotate(
+        scene=image.copy(),
+        detections=detections
+    )
+    ```
+
+    <div class="result" markdown>
+
+        ![background-color-annotator-example](https://media.roboflow.com/supervision-annotator-examples/background-color-annotator-example-purple.png)
+    </div>
+
 <div class="md-typeset">
     <h2><a href="#supervision.annotators.core.BoxAnnotator">BoxAnnotator</a></h2>
 </div>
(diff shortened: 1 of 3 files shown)


LinasKo (changes requested):

LinasKo:
I really like this one - it'd be a key component in a project idea I have.

@capjamesg, I've made some comments. Could you also add a default image to the Colab?
Something like `!wget https://media.roboflow.com/notebooks/examples/dog.jpeg` should do the trick.

I'll do a deeper code review later on, but it generally looks okay.

SkalskiP:
Hi, @capjamesg 👋🏻 would you have time to respond to the comments above?

LinasKo:
Hey @capjamesg,

Could we get an expanded Colab, please?

Here's the tests I'd like to do:
1. Test on image with many detections (you have this already)
2. Test with no detections (you can call `sv.Detections.empty()`
3. In-place call of the annotator:

```
# Instead of:
annotated_image = bounding_box_annotator.annotate(
    
# Call:
bounding_box_annotator.annotate(
    
# This should still annotate the image!
```

Also, since I've seen different behaviours recently, it would be nice to check if both in-place and not in-place work on PIL images. You can use `sv.cv2_to_pillow(image)` before passing to the model. (or just load with PIL)

capjamesg:
@LinasKo See my revisions to the Colab and code.

LinasKo:
Close, but to match the behaviour of annotators, it should work in-place for numpy images. I mean, it should modify the image you pass in: https://colab.research.google.com/drive/1SP4QyD_MgvWz9UI6qdkjAIaF74eRoUTy?usp=sharing

Can you modify and return the same image the annotator gets as an argument, instead of returning a copy with `Image.fromarray(colored_mask[:, :, ::-1])`? `np.copyto` may or may not be useful - I'm not sure.

Let me know if it gets too much - I'm a bit busy, but could help out 😉

SkalskiP:
@LinasKo / @capjamesg, what do you think about renaming it to `BackgroundFillAnnotator` or `BackgroundOverlayAnnotator`?

LinasKo:
Yesss, I like the new name much more

SkalskiP:
@LinasKo which one? :)

LinasKo:
I misread; I thought you mentioned the previous one too.

`BackgroundOverlayAnnotator`makes the most sense to me.

SkalskiP:
@LinasKo, is there any chance you could propagate that name change in this PR? We wouldn't need to ask @capjamesg to work on it once again.

LinasKo:
On it.

LinasKo:
Okay, I can take this one over.

The `return np.copyto` is incorrect, but we're a millimeter away from the solution 😆 

@capjamesg, thanks a lot! It's an awesome annotator to have in our toolbox.

LinasKo:
Changed the name, return type.

Tested here: https://colab.research.google.com/drive/1SP4QyD_MgvWz9UI6qdkjAIaF74eRoUTy?usp=sharing

LinasKo (approved):

SkalskiP (approved):

Issue #1384, opened 2024-07-19
[LabelAnnotator, RichLabelAnnotator, VertexLabelAnnotator] - allow `text_color` to be `Union[Color, ColorPalette]`


SkalskiP:
### Description

[`LabelAnnotator`](https://supervision.roboflow.com/develop/detection/annotators/#supervision.annotators.core.LabelAnnotator), [`RichLabelAnnotator`](https://supervision.roboflow.com/develop/detection/annotators/#supervision.annotators.core.RichLabelAnnotator), and [`VertexLabelAnnotator`](https://supervision.roboflow.com/develop/keypoint/annotators/#supervision.keypoint.annotators.EdgeAnnotator.annotate) currently allow defining the background color as `Union[Color, ColorPalette]`. When the background color is a single `Color`, all boxes are drawn with the same color. If the background color is a `ColorPalette`, the detection index, class_id, or tracker_id is mapped to the corresponding color from the palette. Unfortunately, this functionality is not currently available for `text_color`.

Update `LabelAnnotator`, `RichLabelAnnotator`, and `VertexLabelAnnotator` to enable defining `text_color` as `Union[Color, ColorPalette]` and implement the same mechanism as for the background color.

### Additional

- Note: Please share a Google Colab with minimal code to test the new feature. We know it's additional work, but it will speed up the review process. The reviewer must test each change. Setting up a local environment to do this is time-consuming. Please ensure that Google Colab can be accessed without any issues (make it public). Thank you! 🙏🏻

Bhavay-2001:
Hi @SkalskiP, you can assign this to me. I'll take a look over thr weekend.

SkalskiP:
@Bhavay-2001 Done ✅

SkalskiP:
The expected change was added to `LabelAnnotator` and `RichLabelAnnotator` via https://github.com/roboflow/supervision/pull/1387. @Bhavay-2001 would you be interested in adding this change to `VertexLabelAnnotator` as well?

Bhavay-2001:
Sure

SkalskiP:
@Bhavay-2001 thank you ;)

Issue #1383, opened 2024-07-19
[LabelAnnotator, RichLabelAnnotator, VertexLabelAnnotator] - add smart label positioning


SkalskiP:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Description

TODO

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

Bhavay-2001:
Hi @SkalskiP, could you pls describe the issue a little more and also can i take this issue?

SkalskiP:
Hi @Bhavay-2001 👋🏻 I recommend doing this first: https://github.com/roboflow/supervision/issues/1384

jeslinpjames:
Hey, @SkalskiP 

I'd like to try working on this. Could you provide any specific guidelines or tips for implementing this feature?

Bhavay-2001:
Hi @SkalskiP, could you please tell how to approach this?

LinasKo:
We're opening this up to the community! @jeslinpjames, it's been a long time - are you still interested? I'll leave this open for a few days on the off-chance you're still around.

LinasKo:
With respect to the implementation details, I'm glad to see Piotr's plan as I had the exact same idea, down to the connector line.

1. The mentioned annotators would have a new argument `use_smart_positioning`, activating this feature if set to `True`.
2. Let's treat the annotators as independent. If two different LabelAnnotators are used at once, let's allow their labels to overlap, even if this feature is on.
3. Unmentioned annotators that use labels can be ignored (LineZoneAnnotator, PolygonZoneAnnotator).
4. There are alternate approaches to this.

Let's start with the simplest one:
1. If two boxes intersect with another box, move it either along x or y, based on where the overlap is SMALLEST (this minimizes the distance we need to move).
2. Move proportionally to overlap size size, but with a cap. When everything moves at once, large movements may cause even greater overlaps. (similar to gradient descent!)
3. region boundaries should not allow labels to escape. They should reset the labels to the nearest possible position, at least along one axis. Even if the label moved or started out-of-bounds.

An upgraded version of this would take into account both x and y axes of the overlap and allow arbitrary motion direction. Implement this if you wish, but take care to minimize motion - overemphasize the motion along the SMALLER overlap direction.

An even more robust system uses some random noise to avoid stable states. We don't need this much detail 😉 

```python
find_smart_rectangle_positions(
[...]

PR #1387, opened 2024-07-20
Added Union[Color, ColorPalette] to text_color parameter

Bhavay-2001:
# Description

Fix of issue #1384 

## Type of change

I will update with colab notebook once the changes are approved.

## Docs

### DIFF
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -1023,7 +1023,8 @@ class LabelAnnotator(BaseAnnotator):
         Args:
             color (Union[Color, ColorPalette]): The color or color palette to use for
                 annotating the text background.
-            text_color (Union[Color, ColorPalette]): The color or color palette to use for the text.
+            text_color (Union[Color, ColorPalette]): The color or color palette to use 
+                for the text.
             text_scale (float): Font scale for the text.
             text_thickness (int): Thickness of the text characters.
             text_padding (int): Padding around the text within its background box.
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -1017,7 +1017,6 @@ class LabelAnnotator(BaseAnnotator):
         text_padding: int = 10,
         text_position: Position = Position.TOP_LEFT,
         color_lookup: ColorLookup = ColorLookup.CLASS,
-        text_color_lookup: ColorLookup = ColorLookup.CLASS,
         border_radius: int = 0,
     ):
         """
@@ -1033,8 +1032,6 @@ class LabelAnnotator(BaseAnnotator):
                 Possible values are defined in the `Position` enum.
             color_lookup (ColorLookup): Strategy for mapping colors to annotations.
                 Options are `INDEX`, `CLASS`, `TRACK`.
-            text_color_lookup (ColorLookup): Strategy for mapping text colors to
-                annotations. Options are `INDEX`, `CLASS`, `TRACK`.
             border_radius (int): The radius to apply round edges. If the selected
                 value is higher than the lower dimension, width or height, is clipped.
         """
@@ -1046,7 +1043,6 @@ class LabelAnnotator(BaseAnnotator):
         self.text_padding: int = text_padding
         self.text_anchor: Position = text_position
         self.color_lookup: ColorLookup = color_lookup
-        self.text_color_lookup: ColorLookup = text_color_lookup
 
     @ensure_cv2_image_for_annotation
     def annotate(
@@ -1124,7 +1120,7 @@ class LabelAnnotator(BaseAnnotator):
                 detections=detections,
                 detection_idx=detection_idx,
                 color_lookup=(
-                    self.text_color_lookup
+                    self.color_lookup
                     if custom_color_lookup is None
                     else custom_color_lookup
                 ),
@@ -1232,7 +1228,6 @@ class RichLabelAnnotator(BaseAnnotator):
         text_padding: int = 10,
         text_position: Position = Position.TOP_LEFT,
         color_lookup: ColorLookup = ColorLookup.CLASS,
-        text_color_lookup: ColorLookup = ColorLookup.CLASS,
         border_radius: int = 0,
     ):
         """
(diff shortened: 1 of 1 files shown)


yeldarby:
Nice! Probably also needs `custom_text_color_lookup` override

Bhavay-2001:
Hi @SkalskiP, I have made the changes and this is the [colab](https://colab.research.google.com/drive/1mZb1aYRM-ncKZwsCyKDzQuqABPuH--Mt?usp=sharing). Please let me know if anything is wrong and anything needs to be changed. 

I have only changed the Label Annotator for now and will change the other two based if this is correct.

Bhavay-2001:
Hi @SkalskiP, pls review this one. 
Thanks

supervision/annotators/core.py
```diff
@@ -1050,6 +1055,7 @@ def annotate(
         detections: Detections,
         labels: Optional[List[str]] = None,
         custom_color_lookup: Optional[np.ndarray] = None,
+        custom_text_color_lookup: Optional[np.ndarray] = None,
```
SkalskiP, line 1058: Let's use the same `custom_color_lookup`.

SkalskiP (changes requested):

SkalskiP:
Hi 👋🏻 @yeldarby and @Bhavay-2001. I don't think we need `custom_text_color_lookup`. 

The spirit behind `custom_color_lookup` is to override `color_lookup` while maintaining its consistent behavior. In the cases of `ColorLookup.INDEX`, `ColorLookup.CLASS`, or `ColorLookup.TRACK`, we utilize a single integer value to select a color with the same index from the `color` and `text_color` palettes. I believe the same approach should be adopted here. We should utilize `custom_color_lookup` to pick a color with the same index from the `color` and `text_color` palettes.

supervision/annotators/core.py
Bhavay-2001: ok

supervision/annotators/core.py
```diff
@@ -1369,7 +1382,7 @@ def annotate(
                 xy=(text_x, text_y),
                 text=text,
                 font=self.font,
-                fill=self.text_color.as_rgb(),
+                fill=text_color.as_rgb(),
```
Bhavay-2001, line 1385: In this, do we need to change `as_rgb()` to `as_bgr()` same like we did in LabelAnnotator?

Bhavay-2001:
Hi @SkalskiP, please review the changes.

Bhavay-2001:
Also, if this gets approved, can I start with #1383? I would like to know more about the issue.
Thanks

supervision/annotators/core.py
```diff
@@ -1011,37 +1011,42 @@ class LabelAnnotator(BaseAnnotator):
     def __init__(
         self,
         color: Union[Color, ColorPalette] = ColorPalette.DEFAULT,
-        text_color: Color = Color.WHITE,
+        text_color: Union[Color, ColorPalette] = ColorPalette.LEGACY,
```
onuralpszr, line 1014: Small suggestion:  I think we should keep white as default because it changes readability and also change default behavior, up until this version we put "white" and when people see different color It can create some confusion. It would be better keep "Color.WHITE" as default but keep type change is   "Union[Color, ColorPalette]" in case of multi color change.

Bhavay-2001: @onuralpszr, yess certainly. I changed this because it was overlapping with ColorPalette.DEFAULT and the labels weren't visible but I will change it to White now.

Also, couldn't understand this `but keep type change is "Union[Color, ColorPalette]" in case of multi color change.` Can you explain it further? Thanks

Bhavay-2001:
Hi @SkalskiP, this is open for reviewing now.

supervision/annotators/core.py
Bhavay-2001: I might be wrong on this but I think there is no such method. Although, there is `COLOR.WHITE`

onuralpszr: @Bhavay-2001  ah right I miss typed yes Color.WHITE

supervision/annotators/core.py
SkalskiP: `RichLabelAnnotator` uses Pillow as a rendering engine. Pillow accepts colors in RGB order, while OpenCV in BGR. So we keep `as_rgb()` here.

supervision/annotators/core.py
```diff
@@ -1011,37 +1011,42 @@ class LabelAnnotator(BaseAnnotator):
...
         self,
         color: Union[Color, ColorPalette] = ColorPalette.DEFAULT,
-        text_color: Color = Color.WHITE,
+        text_color: Union[Color, ColorPalette] = Color.WHITE,
         text_scale: float = 0.5,
         text_thickness: int = 1,
         text_padding: int = 10,
         text_position: Position = Position.TOP_LEFT,
         color_lookup: ColorLookup = ColorLookup.CLASS,
+        text_color_lookup: ColorLookup = ColorLookup.CLASS,
```
SkalskiP, line 1020: Let's remove this parameter and use `color_lookup`.

supervision/annotators/core.py
```diff
@@ -1210,19 +1226,20 @@ class RichLabelAnnotator(BaseAnnotator):
...
         self,
         color: Union[Color, ColorPalette] = ColorPalette.DEFAULT,
-        text_color: Color = Color.WHITE,
+        text_color: Union[Color, ColorPalette] = Color.WHITE,
         font_path: Optional[str] = None,
         font_size: int = 10,
         text_padding: int = 10,
         text_position: Position = Position.TOP_LEFT,
         color_lookup: ColorLookup = ColorLookup.CLASS,
+        text_color_lookup: ColorLookup = ColorLookup.CLASS,
```
SkalskiP, line 1235: Let's remove this parameter and use only `color_lookup`.

SkalskiP (changes requested):

SkalskiP:
@Bhavay-2001, thanks a lot for all the work. We are almost there. Just few small changes and we can merge.

Bhavay-2001:
> @Bhavay-2001, thanks a lot for all the work. We are almost there. Just few small changes and we can merge.

Done

SkalskiP:
Tested [here](https://colab.research.google.com/drive/1rwUTZAc24smf254NTAGRlYQFlzxW9MHq?usp=sharing) everything works. Merging!

Issue #1381, opened 2024-07-18
Add support for MMDetection tracking objects


rolson24:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Description

This is somewhat similar to #1308 in that it would be modifying `from_mmdetection()` to support `TrackDataSample`s. This would allow for easy integration with the MMTracking library, which has several trackers implemented in it, as well as integration with the Match and Segment Anything (MASA) model.

### Additional

I imagine it could be pretty simple and would only require checking if the input object has the `pred_track_instances` attribute, and then it could simply add the tracks with their track ID's into a `Detections` object.

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

SkalskiP:
Hi @rolson24 I just managed to plug in your MASA code into my football demo. I'd never be able to do it without your help. As for MASA: not bad, but I was hoping for a slightly better result. Would that allow us to plug in all [MMTracking](https://github.com/open-mmlab/mmtracking) models?

https://github.com/user-attachments/assets/2c3bd164-82a4-40db-b12d-04442f534da0

rolson24:
I don't think so. It seems like MASA is built on the trackers in the MMDet library, which are set up slightly different than the trackers in MMTracking. The trackers in MMTracking seem to output dicts of numpy arrays, while the trackers in MMDet output objects called `TrackDataSample`. Also I tried to test MMTracking and it seems that it does not support the newer versions of their own libraries. (Max MMCV version is 2.1.0 and newest is 2.2.0, and max MMDet version is 3.0.0 and newest is 3.3.0) So in order to use MMTracking with any recent version of CUDA and pytorch, a user would have to compile their own binaries. MMTracking has not had a commit since last April, so it seems like OpenMMLab doesn't have the bandwidth to support it right now.

So maybe not worth trying to support it if its a dead library anyway.

rolson24:
[...]

PR #1404, opened 2024-07-26
add support for mmdet tracking

rolson24:
# Description

Feature request from #1381 

Add support for MMDetection trackers by modifying `from_mmdetection` to support importing MMDetection `TrackDataSamples` into supvervision Detections. This also allows for easy integration with the [MASA tracker] (https://github.com/siyuanliii/masa/tree/1d5d42cdd75276c9a8e918639907916f75a761da)

## Type of change

Tested with this [colab notebook](https://colab.research.google.com/drive/1xnMHshHGhNtDHbDvvtHgFmhOAKDLHmhx?usp=sharing)

Right now the pre-trained MMDetection models are performing quite poorly for tracking, and the built in config files need to be fixed, but if you train your own detection model and write your own config, they work. All you have to do is call `tracker = init_track_model(TRACKER_CONFIG, device=device)' and then for each frame call:

```
track_result = inference_mot(tracker, frame,
                        frame_id=frame_idx,
                        video_len=video_info.total_frames,
                        )
```

### DIFF
--- a/supervision/detection/core.py
+++ b/supervision/detection/core.py
@@ -423,6 +423,16 @@ class Detections:
             detections = sv.Detections.from_mmdetection(result)
             ```
         """  # noqa: E501 // docs
+        if hasattr(mmdet_results, "pred_track_instances") and mmdet_results.pred_track_instances is not None:
+            return cls(
+                xyxy=mmdet_results.pred_track_instances.bboxes.cpu().numpy(),
+                confidence=mmdet_results.pred_track_instances.scores.cpu().numpy(),
+                class_id=mmdet_results.pred_track_instances.labels.cpu().numpy(),
+                mask=mmdet_results.pred_track_instances.masks.cpu().numpy()
+                if "masks" in mmdet_results.pred_track_instances
+                else None,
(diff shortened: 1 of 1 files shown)


onuralpszr:
@rolson24 is there chance tracker can be empty ? If yes we should do same thing we did for mask as well.

SkalskiP:
@rolson24 I won't be able to test it today. I'll leave it to @LinasKo. He should be back on Monday.

rolson24:
> @rolson24 is there chance tracker can be empty ? If yes we should do same thing we did for mask as well.

I tested it and it seems like if there are no tracks, then `pred_track_instances` won't exist, so it will just use the `pred_instances` which are the detection results. But it probably wouldn't hurt to add it especially if there is a tracker that implements `pred_track_instances` slightly differently.

Borda:
Thanks for working on this and the related issue #1381! Since we're deprecating the tracker module in `supervision` (see #2215) in favor of the new standalone [`trackers`](https://github.com/roboflow/trackers) package, tracking integrations like this are best continued there. The `from_mmdetection` Detections support part may still be relevant separately — feel free to open a focused PR for that if needed. Thanks for your contribution! 🙏

Issue #1388, opened 2024-07-21
Add new method to `sv.DetectionDataset` which will create/export proper/final structure of YOLO dataset for training


xaristeidou:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Description

I find myself creating dataset structures and split in train, valid, test and images, labels folders multiple times. The whole process could easily be automated.

### Use case

There are methods that currently exist to load a dataset `sv.DetectionDataset.from_yolo()`, split in selected ratio `sv.DetectionDataset.split()` and export to YOLO format `sv.DetectionDataset.to_yolo()`.

Nevertheless, in creation of YOLO model training dataset structure, someone must write a custom split in train/valid/test (because `split()` is able to split only in two parts), and also create manually the train/valid/test folders needed for the preparation of the dataset. (As of my knowledge ultralytics YOLO models require by default to have train/valid folders that contain valid and not empty annotations, test folder can be empty).

 For that reason I propose a new method to be added in `sv.DetectionDataset` which will combine the arguments of `from_yolo(), split(), to_yolo()` and will run the whole backend for creating train/valid/test folder and images/labels subfolders along with data.yaml file.

At this current point I have developed an implementation of such method which provides the ability to the user to create a YOLO dataset structure with a single line of code. An example of executing such a process can be seen in the following example:

```
import supervision as sv

dataset_directory = "/path/to/directory"

sv.DetectionDataset.create_yolo_dataset(
    images_directory_path=f"{dataset_directory}/images",
    annotations_directory_path=f"{dataset_directory}/labels",
    data_yaml_path=f"{dataset_directory}/data.yaml",
    train_ratio=0.7,
    valid_ratio=0.15,
    folders_export_path=f"{dataset_directory}",
    data_yaml_export_path = f"{dataset_directory}/data.yaml",
)
```

### Additional

- The `test_ratio` is automatically calculated based on the `train_ratio, valid_ratio`.
- The arguments provided in the example are all mandatory arguments to be passed.
- The user can provide any additional optional argument incorporated in `from_yolo(), split(), to_yolo()`.

Let me know if you like this idea, and if you want to submit a PR with the initial implementation.

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

SkalskiP:
Cześć @xaristeidou 👋🏻 

[...]

PR #1422, opened 2024-08-01
Add `train, val, test` folder paths in `data.yaml` at `save_data_yaml()`

xaristeidou:
# Description

In the process of developing a notebook as scheduled in #1388, I used the `sv.DetectionDataset().as_yolo()` method which executes in the backed the `save_data_yaml()` function to create the `data.yaml` file need for the dataset.

As YOLO the model construction for training, and the documentation, the model needs prerequisite `train`, `val` arguments in the `data.yaml` file, and the `test` argument is not needed but could be passed also if a test dataset exists. When trying to run `model.train()` the following error raises:
`SyntaxError: /content/dataset/data.yaml 'train:' key missing ❌.
'train' and 'val' are required in all data YAMLs.`

Therefore in `save_data_yaml()` except the `nc`, `names` arguments we should export also the `train`, `val`, `test` paths in order to be ready for executing the `model.train()` process. I think we should export the default paths as follows:
 - train: train/images
 - val: valid/images
 - test: test/images

If someone has a different working directory than the root of the folder containing the data.yaml, should change these paths manually. At least it will be easier to debug and modify the path if needed than to add the arguments in the yaml file.

None

Using `sv.DetectionDataset.as_yolo()` exports the data.yaml file with prerequisite `train, val, test` paths. 

None

## Docs

-   [ ] Docs updated? **No need to**


xaristeidou:
@SkalskiP Hello there 👋🏻! Yes indeed, if we want to create train, valid, test subset folders for the dataset training we have to run the process one time for each subset. Nevertheless, all 3 subset folder datasets contain the same `names` and `nc` values. Even if we export using `as_yolo()` for train, valid, or test, in the end all of them must have the same `data.yaml`. I think the problem we encounter is when someone doesn't want to export all 3 subset folders (train, valid, test). In this case, it will automatically add `train`, `val`, `test` paths in the `data.yaml` file. Even if the addition of these paths in the yaml file won't raise any error in supervision methods and functions, it kind of assumes that there are 3 subsets of the dataset, which contradicts with the build and usage of the `as_yolo()` method.

I think there are 3 scenarios in that case:

1. Add `train`, `val`, `test` folder paths in `data.yaml` in order to be ready for any training process even if it contradicts (the truth is that now I don't really like the concept that something out of the method scope will be executed)
2. We could add a log message in the execution of `save_data_yaml()` which runs internal in `as_yolo()`, to inform users to define the paths for training process if needed.
3. Change nothing and leave it as it is. In this case the user will have to manually find out for the subset folders paths to define if training process is needed. In my notebook I could add the `train`, `val`, `test` path using code, in order to showcase how could be done in this way also.

SkalskiP:
@xaristeidou, `the problem we encounter is when someone doesn't want to export all 3 subset folders`, or it could be that the person simply doesn't know they should.

One idea I had is that `as_yolo` would create `data.yaml` if it doesn't exist, but append to it if it does. When you call `data.yaml`, you would also specify whether it's the `train`, `test`, or `val` subset. What do you think about it? IS it too complicated?

xaristeidou:
@SkalskiP Well not too complicated. We could modify `as_yolo()` which is the function that users have access to when exporting, and we could add an argument `subset_type: Optional[str] = None` and if provided will accept one of the following ["train", "val", "test"] and run a brief check if the value is in this list. If not provided, won't write in yaml the path. 

`as_yolo would create data.yaml if it doesn't exist, but append to it if it does` the problem in this case is that maybe there is a `data.yaml` already which has some previous configuration (maybe different `names` and `nc`). I think we should stick with the current implementation that modifies the yaml file in order to avoid not updating `names` and `nc` if needed. In the case that someone runs as yolo three times, one for `train`, `valid`, `test` folders respectively, to create a dataset, then `names` and `nc` will be the same (because of `split()` which keeps same attributes), so there will be no changes, if they are not the same this will be a mistake in the dataset from the user beforehand if he/she didn't use `split()` and loaded two different datasets with different configuration, so it should be fixed in a previous code section.

SkalskiP:
@xaristeidou, wanna try to implement the PoC of this solution?

xaristeidou:
@SkalskiP Yes! I will be back when it is ready.

xaristeidou:
@SkalskiP I have added the new changes. Here is a Colab notebook for easy testing.

[https://colab.research.google.com/drive/1BL7c2ycXkuCrEE5JOqh7u7Zf1kJzgb43?usp=sharing](url)

xaristeidou:
@SkalskiP Did you manage to take a look at the new committed changes and test with the notebook?

PR #1442, opened 2024-08-12
Core system for Metrics

LinasKo:
# Description

This PR adds metrics as a standalone supervision module.

## Type of change

* At the time of opening the draft PR, I'm till writing tests for it.
* I timed the code, making sure it doesn't add much overhead.
* I'll try using it in a few application, checking that it's pleasant to use.

If we'll want pandas to work, we'll need to do `pip install 'supervision[metrics]'`.

## Docs

Not updated, but they should be.


supervision/metrics/core.py
```diff
@@ -0,0 +1,210 @@
...
+        class_set = set(class_id)
+        if len(class_set) >= 2 and -1 in class_set:
+            raise ValueError(
+                "Metrics store received results with partially defined classes."
+            )
+
+    def _validate_shape(self, data: npt.NDArray) -> None:
+        if self._datapoint_shape is None:
+            assert self._metric_target == MetricTarget.MASKS
+            self._datapoint_shape = data.shape[1:]
```
SkalskiP, line 205: What's happening here? What if your dataset contained images of different sizes and, by extent, masks of different sizes?

supervision/metrics/intersection_over_union.py
```diff
@@ -0,0 +1,136 @@
...
+    #     combined = np.column_stack((stacked_class_ids, stacked_ious))
+
+    #     column_names = \
+    # ['class_id'] + [f'col_{i+1}' for i in range(stacked_ious.shape[1])]
+    #     result = pd.DataFrame(combined, columns=column_names)
+
+    #     return result
+
+    @staticmethod
+    def _compute_box_iou(
```
SkalskiP, line 119: Why do we need to duplicate the logic from `box_iou_batch`?

pyproject.toml
```diff
@@ -51,15 +51,19 @@ scipy = [
...
 pyyaml = ">=5.3"
 defusedxml = "^0.7.1"
+pillow = ">=9.4"
 opencv-python = { version = ">=4.5.5.64", optional = true }
 opencv-python-headless = ">=4.5.5.64"
 requests = { version = ">=2.26.0,<=2.32.3", optional = true }
 tqdm = { version = ">=4.62.3,<=4.66.5", optional = true }
-pillow = ">=9.4"
+# pandas: picked lowest major version that supports Python 3.8
+pandas = { version = ">=2.0.0", optional = true }
```
SkalskiP, line 60: @onuralpszr, do we need to do something extra here?

supervision/metrics/__init__.py
```diff
@@ -0,0 +1,2 @@
+from supervision.metrics.core import Metric, MetricTarget, UnsupportedMetricTargetError
```
SkalskiP, line 1: Do we need to expose the `UnsupportedMetricTargetError`?

supervision/metrics/intersection_over_union.py
```diff
@@ -0,0 +1,136 @@
...
+    def update(
+        self,
+        data_1: Union[npt.NDArray, Detections],
+        data_2: Union[npt.NDArray, Detections],
+    ) -> Metric:
+        """
+        Add data to the metric, without computing the result.
+
+        The arguments can be:
+        * Boxes of shape (N, 4), float32,
```
SkalskiP, line 44: Will that render properly in `mkdocs`?

supervision/metrics/intersection_over_union.py
```diff
@@ -0,0 +1,136 @@
...
+        self._store = InternalMetricDataStore(metric_target, class_agnostic)
+
+    def reset(self) -> None:
+        self._store.reset()
+
+    def update(
+        self,
+        data_1: Union[npt.NDArray, Detections],
+        data_2: Union[npt.NDArray, Detections],
+    ) -> Metric:
```
SkalskiP, line 39: Do we return `Metric` or `IntersectionOverUnion` here?

supervision/metrics/intersection_over_union.py
```diff
@@ -0,0 +1,136 @@
...
+
+        Returns:
+            Dict[int, npt.NDArray[np.float32]]: A dictionary with class IDs as keys.
+            If no class ID is provided, the key is the value CLASS_ID_NONE.
+        """
+        # TODO: cache computed result.
+        ious = {}
+        for class_id, array_1, array_2 in self._store:
+            if self._metric_target == MetricTarget.BOXES:
+                if array_1 is None or array_2 is None:
```
SkalskiP, line 73: Feels like this logic should be placed in `InternalMetricDataStore`.

supervision/metrics/core.py
```diff
@@ -0,0 +1,210 @@
...
+    The base class for all supervision metrics.
+    """
+
+    @abstractmethod
+    def update(self, *args, **kwargs) -> Metric:
+        """
+        Add data to the metric, without computing the result.
+        Return the metric itself to allow method chaining, for example:
+
+        Example:
```
SkalskiP, line 28: I think we can skip the examples in base class.

SkalskiP (changes requested):

supervision/metrics/__init__.py
LinasKo: Yes. People need to know what to catch in their `with` statements. Catching a general `Exception` is bad practice.

supervision/metrics/intersection_over_union.py
LinasKo: Good catch. I put no effort into docs yet, but I don't think this will look good.

supervision/metrics/intersection_over_union.py
```diff
@@ -0,0 +1,136 @@
...
+        self._metric_target = metric_target
+        self._class_agnostic = class_agnostic
+        self._iou_threshold = iou_threshold
+
+        self._store = InternalMetricDataStore(metric_target, class_agnostic)
+
+    def reset(self) -> None:
+        self._store.reset()
+
+    def update(
```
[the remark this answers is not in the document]
LinasKo: I'll see how `*Detections` feel like. Might be a bit cleaner than a large Union.

supervision/metrics/intersection_over_union.py
LinasKo: Good catch. Should be the latter

supervision/metrics/intersection_over_union.py
LinasKo: Addressed, looks okay.

image

supervision/metrics/core.py
LinasKo: removed

supervision/metrics/intersection_over_union.py
LinasKo: addressed

supervision/metrics/intersection_over_union.py
LinasKo: Removed for now.

My initial idea was to replace `box_iou_batch` with the metric. That's still on the table, but let's talk about it when the metric is complete.

supervision/metrics/core.py
```diff
@@ -0,0 +1,210 @@
...
+        if self._class_agnostic or isinstance(data, np.ndarray):
+            return np.array([CLASS_ID_NONE] * len(data), dtype=int)
+        assert isinstance(data, Detections)
+        if data.class_id is None:
+            return np.array([CLASS_ID_NONE] * len(data), dtype=int)
+        return data.class_id
+
+    def _validate_class_ids(self, class_id: npt.NDArray[np.int_]) -> None:
+        class_set = set(class_id)
+        if len(class_set) >= 2 and -1 in class_set:
```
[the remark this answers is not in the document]
LinasKo: 1.
Changed to `CLASS_ID_NONE`, exposed it in `__init__` and moved the docstring where it would apply to the `CLASS_ID_NONE` variable.

Here's the broader question. Do we try to always return a dict `dict[int, MetricSpecificResult]`, or return only `MetricSpecificResult` when using class-agnostic metrics? Packages like torchmetrics chose the latter, returning the result only, if `class_agnostic` is set.

I'm a proponent of less ambiguity in return types. The question is, whether the following is cumbersome or not:

```python
detections_1 = ...
detections_2 = ...

iou_metric = sv.metrics.IntersectionOverUnion(class_agnostic=True)
iou_metric.update(detections_1, detections_2)

ious = iou_metric.compute()[sv.metrics.CLASS_ID_NONE]
# or
ious = iou_metric.compute()[None]
# or
ious = iou_metric.compute()[-1]

# versus

ious = iou_metric.compute()
```

Case 1 is cumbersome unless we can reduce size of `CLASS_ID_NONE` or put it in global scope.
Case 2 I find rather clear, but people will need to check their keys for None or ignore a lot of mypy squiggles. It may also clash with 0.
Case 3 feels unnatural, as it seems you're popping off the end of the array.

Case 4 is cleanest, but will always need an extra check to determine which of the 2 Union outputs was returned.

That's the reason for looking at case 1 and having `CLASS_ID_NONE`. It gives a fairly reasonable API, even if verbose.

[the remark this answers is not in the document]
LinasKo: 2.
> I'm also not sure what are we checking here?

Good catch, the input should have been formed of both metrics. I did it inside the function now.
I'm still thinking of this one.

supervision/metrics/core.py
```diff
@@ -0,0 +1,210 @@
...
+            obb = data.data.get(config.ORIENTED_BOX_COORDINATES, None)
+            if isinstance(obb, list):
+                obb = np.array(obb, dtype=np.float32)
+            return obb
+        raise ValueError(f"Invalid metric target: {self._metric_target}")
+
+    def _get_class_ids(
+        self, data: Union[npt.NDArray, Detections]
+    ) -> npt.NDArray[np.int_]:
+        if self._class_agnostic or isinstance(data, np.ndarray):
```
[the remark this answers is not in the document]
LinasKo: Exactly. Detections provide the structure that links `class_id` and `xyxy.

Do you think we should allow arrays of shape (N, 5) as well?

supervision/metrics/intersection_over_union.py
```diff
@@ -0,0 +1,136 @@
...
+        return self
+
+    def compute(self) -> Dict[int, npt.NDArray[np.float32]]:
+        """
+        Compute the Intersection over Union metric (IoU)
+        Uses the data set with the `update` method.
+
+        Returns:
+            Dict[int, npt.NDArray[np.float32]]: A dictionary with class IDs as keys.
+            If no class ID is provided, the key is the value CLASS_ID_NONE.
```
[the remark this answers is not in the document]
LinasKo: Broader discussion in: https://github.com/roboflow/supervision/pull/1442#discussion_r1716908777

Leaving this open until we resolve that.

pyproject.toml
onuralpszr: Sorry, I miss this message because was hidden, making optional to add is correct I also see "stub" package which also optional also good.  If I may I would like to also say If you need faster metrics and better RAM management, We should strongly consider "polars" as well.  I can give you raw example and send you colab example what I meant about it.  Plus we can convert to pandas as well.  or read from pandas,numpy etc as well. So cards are still open. You saw speed of "ruff" (rust) you can see this one too.

onuralpszr: Doc references:

https://docs.pola.rs/api/python/stable/reference/functions.html
https://docs.pola.rs/api/python/stable/reference/dataframe/export.html

I can show more and explain as well.

LinasKo:
Here's the status of metrics at the moment:
* mAP is now supported, for masks and boxes
* mAP returns results for all, small, medium, and large objects

To change:
* mAP uses prior IoU function rather than metrics
* IoU metric on masks is much slower, and produces different results
* No plotting function for `mAP` results
* Still need to add `mAR`
* Untested at scale

mkdocs.yml
```diff
@@ -45,50 +44,50 @@ nav:
...
-        - Iterables: utils/iterables.md
-        - Notebook: utils/notebook.md
-        - File: utils/file.md
-        - Draw: utils/draw.md
-        - Geometry: utils/geometry.md
-    - Assets: assets.md
+      - Detection and Segmentation:
+          - Core: detection/core.md
+          - Annotators: detection/annotators.md
+          - Metrics: detection/metrics.md
```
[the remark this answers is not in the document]
LinasKo: Added docs, examples, wrote down the legacy metrics docs, aggregated the pages in a single section.

supervision/metrics/utils/internal_data_store.py
```diff
@@ -0,0 +1,301 @@
...
+
+        mew_mask_list = unify_pad_masks_shape(*self._content_list)
+        self._content_list = [np.vstack(mew_mask_list)]
+
+    def _merge_class_id(self):
+        """Merge all class IDs into a single array."""
+        if len(self._class_id_list) < 2:
+            return
+        self._class_id_list = [np.hstack(self._class_id_list)]
+        print(self._class_id_list)
```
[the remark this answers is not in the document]
LinasKo: done.

PR #1448, opened 2024-08-14
Annotator types and docs: MyPy, consistent 'optional', small tidy-ups

LinasKo:
# Description

MyPy was showing many issues in the `keypoints/annotators.py` and `annotators/core.py`, so I cleaned it all up.

Where the issues branched into other files, I likewise mopped it up. Here's a [Loom](https://www.loom.com/share/32cd83216e8742e5b56e0567756e5a5c?sid=6789b55e-bbcc-4c50-a19e-984e01096fea). I strongly suggest reviewing one commit at-a-time.

Lastly, there are a few minuscule miscellaneous fixes.

Fixes:
1. Changed a few return types from `np.array` to `np.ndarray`.
2. VSCode's mypy would always highlight `Color.X` and `ColorPalette.X` in red, as it would see it as having type `classproperty`. Changed it so it correctly sees `Color` and `ColorPalette`.
3. We have many variables marked optional. Now, in annotators, they're all the same - marked `Optional[dtype]` if and only if they can be `None`.
4. `ImageType` wouldn't be seen as `np.ndarray`, as we type-annotate it as `ImageType`. I added assertions, which are the easiest way to tell mypy which one it is.
    * There was also one case of missing decorator
5. Commit "[Add error handling, clean up mypy confusion](https://github.com/roboflow/supervision/pull/1448/commits/0477499b773975d1addc2c34029375e29e33960b)" addresses multiple issues that were revealed after superficial MyPy issues were cleaned up.
6. Some examples were missing `sv.`.
7. Keypoint docs previously said that `sv.KeyPoints` holds things of shape `(N, 2)`, where it's actually `(N, M, 2)`. 

## Type of change

Tested locally.
Colab: https://colab.research.google.com/drive/1UjRYth4NUfAsvQiz5-CfyRJ9zbXxj2i-?usp=sharing

Many docstrings affected in a small way - have a quick look at type signatures, etc.

## Docs

### DIFF
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -3,6 +3,7 @@ from typing import List, Optional, Tuple, Union
 
 import cv2
 import numpy as np
+import numpy.typing as npt
 from PIL import Image, ImageDraw, ImageFont
 
 from supervision.annotators.base import BaseAnnotator, ImageType
@@ -263,7 +264,7 @@ class OrientedBoxAnnotator(BaseAnnotator):
             return scene
 
         for detection_idx in range(len(detections)):
-            bbox = np.intp(detections.data.get(ORIENTED_BOX_COORDINATES)[detection_idx])
+            bbox = detections.xyxy[detection_idx].astype(int)
             color = resolve_color(
                 color=self.color,
                 detections=detections,
@@ -1139,8 +1140,8 @@ class LabelAnnotator(BaseAnnotator):
 
             if labels is not None:
                 text = labels[detection_idx]
-            elif detections[CLASS_NAME_DATA_FIELD] is not None:
-                text = detections[CLASS_NAME_DATA_FIELD][detection_idx]
+            elif CLASS_NAME_DATA_FIELD in detections.data:
+                text = detections.data[CLASS_NAME_DATA_FIELD][detection_idx]
             elif detections.class_id is not None:
                 text = str(detections.class_id[detection_idx])
             else:
@@ -1355,8 +1356,8 @@ class RichLabelAnnotator(BaseAnnotator):
 
             if labels is not None:
                 text = labels[detection_idx]
-            elif detections[CLASS_NAME_DATA_FIELD] is not None:
-                text = detections[CLASS_NAME_DATA_FIELD][detection_idx]
+            elif CLASS_NAME_DATA_FIELD in detections.data:
+                text = detections.data[CLASS_NAME_DATA_FIELD][detection_idx]
             elif detections.class_id is not None:
                 text = str(detections.class_id[detection_idx])
             else:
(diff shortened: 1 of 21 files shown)


LinasKo:
Testing checklist:
- [ ] classproperty: use of Color, ColorPalette
- [ ] Optional in docs
- [ ] got @SkalskiP's opinion on assertions in annotators
- [ ] HeatmapAnnotator

onuralpszr:
@LinasKo I added my first initial very small change into toml file about reduce mypy error and make sure it does proper mypy check I got them from by using  `mypy --install-types` and added into toml as "optional" so it won't install unless type `poetry install --with typechcek` so I didn't want to burden anyone to install type check libraries in development.

onuralpszr:
I removed our "isort" original and params of isort and doing all via ruff and remove extra isort from pre-commit too.

LinasKo:
Hi @onuralpszr ,

I've had a look at this, and I'd like you to remove your commits for now.

The PR is mostly focused on annotators typing, and is already past the limits of how much should be contained in a single PR.

I see the isort is immediately active, yet if some of the changes need a separate install, it'd likely sit idle as dead code until an indeterminate time in the future, which I don't want.

onuralpszr:
@LinasKo I backup my changes and I will move to another PR so we can merge mypy/isort formatting changes. I guess that would work

LinasKo:
Yep, happy with that!

supervision/utils/internal.py
```diff
@@ -134,17 +137,27 @@ def my_method(cls):
...
+            owner_self: The instance through which the attribute was accessed, or None.
+                Irrelevant for class properties.
+            owner_cls: The class through which the attribute was accessed.
 
         Returns:
-        The result of calling the function stored in 'fget' with 'owner_cls'.
+            The result of calling the function stored in 'fget' with 'owner_cls'.
         """
+        if self.fget is None:
+            raise AttributeError("unreadable attribute")
```
SkalskiP, line 160: This error message seems a bit nondescriptive. I know it's an internal function, and only we will use it, but I think we can still provide a bit more information.

supervision/annotators/core.py
```diff
@@ -1975,7 +2006,7 @@ def __init__(
         border_color: Color = Color.BLACK,
         position: Position = Position.TOP_CENTER,
         color_lookup: ColorLookup = ColorLookup.CLASS,
-        border_thickness: int = None,
+        border_thickness: Optional[int] = None,
```
SkalskiP, line 2009: Following your logic. shouldn't `border_thickness` be marked as `Optional[int]` in docs?

supervision/dataset/core.py
```diff
@@ -181,11 +181,11 @@ def split(
             using the provided split_ratio.
 
         Args:
-            split_ratio (float, optional): The ratio of the training
+            split_ratio (float): The ratio of the training
```
SkalskiP, line 184: I noticed that we added types here but the actually function definition does not have type hints :/

SkalskiP (changes requested):

supervision/utils/internal.py
LinasKo: This matches the API for properties - I tried to get as close as possible to the actual implementation. Do we still want to be more descriptive?

supervision/annotators/core.py
LinasKo: It absolutely should. Thank you!

LinasKo:
@SkalskiP, I've addressed the requests, except for `classproperty` error message: https://github.com/roboflow/supervision/pull/1448#discussion_r1731121297

PR #1473, opened 2024-08-22
formatter: 🔧 improvements for pre-commit and ruff also apply minor mypy/pep fixes

onuralpszr:
This changes also introduce enablement of extra ruff rules as well

- https://docs.astral.sh/ruff/rules/#flake8-quotes-q
- https://docs.astral.sh/ruff/rules/#flake8-builtins-a
- https://docs.astral.sh/ruff/rules/#warning-w
- https://docs.astral.sh/ruff/rules/#isort-i
- Yanked setuptools version fixed in lock files.

This changes introduce small fixes but there are other rules I would like to enable in time slowly transition comply better to pep and other python guideline rules as well. 

Thank you.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,6 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.6.0
     hooks:
-      - id: end-of-file-fixer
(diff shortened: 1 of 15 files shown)


LinasKo (approved):

LinasKo:
Merging. Thank you @onuralpszr!

Issue #1438, opened 2024-08-08
[KeypointsDataset] - create a new dataset format to load keypoints


SkalskiP:
### Description

Supervision can currently load object detection, instance segmentation, and classification datasets. Let's add the ability to handle keypoint datasets. The sv.KeypointDataset should offer the same functionalities as [sv.DetectionDataset](https://supervision.roboflow.com/latest/datasets/core/) and [sv.ClassificationDataset](https://supervision.roboflow.com/latest/datasets/core/#supervision.dataset.core.ClassificationDataset), including `merge`, `split`, `__getitem__`, `__iter__`, `from_yolo`, and `as_yolo`.

### Additional

- Note: Please share a Google Colab with minimal code to test the new feature. We know it's additional work, but it will speed up the review process. The reviewer must test each change. Setting up a local environment to do this is time-consuming. Please ensure that Google Colab can be accessed without any issues (make it public). Thank you! 🙏🏻

shaddu:
Hi @SkalskiP 👋 

I was analyzing the issue and understood that KeypointsDataset will be similar to DetectionDataset, with the difference being the type of annotations.

```
classes: List[str],
images: Union[List[str], Dict[str, np.ndarray]],
annotations: Dict[str, KeyPoints]
```

If this is a correct understanding then I will like to work on this issue.

LinasKo:
Let's only implement the new, lazy-loaded, non-deprecated API, with `images: List[str]`. This will save work in the future.

LinasKo:
@shaddu, I also made this template, which should help with starting your colab: [Colab Template](https://colab.research.google.com/drive/1rin7WrS-UvVIe-_Gfxmu-yVslGphOq89?usp=sharing).
Feel free to use as much or as little of it as you wish 😉

shaddu:
Hi @LinasKo  👋

I also noticed the deprecated warning in the Datasets classes, and the clarification around `images: List[str]` is helpful. 

Thank you for sharing the Colab template 😎 . It will be helpful, and I will suggest adding some similar Colab templates to the contribution guide.

miteshashar:
@shaddu Did you happen to take this up? I was thinking of reviewing this to evaluate whether it is something I can take up to attempt.

SkalskiP:
Hi @miteshashar 👋🏻 if you are up for the task I'm happy to assign it to you ;)

shaddu:
Hi @SkalskiP and @miteshashar  👋🏻

[...]

PR #1477, opened 2024-08-24
feat : added support for keypoints dataset

shaddu:
# Description

Added support for keypoints dataset

## Type of change

[Keypointdataset Colab](https://colab.research.google.com/drive/1zcKmoc9Y-HBystC7mE8IQUofrP-vGJqD?usp=sharing)

## Docs

### DIFF
--- a/supervision/dataset/core.py
+++ b/supervision/dataset/core.py
@@ -21,6 +21,7 @@ from supervision.dataset.formats.pascal_voc import (
 )
 from supervision.dataset.formats.yolo import (
     load_yolo_annotations,
+    load_yolo_keypoint_annotations,
     save_data_yaml,
     save_yolo_annotations,
 )
@@ -32,6 +33,7 @@ from supervision.dataset.utils import (
     train_test_split,
 )
 from supervision.detection.core import Detections
+from supervision.keypoint.core import KeyPoints
 from supervision.utils.internal import deprecated, warn_deprecated
 from supervision.utils.iterables import find_duplicates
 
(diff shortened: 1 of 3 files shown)


onuralpszr:
First suggestion I would say we should split the code to 2 files detections and keypoints and maybe also add `base` too because file getting really bigger and mix both detections and keypoints not the way I would do it. I would like to hear from others as well. 

cc @LinasKo  @SkalskiP

LinasKo:
@onuralpszr, you're very right that it needs a split. It's a massive file.

Let's do that right after the next release, but merge this PR as is.
(after I review it)

@shaddu, fantastic work, as always.

I have one request early request: please add types to each of the `split` methods, for both classification, detection and keypoint dataset.

```
    def split(
        self,
        split_ratio: float = 0.8,
        random_state: Optional[int] = None,
        shuffle: bool = True,
```

Another PR will introduce this change soon, so adding it here will pre-empt the conflict.

shaddu:
Hi @onuralpszr and @LinasKo ,

I also feel that files should be split based on keypoints and detection types as there could be confusion while working with annotations. 

@LinasKo  I have added type to the split method for keypoint dataset and can see you have already added for the rest. 

Thank you for reviewing the PR ;)

miteshashar:
Thanks for the update on #1438, @shaddu :)

@SkalskiP @shaddu I'd been evaluating the task and also checked the PR code.

Given their relevance & existing support in the `Keypoint` class, I feel it would be helpful to also add `from_coco`, `as_coco`, `from_mediapipe` and `as_mediapipe` methods.

Borda:
@shaddu, I think this was implemented by @SkalskiP in some parallel path…?

Borda:
Thank you, @shaddu, for putting this together. The goal of making pose/keypoint datasets easier to load and export is still reasonable, and I appreciate the work that went into the initial implementation.

However, I’m closing this PR because it has drifted too far from the current codebase to review or merge safely. The branch conflicts with `develop`, uses package paths that no longer exist, and the proposed implementation predates the current `KeyPoints` API and dataset conventions.

More importantly, YOLO pose data needs a fresh design, not a conflict-resolution pass. A pose label includes the object class, bounding box, keypoints, and, where applicable, per-keypoint visibility. The current patch does not preserve that full record or the pose metadata needed in the dataset YAML, so it cannot reliably round-trip current YOLO pose datasets.

If there is still demand for this capability, please open a new PR from the current `develop` branch. It should begin with a short proposal for the container/API shape, then include round-trip tests for 2-D and visibility-based keypoints, multiple instances, empty annotations, bounding-box behavior, and the required pose YAML fields such as `kpt_shape` and `flip_idx`.

Keeping this PR as historical context is useful; a clean, current implementation will be much easier to review and maintain.

shaddu:
@Borda Thanks for the update, indeed the PR was quite old by now and YOLO pose data needs a fresh design.

PR #1526, opened 2024-09-19
refactor: ✨ ruff rules enabled and fixed and code refactor made

onuralpszr:
# Description

I added "ruff" formatting rule to be enabled (https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf)

References

https://peps.python.org/pep-0448/
https://docs.astral.sh/ruff/rules/unnecessary-iterable-allocation-for-first-element/
https://docs.astral.sh/ruff/rules/mutable-class-default/

### DIFF
--- a/supervision/keypoint/skeletons.py
+++ b/supervision/keypoint/skeletons.py
@@ -1,11 +1,11 @@
 from enum import Enum
-from typing import Dict, List, Tuple
+from typing import ClassVar, Dict, List, Tuple
 
 Edges = List[Tuple[int, int]]
 
 
 class Skeleton(Enum):
-    COCO = [
+    COCO: ClassVar[List[Tuple[int, int]]] = [
         (1, 2),
         (1, 3),
         (2, 3),
@@ -25,7 +25,7 @@ class Skeleton(Enum):
         (17, 15),
     ]
 
-    GHUM = [
+    GHUM: ClassVar[List[Tuple[int, int]]] = [
         (1, 2),
         (1, 5),
         (2, 3),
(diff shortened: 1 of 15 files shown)


supervision/keypoint/skeletons.py
```diff
@@ -1,11 +1,11 @@
 from enum import Enum
-from typing import Dict, List, Tuple
+from typing import ClassVar, Dict, List, Tuple
 
 Edges = List[Tuple[int, int]]
 
 
 class Skeleton(Enum):
-    COCO = [
+    COCO: ClassVar[List[Tuple[int, int]]] = [
```
LinasKo, line 8: 1. We can use `Edges` here.
2. Is `ClassVar` required here? Does it capture something about the contained data that would otherwise confuse MyPy or the user? In C++ terms - it feels more related to the `lvalue` than the `rvalue`. More about how the data is stored than what the data is.

LinasKo (changes requested):

supervision/keypoint/skeletons.py
LinasKo: Reading up about mutable class var annotations

LinasKo: This is an enum, so it would make sense for the values to be immutable. Can we try changing Edges to `Tuple[Tuple[int, int]]`? If there's issues somewhere, let's then keep `ClassVar[Edges]`.

onuralpszr: @LinasKo  done

LinasKo (approved):
Looks good!

LinasKo:
Thanks @onuralpszr! Merging.

PR #1528, opened 2024-09-19
let ByteTrack to maintain track ID per instance

grzegorz-roboflow:
# Description

`BaseTrack` implementation includes `_count` class var, similarly `STrack` implementation includes `_external_count` class var. This results in undesirable behavior when running more than one tracker in the same process. In [inference](https://github.com/roboflow/inference/pull/642) we added a block wrapping ByteTrack, if more than 1 video is processed, objects across videos will receive non-deterministic track IDs.

In this change class vars are removed and track ID assignment is handled by upper layer (`ByteTrack`)

## Type of change

Added unit test to cover this scenario.

N/A

## Docs

N/A

### DIFF
--- a/supervision/tracker/byte_tracker/basetrack.py
+++ b/supervision/tracker/byte_tracker/basetrack.py
@@ -12,9 +12,8 @@ class TrackState(Enum):
 
 
 class BaseTrack:
-    _count = 0
-
     def __init__(self):
+        self._count = 0
         self.track_id = 0
         self.is_activated = False
         self.state = TrackState.New
@@ -34,18 +33,16 @@ class BaseTrack:
     def end_frame(self) -> int:
         return self.frame_id
 
-    @staticmethod
-    def next_id() -> int:
-        BaseTrack._count += 1
-        return BaseTrack._count
+    def next_id(self) -> int:
+        self._count += 1
+        return self._count
 
-    @staticmethod
-    def reset_counter():
-        BaseTrack._count = 0
-        BaseTrack.track_id = 0
-        BaseTrack.start_frame = 0
-        BaseTrack.frame_id = 0
-        BaseTrack.time_since_update = 0
+    def reset_counter(self):
+        self._count = 0
+        self.track_id = 0
+        self.start_frame = 0
+        self.frame_id = 0
+        self.time_since_update = 0
 
     def activate(self, *args):
         raise NotImplementedError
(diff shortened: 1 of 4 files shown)


LinasKo:
I finally found time to look at this. The PR does exactly what it claims, but it only solves half of the greater problem we have - dependence between multiple `ByteTrack` instances.

Two `ByteTrack` objects should not share state. In a single video, all `STracks` of one tracker *should* share:
1. ID generation logic
2. Kalman Filter data\*  (error, weights - not `xyah+Δxyah`)

The proposed solution moves ID generation up the hierarchy, but not the Kalman filter. I'll clean this one up, merge it, and immediately follow-up with another PR, moving the Kalman Filter.

\* I'm a bit skeptical whether `shared_kalman` makes sense, even when the original authors have it in [their codebase](https://github.com/ifzhang/ByteTrack/blob/main/yolox/tracker/byte_tracker.py). Yes, every object in a single scene may have a similar motion and observation error. But does it make more sense than giving each one a Kalman, or having a Kalman per-class?

grzegorz-roboflow:
In inference we have batch of images - more than 1 video can be processed simultaneously, I'd imagine each stream requires separate tracker and this is how we handle it in inference, by creating separate instance of ByteTrack per stream. We want tracker_id to be consistent (i.e. we don't want one stream to interfere with tracker_id assigned in another stream)

LinasKo:
I slept on it, and was slightly wrong: there's no ambiguity - we'll never support multiple trackers at once applied to the same `Detections` object. Your proposal, save for the issue described, was correct.

---

Testing unfortunately revealed a regression in the external ID system. It's used to trim down the visible number of track IDs - see video below.

Since only individual `STracks` know if an external ID is needed, they need to control when the external ID is set - `ByteTrack` does not know that. Yet, there needs to be state above to hold this info. Long story short - we now have an `IdCounter` class.

Video before my change:

https://github.com/user-attachments/assets/7b832c24-476c-48a9-a404-8080844695b1

Video after my change:

https://github.com/user-attachments/assets/84f5fbd2-0344-4a31-944b-5ac0b38916f1

Thanks @grzegorz-roboflow! I can finally merge it. It was also my first time engaging with the tracker code. Looking at it, let's say a *tiny bit* of remodeling coming. 🔥🪓

LinasKo (approved):

PR #1536, opened 2024-09-22
Create lintscan

sharkeshd:
lint Complete code scanner

# Description

## Type of change

YOUR_ANSWER

## Docs

### DIFF
--- a/.github/workflows/lintscan
+++ b/.github/workflows/lintscan
@@ -1,0 +1,40 @@
+name: CI Pipeline
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  checkout:
+    runs-on: ubuntu-latest
+    outputs:
+      working-directory: ${{ steps.checkout.outputs.working-directory }}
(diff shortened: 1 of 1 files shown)


onuralpszr:
@sharkeshd 👋  hello, first of all thank you for pull request. I understand that devskim github action created and I checked devskim cli and setup in my computer to see which security problems we can detect but most of them was false positive and personally I found hard to understand but in cli because json file parsing not enough and we have "bandit" and "synk" If you check our readme file you can see health plus we do have github dependency bot checking package version for keep us update to date and also check for security vulnerabilities. If I was going to setup github action I personally would choice "https://github.com/pyupio/safety" which is more optimized to python code and gives better result, I think devskim not mature enough to enable in my tests. Maybe there is something I don't know but I spend good amount of time to see how is it going to react but I didn't see much of a benefit as well. 

Thank you.

cc @LinasKo

LinasKo:
Thank you for checking this, @onuralpszr.

If it adds no benefit thus far and even shows false-positives, then it's not the right tool for the job.

PR #1540, opened 2024-09-23
Extend LineZone to filter out miscounts

grzegorz-roboflow:
# Description

LineZone is checking if line was crossed by comparing detection position relative to the line across 2 frames. This is resulting in miscounts when bbox flickers.

Below video shows difference in counts between current develop and this PR.

https://github.com/user-attachments/assets/3ec55604-c2a1-49f2-8416-129e55665ec2

In this PR new init param is added called `max_linger`, which is by default set to `1` to ensure unchanged default behavior. When `max_linger` is set to a number greater than `1` then detection must cross on all frames in order to be counted. This results in slight delay (i.e. when `max_linger` is set to `10` on `10fps` video then objects will be counted with `1s` delay) however more accurate counts.

## Type of change

Added unit tests

N/A

## Docs

I updated `LineZone` init docstring.


LinasKo:
Hi @grzegorz-roboflow,

I'd witnessed this but hadn't verified it. Very nice contribution yet again! I'll take some time to cehck it, especially regarding how the model IDs / edge cases.

supervision/detection/line_zone.py
```diff
@@ -233,21 +241,65 @@ def trigger(self, detections: Detections) -> Tuple[np.ndarray, np.ndarray]:
...
 
-            if self.tracker_state.get(tracker_id) == tracker_state:
+            crossing_in_progress = (
+                crossing_state.count(True) != 0 and crossing_state.count(False) != 0
+            )
+            crossing_state.appendleft(tracker_state)
+            all_on_same_side = crossing_state.count(not tracker_state) == 0
+            if class_ids:
+                if len(class_ids) != len(crossing_state_class_ids) or not all(
+                    class_ids == crossing_state_class_ids
```
LinasKo, line 263: Note to self: possible bug here, as `class_ids` is actually `int` - not a list. (fixing along with my changes)

supervision/detection/line_zone.py
```diff
@@ -233,21 +241,65 @@ def trigger(self, detections: Detections) -> Tuple[np.ndarray, np.ndarray]:
...
+                ):
+                    self.crossing_state[tracker_id] = (class_ids, tracker_state)
```
LinasKo, line 265: Inconsistency here: second element was `dequeue`, but adding tracker_state which is `bool`.

(I'll figure it out, no need to reply :)

LinasKo:
Hey @grzegorz-roboflow, I made some heavy edits and think it's good to go.

You wouldn't happen to have the traffic video you used to test it, would you?

LinasKo:
Tested in a Colab: https://colab.research.google.com/drive/1rVuuJGPH-la2ExDmEUkKxwBnxRZMvgTK?usp=sharing

It works very similarly to how it was presented in Grzegorz's video.

image

Final results at different thresholds:

* threshold=1: in=12, out=65
* threshold=2: in=5,  out=59
* threshold=3: in=5,  out=56

LinasKo (approved):
Thank you very much @grzegorz-roboflow! At long last, we're adding it to the repo :)

Expect to see it in the new supervision version this weekend!

PR #1542, opened 2024-09-24
ci: 👷 pypi username - password swap with token

onuralpszr:
# Description

Pypi username / password has been replaced to with pypi token

### DIFF
--- a/.github/workflows/publish-test.yml
+++ b/.github/workflows/publish-test.yml
@@ -1,4 +1,4 @@
-name: Supervision Test Releases to PyPi
+name: Publish Supervision Pre-Releases to PyPI and TestPyPI
 on:
   push:
     tags:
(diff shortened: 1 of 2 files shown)


LinasKo:
As discussed privately, we'll instead set this up with Trusted Providers.

@onuralpszr , we now have the publishers set up on PyPi. We can proceed with changing the code.

onuralpszr:
> As discussed privately, we'll instead set this up with Trusted Providers.
> 
> @onuralpszr , we now have the publishers set up on PyPi. We can proceed with changing the code.

Single file or double file ?

LinasKo:
> Are we going to use single file or both files (change trigger control done via "release") ?

We've added 2 publishers. One for release via `publish.yml` and one for test, via `publish-test.yml`.

LinasKo:
Why do both files have a "🚀 Publish to Test-PyPi" and "🚀 Publish to PyPi"? Should it not be one for each?

onuralpszr:
> Why do both files have a "🚀 Publish to Test-PyPi" and "🚀 Publish to PyPi"? Should it not be one for each?

Current setup

pre-release versions goes to pypi and pypi-test server
release version goes to pypi and pypi-test server

LinasKo:
So publish-test isn't actually for test, but for pre-release. Can we add a comment in the file? This isn't automatically obvious for someone new to the system.

LinasKo (approved):
Approving prremptively with small changes incoming as agreed in a separate conversation

Issue #1543, opened 2024-09-25
[InferenceSlicer] Contradictory documentation regarding overlap_ratio_wh


tibeoh:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar bug report.

### Bug

`InferenceSlicer` object in the latest update. Specifically, the `overlap_ratio_wh` parameter is mentioned as both a new feature and deprecated in different sections of the release notes.

In the "Changed" section, it states:
> InferenceSlicer now features an `overlap_ratio_wh` parameter, making it easier to compute slice sizes when handling overlapping slices. [#1434](https://github.com/supervisely/supervisely/issues/1434)

However, in the "Deprecated" section, it mentions:
> `overlap_ratio_wh` in InferenceSlicer.__init__ is deprecated and will be removed in supervision-0.27.0. Use `overlap_wh` instead.

This seems contradictory: on one hand, `overlap_ratio_wh` is presented as a new feature, while on the other hand, it is marked for deprecation and removal in favor of `overlap_wh`.

Could you clarify whether `overlap_ratio_wh` should be used, or if we should transition directly to `overlap_wh`? Additionally, updating the documentation to make this distinction clear would help avoid confusion for future users.

Thank you!

### Environment

- Supervision 0.23.0
- Python 3.11

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

LinasKo:
Hi @tibeoh 👋 

Thanks for spotting it! That was my mistake — the new recommended parameter is `overlap_wh`, as computing the slices using the ratio was more complicated.

tibeoh:
Hi @LinasKo 

My PR is ready for review. 👍 

Please note that this is my first contribution to this project, so I hope everything is done correctly. I’m open to any feedback or suggestions for improvement.

LinasKo:
Thank you very much! That's a very high-quality contribution. Let's continue the discussion in the PR.

LinasKo:
Thank you for your contribution @tibeoh.
If you spot any other ares to improve, let us know!

PR #1547, opened 2024-09-26
fix: InferenceSlicer overlap_ratio_wh argument changed to None by default

tibeoh:
# Description

Issue #1543

Changed the default value of the `overlap_ratio_wh` argument to None, as this parameter will be deprecated in version 0.27.0. This change allows a smoother transition by not forcing users to explicitly set it to None.

⚠️ However, this may impact users who were relying on the default value of `overlap_ratio_wh` being `(0.2, 0.2)`. They will now need to explicitly specify this value if they still wish to use it, as the default is now set to None.

## Type of change

I have initiated a new test suite for this class (this is a first step). The tests cover validation of the overlap arguments (`overlap_wh` and `overlap_ratio_wh`) to ensure proper error handling and correct behavior. Additionally, initial tests for the `_generate_offset` method have been added to verify that slices are generated correctly based on the provided overlap settings.

N/A

## Docs

-   [ ] Docs updated? What were the changes: Default argument `overlap_ratio_wh` in `InferenceSlicer` class

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -96,7 +96,7 @@ detections = sv.Detections.from_sam(sam_result=sam_result)
 
 - Added [#1409](https://github.com/roboflow/supervision/pull/1409): `text_color` option for [`VertexLabelAnnotator`](https://supervision.roboflow.com/0.23.0/keypoint/annotators/#supervision.keypoint.annotators.VertexLabelAnnotator) keypoint annotator.
 
-- Changed [#1434](https://github.com/roboflow/supervision/pull/1434): [`InferenceSlicer`](https://supervision.roboflow.com/0.23.0/detection/tools/inference_slicer/) now features an `overlap_ratio_wh` parameter, making it easier to compute slice sizes when handling overlapping slices.
+- Changed [#1434](https://github.com/roboflow/supervision/pull/1434): [`InferenceSlicer`](https://supervision.roboflow.com/0.23.0/detection/tools/inference_slicer/) now features an `overlap_wh` parameter, making it easier to compute slice sizes when handling overlapping slices.
 
 - Fix [#1448](https://github.com/roboflow/supervision/pull/1448): Various annotator type issues have been resolved, supporting expanded error handling.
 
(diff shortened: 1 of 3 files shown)


LinasKo:
Hi @tibeoh,

It's very refreshing to see a thorough test suite together with the changes. Thank you for that!

I'd like to ask for a rollback of the deprecated parameter to `(0.2, 0.2)`. Yes, changing it to `None` makes the transition more seamless for new users, but keeping support of old values is non-negotiable. Who knows - maybe this is keeping a manufacturing line operational somewhere in the world.

Yes, this makes the deprecation a bit awkward. When the day comes, users will have an explicit `overlap_ratio_wh=None` argument in their code. That is a problem for another day, but we may declare a secondary deprecation stage, where we allow `overlap_ratio_wh=None`, but warn the users that it does nothing and should be removed.

LinasKo:
Also, would you mind creating a usage example in a Colab? You may use our [Starter Template](https://colab.research.google.com/drive/1rin7WrS-UvVIe-_Gfxmu-yVslGphOq89#scrollTo=pjmCrNre2g58). This eases our review considerably and creates a way for us to look back at how features worked in the past.

tibeoh:
Hi @LinasKo, 

Thank you for the feedback.

I completely agree with rolling back the default value of overlap_ratio_wh to (0.2, 0.2), and I will make that change shortly.

However, I currently don't have the time to work on the Colab example. Is it possible to proceed with the PR without the Colab for now, or would it be better to address it later?

Additionally, besides reverting the default value and the tests I’ve already added, do you expect any further changes? For example, I noticed that the current documentation could be a bit contradictory. Would you like me to edit that as well?

LinasKo:
This time - sure, let's skip the Colab. The change is small, and I can do a few tests on my end instead.

LinasKo:
If there's contradictions in the docs, feel free to address that! We'll accept all the help we can get.

Alternatively, if you find yourself lacking time, let me know where and I'll work on it when I find a spare moment.

LinasKo:
Hi @tibeoh 👋

We'd love to have this fix in the new release - there are a few outstanding changes, but I'll take over and get it merged.

Thank you for your contribution!

LinasKo (approved):

tibeoh:
Hi @LinasKo 
I was very busy, so thank you for finalizing my PR. I’ll try to submit new PRs in the future if I have any proposals for this library.

PR #1550, opened 2024-09-26
feat: add __len__ to ColorPalette

ethanwhite:
# Description

For downstream use being able to determine how many colors are in a ColorPalette object can be useful. E.g., checking if a user has provided a color palette that matches the number of labels. ColorPalette doesn't currenly have a __len__ attribute meaning that `mypalette.colors` has to be used for calculating lengths. This change adds `__len__` to make this more intutitive by allowing `len(mypalette)`.

We found the lack of this functionality confusing: https://github.com/weecology/DeepForest/issues/790

## Type of change

I built the package with the change and confirmed that `len(mypalette)` returns the appropriate result

## Docs

No associate docs changes

### DIFF
--- a/supervision/draw/color.py
+++ b/supervision/draw/color.py
@@ -397,6 +397,15 @@ class ColorPalette:
         idx = idx % len(self.colors)
         return self.colors[idx]
 
+    def __len__(self) -> int:
+        """
+        Returns the number of colors in the palette.
+
+        Returns:
+            int: The number of colors.
+        """
+        return len(self.colors)
+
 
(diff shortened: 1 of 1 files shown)


LinasKo (approved):
Perfectly reasonable. Thank you for the contribution!

The change will be released on Friday next week.
Meanwhile, you can install the develop version of supervision with:
```bash
pip install git+https://github.com/roboflow/supervision.git@develop
```

It feels great to see us powering projects like yours. Let me know if you ever need any help with Roboflow libraries.

ethanwhite:
Thanks @LinasKo! We love the work you all are doing and will definitely let you know if we need any help going forward.

Issue #1585, opened 2024-10-09
Detections Metadata


LinasKo:
# Detections Metadata

> [!TIP]
> [Hacktoberfest](https://hacktoberfest.com/) is calling! Whether it's your first PR or your 50th, you’re helping shape the future of open source. Help us build the most reliable and user-friendly computer vision library out there! 🌱

This is a summarized version of #1226.
See the original discussion for more context.

I expect this to be a hard issue, as it involves many elements within the library.

---

`Detections` is a class for encoding the results of any model - detection, segmentation, etc. Inside it contains multiple arrays: `xyxy`, `mask`, `class_id`, `tracker_id`, `confidence`, as well as a `data` value for anything user-defined. All of these, as well as `data` contents are either:
1. `None`: Model will never detect anything of that field
2. `empty array`: Model detected no elements this time.
3. Array of N elements: N objects were detected in your image. 

For historical reasons, `Detections.empty()` and `Detections.is_empty()` are an exception. A model may return an empty set of detection and not match the None/Array/ArrayN definitions of successful detections. Therefore, empty detections are mostly checked with `Detections.is_empty()` and treated as a special case.

If one of these elements is an `N`-long array, all other elements that aren't `None`, will also have length `N`. This includes all values of `data`. What if we want to store 1 value per-list? E.g. a video name for what every detection was extracted from? Or camera parameters?

Let's introduce a new field: `metadata`. Detections will now look as follows:

```
@dataclass
class Detections:
    xyxy: np.ndarray
    mask: Optional[np.ndarray] = None
    confidence: Optional[np.ndarray] = None
    class_id: Optional[np.ndarray] = None
    tracker_id: Optional[np.ndarray] = None
    data: Dict[str, Union[np.ndarray, List]] = field(default_factory=dict)
    metadata: Dict[str, Any] = field(default_factory=dict)
```

The users can set it directly by doing `detections.metadata["key"] = "val"`.

The primary complexity is caused by functions that merge, slice, split and index into detections.

Relevant methods to be updated:
* `__eq__` should use metadata for comparison
* `is_empty` should borrow the data for comparison, just like it does with `data`.
* `__iter__`, **should NOT** return metadata.
* None of the `from_...` methods should be affected
* `merge` should:
  * retain metadata even when all detections passed were empty.
  * call a new `merge_metadata` function.
[...]

PR #1589, opened 2024-10-10
Feat: Added detection metadata

souhhmm:
# Description

As per,
- #1585 

implemented detection metadata and updated related methods. Please let me know if anything needs to be fixed/changed.

## Type of change

-   [x] This change requires a documentation update (probably)

Here's the [link to the colab](https://colab.research.google.com/drive/1HaK7SUhFTsHgsi4QUcf6FTSe70SLstGj?usp=sharing) that tests the implemented code.

## Additional Notes

In the current implementation, `is_empty` returns True even when a detection has a non empty metadata. When two empty detections with metadata are merged, the metadata is retained in the merge, with everything else being empty. If this is not how it is intended to work, simply adding `empty_detections.metadata = self.metadata` in `is_empty` retains metadata in the merge, however, this returns False when metadata is non empty (with everything else being empty).


souhhmm:
Hey @LinasKo, it says the 'Welcome Workflow' check has failed. Is this the reason the review has been halted?

LinasKo:
It's a mistake on our part. Don't worry about that one :wink:

souhhmm:
Hey @LinasKo, any updates/feedback?

LinasKo:
On it, on it. This one will take time, likely the whole week. It can have consequences across the repo, so I must be very careful.

souhhmm:
No worries, take your time.

supervision/detection/core.py
```diff
@@ -977,15 +983,21 @@ def empty(cls) -> Detections:
...
-        empty_detections.data = self.data
-        return self == empty_detections
+        return (
+            len(self.xyxy) == 0
+            and (self.mask is None or len(self.mask) == 0)
+            and (self.class_id is None or len(self.class_id) == 0)
+            and (self.confidence is None or len(self.confidence) == 0)
+            and (self.tracker_id is None or len(self.tracker_id) == 0)
+            and not self.data
+        )
```
LinasKo, line 1000: This widens the description of what `empty` is allowed to be. Let's keep it as it was. It'll get simplified when we solve #943

supervision/detection/core.py
```diff
@@ -133,6 +137,7 @@ class simplifies data manipulation and filtering, providing a uniform API for
     class_id: Optional[np.ndarray] = None
     tracker_id: Optional[np.ndarray] = None
     data: Dict[str, Union[np.ndarray, List]] = field(default_factory=dict)
+    metadata: Dict[str, Any] = field(default_factory=dict)
 
     def __post_init__(self):
         validate_detections_fields(
```
LinasKo, line 143: In `Detections.empty()` you set the metadata to `metadata=metadata if metadata is not None else`. Let's add some validation, making sure it's a `dict`.

supervision/detection/core.py
LinasKo: I like the optional metadata argument for `empty`, but let's avoid checking for it here (similarly to how we avoided `data` check).

Suppose there is a model that returns values `{detection, X, Y}` when detections are successful and `{[], X}` when nothing is detected. This prevents the developers from using the metadata system if they ever want to add `Y`.

Ultimately, I'd like to simplify this to `len(self.xyxy) == 0`, but only after the next release, to have time for testing.

supervision/detection/core.py
```diff
@@ -185,6 +190,7 @@ def __eq__(self, other: Detections):
                 np.array_equal(self.confidence, other.confidence),
                 np.array_equal(self.tracker_id, other.tracker_id),
                 is_data_equal(self.data, other.data),
+                self.metadata == other.metadata,
```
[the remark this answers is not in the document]
LinasKo: I suggest creating another function `is_metadata_equal` which checks that keys are equal. Then, for each value, if it's a numpy array, it checks with `np.array_equal` and then normal `==`.

In later PRs we could also call it recursively if it's a dict, but maybe let's not do that yet.

supervision/detection/utils.py
```diff
@@ -866,6 +866,35 @@ def merge_data(
     return merged_data
 
 
+def merge_metadata(metadata_list: List[Dict[str, Any]]) -> Dict[str, Any]:
```
LinasKo, line 869: Contrary to how leniently we treat `Detections.empty`, let's be more aggressive here.

Using similar logic to `__eq__`, let's make sure that for all non-empty metadata, all the keys and values are the same.

If it's different, let's raise an error.

LinasKo: My goal is to tighten new data structures, while keeping old ones, with known issues, more lenient.

LinasKo (changes requested):

LinasKo:
I've made some comments. I like your changes so far!

It shows you've thought through the problem, in places like introducing of optional variable in `empty`, while avoiding the pitfall of incorrectly defaulting to `{}`.

souhhmm:
Hey @LinasKo, I have made the required changes. Please let me know about further changes/fixes. Thanks :)

LinasKo:
That looks good.

The next request for every contribution is to create a Colab to test the changes. This would involve creating Detections with and without metadata showing how it works & testing with multiple values. It's slightly easier than unit testing, and is also a doc we refer to later on, when we wish to test for regressions.

Got any time for that, @souhhmm? There's a [Starter Template](https://colab.research.google.com/drive/1rin7WrS-UvVIe-_Gfxmu-yVslGphOq89#scrollTo=pjmCrNre2g58) you may use 😉

souhhmm:
I had linked a colab in the og PR, [linking once again](https://colab.research.google.com/drive/1HaK7SUhFTsHgsi4QUcf6FTSe70SLstGj?usp=sharing). But now that I have changed some code, I'll re-run it.

LinasKo:
Hi @souhhmm 👋 

My apologies, I missed your Colab. That is indeed an amazing set of tests.
I'll take over from this point - there's a few checks I need to do before I merge this.

souhhmm:
No worries, thanks to you too for your great feedback :)

LinasKo:
Thank you for a fantastic contribution @souhhmm!

Most take me hours to verify, but with your test suite and thoughtful changes, I barely had to do anything. Expect to see it in the next supervision release at the end of this week!

LinasKo (approved):

PR #1595, opened 2024-10-14
feat: 🚀 python3.13 support for poetry and library

onuralpszr:
# Description

Proper python3.13 support configuration and scipy, numpy version enforced to use supported py3.13 package

### DIFF
--- a/.github/workflows/poetry-test.yml
+++ b/.github/workflows/poetry-test.yml
@@ -1,0 +1,37 @@
+name: 🔧 Poetry Check and Installation Test Workflow
+on:
+  push:
+    paths:
+      - 'poetry.lock'
+      - 'pyproject.toml'
+  pull_request:
+    paths:
+      - 'poetry.lock'
+      - 'pyproject.toml'
+  workflow_dispatch:
+
+jobs:
+  poetry-tests:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: 📥 Checkout the repository
+        uses: actions/checkout@v4
+
+      - name: 🐍 Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: 📦 Install the base dependencies
+        run: python -m pip install --upgrade poetry
+
+      - name: 🔍 Check the correctness of the project config
+        run: poetry check
+
+      - name: 🚀 Do Install the package Test
+        run: poetry install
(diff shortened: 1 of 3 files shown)


LinasKo:
Thanks @onuralpszr! This looks like it required a lot of tweaking to get working. Good work!

LinasKo:
Out of curiosity, how do lock files react to conditional installs? Do they encode the conditional logic, or do we have a snapshot of specific requirements for a specific python version?

onuralpszr:
> Out of curiosity, how do lock files react to conditional installs? Do they encode the conditional logic, or do we have a snapshot of specific requirements for a specific python version?

Lock files, such as poetry.lock, capture a snapshot of the exact versions of dependencies that are installed, including any conditional dependencies. They do not encode the conditional logic itself but rather the resolved dependencies for the environment in which the lock file was generated.

LinasKo:
What happens when we install supervision when we have 3.13, 3.10 vs 3.8 installed?

Based on the previous answer, it seems it will install exact version if the version matches what's available in the lock file, but resolve the dependencies anew (and generate a new file) if not. Is that true?

LinasKo:
Also, I see some CI fails, so I'm, not going to bother you in case you still need to fix things 😄

onuralpszr:
> Also, I see some CI fails, so I'm, not going to bother you in case you still need to fix things 😄

CI is weird :) It suppose to work but for some magical reason macos 3.8 and 3.13 decided to not play with me but I will handle it :)

onuralpszr:
@LinasKo I added contourpy as new dependency but it is already sub-dependency of Matplotlib and my description to toml like this.

```toml
# Matplotlib sub-dependency
# The 'contourpy' package is required by Matplotlib for contour plotting.
# We need to ensure compatibility with both Python 3.8 and Python 3.13.
#
# For Python 3.8 and above, we use version 1.0.7 or higher, as it is the lowest major version that supports Python 3.8.
# For Python 3.13 and above, we use version 1.3.0 or higher, as it is the first version that explicitly supports Python 3.13.
contourpy = [
    { version = ">=1.0.7", python = ">=3.8" },
    { version = ">=1.3.0", python = ">=3.13" },
]
```

Last action run: https://github.com/roboflow/supervision/actions/runs/11326558737/job/31495751852?pr=1595
My last commit force just adding comments so It won't effect out come of last one.

LinasKo:
I had some issues with building `matplotlib`, which were solved by updating xcode via the standard system update. Tests pass, docs seem unaffected.

Currently neither `ultralytics` nor `inference` support 3.13, so I cannot test thoroughly.

Approving - let's see how other packages develop and fix anything new that arises.

LinasKo (approved):

LinasKo:
I may revert this. Upon merging and setting up a new repo, I immediately got an issue, failing to install scipy. I'll take a bit of time to confirm.

onuralpszr:
> I may revert this. Upon merging and setting up a new repo, I immediately got an issue, failing to install scipy. I'll take a bit of time to confirm.

Okay let me know what I can do

LinasKo:
@onuralpszr False alarm - I think I found an issue in my PR cloning tool :)

Issue #1582, opened 2024-10-09
Allow ByteTrack to track detections without IDs


LinasKo:
# Allow ByteTrack to track detections without IDs

> [!TIP]
> [Hacktoberfest](https://hacktoberfest.com/) is calling! Whether it's your first PR or your 50th, you’re helping shape the future of open source. Help us build the most reliable and user-friendly computer vision library out there! 🌱

For a long time, all of models we support would return a `class_id` which we set inside Detections (`class_id` is an array of N integers). However, we've seen more and more models where that isn't provided.

[ByteTrack](https://supervision.roboflow.com/latest/trackers/#bytetrack) allows you to track objects in the scene. It uses the class and location of the detection to figure out where it is. It can track objects where all class IDs are the same. However, if you try passing something without an ID, it will fail. See [this Colab](https://colab.research.google.com/drive/1oAi30puYa7ZgJIAYCe2E_-7Q_dCap4Sr?usp=sharing).

Let's change ByteTrack so it can track detections, even if they didn't have a `class_id`. I see two ways of doing this:
1) Change the algorithm such that class_ids are not required
2) Set an internal class ID such as `-1`.

As a result, I expect the videos produced in the [Colab](https://colab.research.google.com/drive/1oAi30puYa7ZgJIAYCe2E_-7Q_dCap4Sr?usp=sharing) to be visually indistinguishable.

It would also be useful to test on a video where some frames feature no detections (you can simulate with `detections = sv.Detections.empty()`)

Also, check what happens when some detections have `class_id` and some do not.

---

Helpful links:
* [Contribution guide](https://supervision.roboflow.com/develop/contributing/#how-to-contribute-changes)
* ByteTrack: [docs](https://supervision.roboflow.com/latest/trackers/#bytetrack), [code](https://github.com/roboflow/supervision/blob/70869c73379026614c651828123f35e771f65e67/supervision/tracker/byte_tracker/core.py#L190)
* Detections: [docs](https://supervision.roboflow.com/latest/detection/core/), [code](https://github.com/roboflow/supervision/blob/develop/supervision/detection/core.py)
* [Supervision Cheatsheet](https://roboflow.github.io/cheatsheet-supervision/)
* [Colab Starter Template](https://colab.research.google.com/drive/1rin7WrS-UvVIe-_Gfxmu-yVslGphOq89#scrollTo=pjmCrNre2g58)
[...]

PR #1618, opened 2024-10-23
feat(ByteTrack): Allow ByteTrack to track detection without class ids

Kadermiyanyedi:
**_Description:_**
This PR modifies ByteTrack to handle detections that do not include a class_id, improving the flexibility of the tracking algorithm. In cases where models don’t return a class_id, ByteTrack will now assign a default internal ID to ensure continued tracking.

**_Key Changes:_**
Default class_id Assignment:

- If class_id is missing from the detection, a default value of -5 is assigned to all detections.
- If class_id is available, the algorithm uses it as normal.

This ensures that ByteTrack works seamlessly with models that do not provide class_id, without breaking existing functionality.

You can test the functionality with the provided [Colab Notebook](https://colab.research.google.com/drive/1SXZndmumMotpHobM3kTefuv2pkggyy_R?usp=sharing) which demonstrates ByteTrack handling cases with and without class_id.

#1582

### DIFF
--- a/supervision/tracker/byte_tracker/core.py
+++ b/supervision/tracker/byte_tracker/core.py
@@ -276,11 +276,16 @@ class ByteTrack:
             ```
         """
 
+        num_rows = detections.xyxy.shape[0]
+        class_ids = np.full(num_rows, -5)
+        if detections.class_id is not None:
+            class_ids = detections.class_id
+
         tensors = np.hstack(
             (
                 detections.xyxy,
                 detections.confidence[:, np.newaxis],
-                detections.class_id[:, np.newaxis],
+                class_ids[:, np.newaxis],
             )
         )
         tracks = self.update_with_tensors(tensors=tensors)


LinasKo:
Hey @Kadermiyanyedi 👋 

Did you try checking what happens with the `class_id` after it's given to ByteTrack? I'm still super curious - maybe we can remove it altogether.

Is that something you'd have time for?

LinasKo:
I suggested a special value other than `-1` myself, but it does feel a bit arbitrary.

However, I'll take an unconventional approach and merge the PR, in hopes of finding a way to remove the need for classes entirely. If that fails, I'll come back and change the special value to something else.

Thank you @Kadermiyanyedi! Expect to see this in the next supervision release 😉

LinasKo (approved):

Kadermiyanyedi:
@LinasKo Hi, due to the busy schedule, I missed your comment on the PR and forgot to reply—sorry about that. I tested your suggestion and confirmed that class IDs are not necessary; it works without them. I’m opening a PR for this right away. Would it be possible to revert the PR we merged?

LinasKo:
✅ Reverted

PR #1625, opened 2024-10-28
Added smart positioning (non overlapping labels) to VertexLabelAnnotator

kshitijaucharmal:
# Description

Resolved issue #1383 regarding overlapping labels in output.
Changes Done in VertexLabelAnnotator, LabelAnnotator and RichLabelAnnotator
No Extra dependencies used

## Type of change

Have added an optional argument to the VertexLabelAnnotator called `use_smart_positioning` which defaults to `False`.
Enabling this prevents overlapping.

Before:

https://github.com/user-attachments/assets/2906efd5-2a88-4ee0-ba86-cfb250d3fc8f

After:

VertexLabelAnnotator:

https://github.com/user-attachments/assets/5c4bf897-a695-4b88-962e-217f881248fe

Label Annotator:

https://github.com/user-attachments/assets/8ded0d84-cca0-4c05-8504-f1478ac20771

Rich Label Annotator (font: [Vampire Wars](https://www.dafont.com/vampire-wars.font?back=theme)):

https://github.com/user-attachments/assets/4065dd38-0151-4f89-b5b7-b0b613431608

## Docs

-  Changes to docstrings in:
1. VertexLabelAnnotator
2. LabelAnnotator
3. RichLabelAnnotator

Collab Link: https://colab.research.google.com/drive/1GPHr7PpZ_eNs8Gkxx_A4w6PL5p4pJAn1?usp=sharing#scrollTo=ZHptw0Fk4lk4

### DIFF
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -1139,9 +1139,9 @@ class LabelAnnotator(BaseAnnotator):
 
         labels = self._get_labels_text(detections, labels)
         label_properties = self._get_label_properties(detections, labels)
-        xyxy = label_properties[:, :4]
 
         if self.smart_position:
+            xyxy = label_properties[:, :4]
             xyxy = spread_out_boxes(xyxy)
             label_properties[:, :4] = xyxy
 
@@ -1412,9 +1412,9 @@ class RichLabelAnnotator(BaseAnnotator):
         draw = ImageDraw.Draw(scene)
         labels = self._get_labels_text(detections, labels)
         label_properties = self._get_label_properties(draw, detections, labels)
-        xyxy = label_properties[:, :4]
 
         if self.smart_position:
+            xyxy = label_properties[:, :4]
             xyxy = spread_out_boxes(xyxy)
             label_properties[:, :4] = xyxy
 
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -1,4 +1,3 @@
-from dataclasses import dataclass
 from functools import lru_cache
 from math import sqrt
 from typing import List, Optional, Tuple, Union
@@ -1047,14 +1046,6 @@ class LabelAnnotator(BaseAnnotator):
     A class for annotating labels on an image using provided detections.
     """
 
-    @dataclass
-    class _TextProperties:
-        text: str
-        width: int
-        height: int
-        width_padded: int
-        height_padded: int
-
     def __init__(
         self,
         color: Union[Color, ColorPalette] = ColorPalette.DEFAULT,
@@ -1147,19 +1138,17 @@ class LabelAnnotator(BaseAnnotator):
         self._validate_labels(labels, detections)
 
         labels = self._get_labels_text(detections, labels)
-        text_properties = self._get_text_properties(labels)
-
-        xyxy = self._calculate_label_positions(
-            detections, text_properties, self.text_anchor
-        )
+        label_properties = self._get_label_properties(detections, labels)
+        xyxy = label_properties[:, :4]
 
         if self.smart_positions:
             xyxy = spread_out_boxes(xyxy)
+            label_properties[:, :4] = xyxy
 
         self._draw_labels(
             scene=scene,
-            xyxy=xyxy,
-            text_properties=text_properties,
+            labels=labels,
+            label_properties=label_properties,
             detections=detections,
             custom_color_lookup=custom_color_lookup,
         )
@@ -1174,11 +1163,17 @@ class LabelAnnotator(BaseAnnotator):
                 f"should have exactly 1 label."
             )
 
-    def _get_text_properties(self, labels: List[str]) -> List[_TextProperties]:
-        """Gets text content and dimensions for all detections."""
-        text_properties = []
+    def _get_label_properties(
+        self,
+        detections: Detections,
+        labels: List[str],
+    ) -> np.ndarray:
+        label_properties = []
+        anchors_coordinates = detections.get_anchors_coordinates(
+            anchor=self.text_anchor
+        ).astype(int)
 
-        for label in labels:
+        for label, center_coords in zip(labels, anchors_coordinates):
             (text_w, text_h) = cv2.getTextSize(
                 text=label,
                 fontFace=CV2_FONT,
@@ -1186,17 +1181,23 @@ class LabelAnnotator(BaseAnnotator):
                 thickness=self.text_thickness,
             )[0]
 
-            text_properties.append(
-                self._TextProperties(
-                    text=label,
-                    width=text_w,
-                    height=text_h,
-                    width_padded=text_w + 2 * self.text_padding,
-                    height_padded=text_h + 2 * self.text_padding,
-                )
+            width_padded = text_w + 2 * self.text_padding
+            height_padded = text_h + 2 * self.text_padding
+
+            text_background_xyxy = resolve_text_background_xyxy(
+                center_coordinates=tuple(center_coords),
+                text_wh=(width_padded, height_padded),
+                position=self.text_anchor,
             )
 
-        return text_properties
+            label_properties.append(
+                [
+                    *text_background_xyxy,
+                    text_h,
+                ]
+            )
+
+        return np.array(label_properties).reshape(-1, 5)
 
     @staticmethod
     def _get_labels_text(
@@ -1215,41 +1216,17 @@ class LabelAnnotator(BaseAnnotator):
                 labels.append(str(idx))
         return labels
 
-    def _calculate_label_positions(
-        self,
-        detections: Detections,
-        text_properties: List[_TextProperties],
-        text_anchor: Position,
-    ) -> np.ndarray:
-        anchors_coordinates = detections.get_anchors_coordinates(
-            anchor=text_anchor
-        ).astype(int)
-
-        xyxy = []
-        for idx, center_coords in enumerate(anchors_coordinates):
-            text_background_xyxy = resolve_text_background_xyxy(
-                center_coordinates=tuple(center_coords),
-                text_wh=(
-                    text_properties[idx].width_padded,
-                    text_properties[idx].height_padded,
-                ),
-                position=text_anchor,
-            )
-            xyxy.append(text_background_xyxy)
-
-        return np.array(xyxy)
-
     def _draw_labels(
         self,
         scene: np.ndarray,
-        xyxy: np.ndarray,
-        text_properties: List[_TextProperties],
+        labels: List[str],
+        label_properties: np.ndarray,
         detections: Detections,
         custom_color_lookup: Optional[np.ndarray],
     ) -> None:
-        assert len(xyxy) == len(text_properties) == len(detections), (
-            f"Number of text properties ({len(text_properties)}), "
-            f"xyxy ({len(xyxy)}) and detections ({len(detections)}) "
+        assert len(labels) == len(label_properties) == len(detections), (
+            f"Number of label properties ({len(label_properties)}), "
+            f"labels ({len(labels)}) and detections ({len(detections)}) "
             "do not match."
         )
 
@@ -1259,7 +1236,7 @@ class LabelAnnotator(BaseAnnotator):
             else self.color_lookup
         )
 
-        for idx, box_xyxy in enumerate(xyxy):
+        for idx, label_property in enumerate(label_properties):
             background_color = resolve_color(
                 color=self.color,
                 detections=detections,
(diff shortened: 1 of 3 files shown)


LinasKo:
Accepting this based on the videos. I sense there's a lot of reviewing to do 🙂

Would you be able to create a Colab showcasing the usage?

kshitijaucharmal:
Here's the collab link: https://colab.research.google.com/drive/1GPHr7PpZ_eNs8Gkxx_A4w6PL5p4pJAn1?usp=sharing#scrollTo=ZHptw0Fk4lk4

LinasKo:
Improved Colab: https://drive.google.com/file/d/1PYoqJDdPfJaoogvEIM7xSIxhQLbXs3QR/view?usp=sharing

LinasKo (approved):

supervision/detection/utils.py
```diff
@@ -1039,3 +1039,60 @@ def cross_product(anchors: np.ndarray, vector: Vector) -> np.ndarray:
...
+        np.fill_diagonal(iou, 0)
+
+        if np.all(iou == 0):
+            break
+
+        i, j = np.unravel_index(np.argmax(iou), iou.shape)
+
+        xyxy_i, xyxy_j = xyxy_padded[i], xyxy_padded[j]
+        box_intersection = get_box_intersection(xyxy_i, xyxy_j)
+        assert (
```
SkalskiP, line 1082: Why do we check that condition if it should never happen?

supervision/detection/utils.py
```diff
@@ -1039,3 +1039,60 @@ def cross_product(anchors: np.ndarray, vector: Vector) -> np.ndarray:
...
+        ), "Since we checked IoU already, boxes should always intersect"
+
+        intersection_center = (box_intersection[:2] + box_intersection[2:]) / 2
+        xyxy_i_center = (xyxy_i[:2] + xyxy_i[2:]) / 2
+        xyxy_j_center = (xyxy_j[:2] + xyxy_j[2:]) / 2
+
+        unit_vector_i = get_unit_vector(intersection_center, xyxy_i_center)
+        unit_vector_j = get_unit_vector(intersection_center, xyxy_j_center)
+
+        xyxy_padded[i, [0, 2]] += int(unit_vector_i[0] * step)
```
SkalskiP, line 1093: We have `move_boxes` util. Looks like we could replace:

```python
xyxy_padded[i, [0, 2]] += int(unit_vector_i[0] * step)
xyxy_padded[i, [1, 3]] += int(unit_vector_i[1] * step)
```

With:

```python
offset=(unit_vector_i * step).astype(int)
xyxy_padded = move_boxes(xyxy=xyxy_padded, offset=offset)
```

I find it a lot more readable.

supervision/detection/utils.py
```diff
@@ -1039,3 +1039,60 @@ def cross_product(anchors: np.ndarray, vector: Vector) -> np.ndarray:
...
+
+    xyxy_padded = pad_boxes(xyxy, px=step)
+    for _ in range(max_iterations):
+        iou = box_iou_batch(xyxy_padded, xyxy_padded)
+        np.fill_diagonal(iou, 0)
+
+        if np.all(iou == 0):
+            break
+
+        i, j = np.unravel_index(np.argmax(iou), iou.shape)
```
SkalskiP, line 1078: Why are we processing only one box pair (the one with the greatest overlap) per iteration? It seems like it should be possible to process all overlapping boxes in parallel.

supervision/annotators/core.py
```diff
@@ -1128,84 +1142,156 @@ def annotate(
...
-        ).astype(int)
+        self._validate_labels(labels, detections)
+
+        labels = self._get_labels_text(detections, labels)
+        text_properties = self._get_text_properties(labels)
+
+        xyxy = self._calculate_label_positions(
+            detections, text_properties, self.text_anchor
+        )
+
```
SkalskiP, line 1155: It seems to me that `_get_text_properties` and `_calculate_label_positions` should be merge into a single function that returns numpy array. You can than cut out `xyxy` portion of that numpy array and pass it to `spread_out_boxes`.

supervision/annotators/core.py
```diff
@@ -1346,88 +1438,169 @@ def annotate(
...
-        anchors_coordinates = detections.get_anchors_coordinates(
-            anchor=self.text_anchor
-        ).astype(int)
+        labels = self._get_labels_text(detections, labels)
+        text_properties = self._get_text_properties(draw, labels)
+
+        xyxy = self._calculate_label_positions(
+            detections, text_properties, self.text_anchor
+        )
+
```
SkalskiP, line 1450: It seems to me that `_get_text_properties` and `_calculate_label_positions` should be merge into a single function that returns numpy array. You can than cut out `xyxy` portion of that numpy array and pass it to `spread_out_boxes`.

SkalskiP (changes requested):

supervision/annotators/core.py
```diff
@@ -1044,6 +1045,16 @@ class LabelAnnotator(BaseAnnotator):
     A class for annotating labels on an image using provided detections.
     """
 
+    @dataclass
+    class _TextProperties:
```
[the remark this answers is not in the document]
LinasKo: I'll change it, but in terms of your question:

Arrays only carry info about the datatype, but not what the values are. Generally, when arrays are used to pass multiple different values, one needs to find all places it's been modified to learn about its contents.

Therefore, a `struct` or in Python's case, a `dataclass` or `namedtuple` makes more sense.

supervision/detection/utils.py
LinasKo: I'm deliberately using an `assert` instead of a `ValueError` since this is documents my assumption. I can't come up with a reason for why it would fail now, but if it does, the following code would be invalid. In my experience, more than half of these and up triggered eventually, despite of how certain I am about the outcome.

As a side-effect it calms down MyPy, and I'm keeping it. But that's not the main reason it's here.

supervision/detection/utils.py
LinasKo: I didn't think about it - good point.

However, the current algo moves one box at a time, and your suggestion would be part of the vectorized approach.

My plan for this was to not entirely rewrite the contributor's code, but to open a second issue for the community, adding vectorization as a second step.

supervision/annotators/core.py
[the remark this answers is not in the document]
LinasKo: addressed, removed my struct, merged functions (definitely a good idea!), and used numpy for passing data (feels very wrong). Tested for regressions - all is well.

supervision/annotators/core.py
```diff
@@ -1256,6 +1342,16 @@ class RichLabelAnnotator(BaseAnnotator):
     with support for Unicode characters by using a custom font.
     """
 
+    @dataclass
+    class _TextProperties:
```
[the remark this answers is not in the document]
LinasKo: addressed, removed my struct, merged functions (definitely a good idea!), and used numpy for passing data (feels very wrong). Tested for regressions - all is well.

I expect we could tidy up and remove the labels argument, but leaving that for now.

supervision/annotators/core.py
LinasKo: addressed, removed my struct, merged functions (definitely a good idea!), and used numpy for passing data (feels very wrong). Tested for regressions - all is well.

LinasKo:
New Colab: https://drive.google.com/file/d/1o8rVh2ngJIf8Oxd_kotL6solRKMhDSjt/view?usp=sharing

LinasKo (approved):

supervision/annotators/core.py
```diff
@@ -1054,6 +1056,7 @@ def __init__(
         text_position: Position = Position.TOP_LEFT,
         color_lookup: ColorLookup = ColorLookup.CLASS,
         border_radius: int = 0,
+        smart_positions: bool = False,
```
SkalskiP, line 1059: Let's drop `s` and rename it to `smart_position` as we have other arguments like `text_position`.

supervision/annotators/core.py
```diff
@@ -1266,6 +1331,7 @@ def __init__(
...
```
SkalskiP, line 1334: Let's drop `s` and rename it to `smart_position` as we have other arguments like `text_position`.

supervision/keypoint/annotators.py
```diff
@@ -201,6 +202,7 @@ def __init__(
         text_thickness: int = 1,
         text_padding: int = 10,
         border_radius: int = 0,
+        smart_positions: bool = False,
```
SkalskiP, line 205: Let's drop `s` and rename it to `smart_position` as we have other arguments like `text_position`.

supervision/annotators/core.py
```diff
@@ -1128,84 +1133,144 @@ def annotate(
...
         assert isinstance(scene, np.ndarray)
-        font = cv2.FONT_HERSHEY_SIMPLEX
-        anchors_coordinates = detections.get_anchors_coordinates(
-            anchor=self.text_anchor
-        ).astype(int)
+        self._validate_labels(labels, detections)
+
+        labels = self._get_labels_text(detections, labels)
+        label_properties = self._get_label_properties(detections, labels)
+        xyxy = label_properties[:, :4]
```
SkalskiP, line 1142: That line is only needed if `smart_position` is `True`

supervision/annotators/core.py
```diff
@@ -1128,84 +1133,144 @@ def annotate(
...
-
-            if labels is not None:
-                text = labels[detection_idx]
-            elif CLASS_NAME_DATA_FIELD in detections.data:
-                text = detections.data[CLASS_NAME_DATA_FIELD][detection_idx]
-            elif detections.class_id is not None:
-                text = str(detections.class_id[detection_idx])
-            else:
-                text = str(detection_idx)
+    def _get_label_properties(
```
SkalskiP, line 1166: It is a lot cleaner! 🔥 I know we usually do not include docstrings for private / protected methods but I'd do it this time.

supervision/detection/utils.py
```diff
@@ -1039,3 +1039,75 @@ def cross_product(anchors: np.ndarray, vector: Vector) -> np.ndarray:
     )
     vector_start = np.array([vector.start.x, vector.start.y])
     return np.cross(vector_at_zero, anchors - vector_start)
+
+
+def get_box_intersection(
```
SkalskiP, line 1044: this function is no longer used

supervision/detection/utils.py
```diff
@@ -1039,3 +1039,75 @@ def cross_product(anchors: np.ndarray, vector: Vector) -> np.ndarray:
...
+    overlap_xmax = min(xyxy_1[2], xyxy_2[2])
+    overlap_ymax = min(xyxy_1[3], xyxy_2[3])
+
+    if overlap_xmin < overlap_xmax and overlap_ymin < overlap_ymax:
+        return np.array([overlap_xmin, overlap_ymin, overlap_xmax, overlap_ymax])
+    else:
+        return None
+
+
+def get_unit_vector(xy_1: np.ndarray, xy_2: np.ndarray) -> np.ndarray:
```
SkalskiP, line 1058: this function is no longer used

supervision/detection/utils.py
```diff
@@ -1039,3 +1039,75 @@ def cross_product(anchors: np.ndarray, vector: Vector) -> np.ndarray:
...
+
+        # Nx2
+        centers = (xyxy_padded[:, :2] + xyxy_padded[:, 2:]) / 2
+
+        # NxNx2
+        delta_centers = centers[:, np.newaxis, :] - centers[np.newaxis, :, :]
+        delta_centers *= overlap_mask[:, :, np.newaxis]
+
+        # Nx2
+        force_vectors = np.sum(delta_centers, axis=1)
```
SkalskiP, line 1099: I was playing with some visualization ideas and noticed some strange behavior in "smart mode." Small changes in box position can lead to large changes in label position. As a result, the overall effect can feel choppy.

I'm afraid math breaks here. :/ 

The movement vector should be proportional to the IoU - the more the boxes overlap, the more we need to pull them apart. Instead, the force with which we push them apart is proportional to the distance between their centers. If the centers are further apart, the boxes are pushed apart more strongly.

SkalskiP (changes requested):

LinasKo (approved):

LinasKo:
I've witnessed that the new algo sometimes produces more jerky motions of the label when applied to the video. I pushed the most stable combination I could find, but there's definitely room for improvement.

Merging this in! Thank you very much, @kshitijaucharmal, @SkalskiP!

Issue #1554, opened 2024-09-28
Allow TIFF (and more) image formats in `load_yolo_annotations`


patel-zeel:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Description

* Currently, `load_yolo_annotations` only allows `png,` `jpg`, and `jpeg` file formats. `load_yolo_annotations` is internally called by `sv.DetectionDataset.from_yolo` functionality.

https://github.com/roboflow/supervision/blob/1860fdb0a4e21edc5fa03d973e9f31c055bdcf4f/supervision/dataset/formats/yolo.py#L156

* Ultralytics supports a wide variety of image formats. Copied the following table from [their website](https://docs.ultralytics.com/modes/predict/#image-and-video-formats:~:text=The%20below%20table%20contains%20valid%20Ultralytics%20image%20formats.).

| Image Suffix  | Example Predict Command           | Reference                       |
|---------------|-----------------------------------|---------------------------------|
| .bmp          | yolo predict source=image.bmp     | Microsoft BMP File Format       |
| .dng          | yolo predict source=image.dng     | Adobe DNG                       |
| .jpeg         | yolo predict source=image.jpeg    | JPEG                            |
| .jpg          | yolo predict source=image.jpg     | JPEG                            |
| .mpo          | yolo predict source=image.mpo     | Multi Picture Object            |
| .png          | yolo predict source=image.png     | Portable Network Graphics       |
| .tif          | yolo predict source=image.tif     | Tag Image File Format           |
| .tiff         | yolo predict source=image.tiff    | Tag Image File Format           |
| .webp         | yolo predict source=image.webp    | WebP                            |
| .pfm          | yolo predict source=image.pfm     | Portable FloatMap               |

* Use of TIFF files is common in satellite imagery such as Sentinel-2. One may prefer to preserve the TIFF format over convert it to PNG/JPG because TIFF allows the storage of georeferencing information. 
* I see that the `load_yolo_annotations` uses `cv2.imread` to read the image files. [OpenCV seems to support](https://docs.opencv.org/4.x/d4/da8/group__imgcodecs.html#gacbaa02cffc4ec2422dfa2e24412a99e2) many of the Ultralytics-supported formats.

https://github.com/roboflow/supervision/blob/1860fdb0a4e21edc5fa03d973e9f31c055bdcf4f/supervision/dataset/formats/yolo.py#L170

### Proposals
[...]

PR #1636, opened 2024-10-31
Allow "tiff" and more extensions in `DetectionDataset.from_yolo` function

patel-zeel:
# Description

Addresses #1554 as discussed with @LinasKo.

## Highlights
- [x] Allows any image extension in `DetectionDataset.from_yolo`.
Tested extensions (listed on [Ultralytics predict page](https://docs.ultralytics.com/modes/predict/#images)):
    - [x] bmp
    - [x] jpg
    - [x] mpo
    - [x] png
    - [x] tif
    - [x] webp
    - [ ] pfm - `Pillow` could not read `pfm` images, `cv2` is able to read them. I saw [here](https://github.com/ultralytics/ultralytics/blob/f4e7756bff5c1b0c246afb491f42e9f9de84dc84/ultralytics/models/yolo/classify/predict.py#L46) that even Ultralytics uses `Pillow` to read images. In that case, I am not sure if Ultralytics can "really" support `pfm` format, as their documentation suggests.
    - [x] dng
- [x] Image must be of "RGB" type, raise `ValueError` otherwise.

* `Pillow`

## Type of change

The changes are tested with [this colab notebook](https://colab.research.google.com/drive/17oe0lxQXHCABllLynG93cVMVpuwww2bb?usp=sharing).

### Test Case

100 images of each extension were generated along with their corresponding dummy labels and dummy `data.yml`. Test run checks if the function works for a particular extension and computes time taken to load the dataset.

### Time taken (in seconds) to run the test case
* Columns 0,1,2,3,4 are seed numbers. 
* `-1.11` value indicates that `supervision` doesn't support that extension yet.
#### Result for `supervision` version `0.24.0`, which uses `cv2` to check the shape of images.

|      |     0 |     1 |     2 |     3 |     4 |   Mean |   Std |
|:-----|------:|------:|------:|------:|------:|-------:|------:|
| bmp  | -1.11 | -1.11 | -1.11 | -1.11 | -1.11 |  -1.11 |  0.00    |
| jpg  |  0.98 |  1.16 |  0.85 |  0.76 |  0.77 |   0.90  |  0.15 |
| mpo  | -1.11 | -1.11 | -1.11 | -1.11 | -1.11 |  -1.11 |  0.00    |
| png  |  0.58 |  0.56 |  0.42 |  0.4  |  0.41 |   0.48 |  0.08 |
| tif  | -1.11 | -1.11 | -1.11 | -1.11 | -1.11 |  -1.11 |  0.00  |
| webp | -1.11 | -1.11 | -1.11 | -1.11 | -1.11 |  -1.11 |  0.00    |
| dng  | -1.11 | -1.11 | -1.11 | -1.11 | -1.11 |  -1.11 |  0.00   |

#### Result for this PR, which uses `Pillow` to check the shape and type (RGB or not) of images.

|      |    0 |    1 |    2 |    3 |    4 |   Mean |   Std |
|:-----|-----:|-----:|-----:|-----:|-----:|-------:|------:|
| bmp  | 0.04 | 0.02 | 0.02 | 0.02 | 0.03 |   0.03 |  0.01 |
| jpg  | 0.03 | 0.02 | 0.03 | 0.02 | 0.02 |   0.03 |  0.00    |
| mpo  | 0.04 | 0.04 | 0.06 | 0.04 | 0.04 |   0.04 |  0.01 |
| png  | 0.02 | 0.02 | 0.04 | 0.02 | 0.02 |   0.02 |  0.01 |
| tif  | 0.05 | 0.06 | 0.08 | 0.05 | 0.05 |   0.06 |  0.01 |
| webp | 0.10  | 0.11 | 0.11 | 0.09 | 0.11 |   0.10  |  0.00    |
| dng  | 0.06 | 0.06 | 0.05 | 0.06 | 0.06 |   0.06 |  0.00    |

## Docs

-   [ ] Docs updated? What were the changes: **Not changed, please suggest if we need changes in the docs**


LinasKo:
You never cease to surprise, @patel-zeel; such a thorough analysis! ⭐

Adding this to the back of my review backlog, but I already know it will be a delight.

patel-zeel:
Thank you for your kind words, @LinasKo. It's a pleasure to help improve a widely used library. Looking forward to your feedback!

patel-zeel:
@LinasKo a small correction. I recently found that Ultralytics uses [this function](https://github.com/ultralytics/ultralytics/blob/f932a611abf0ed0e359deecbb82fa8d361e23cfe/ultralytics/utils/patches.py#L15-L26) to read `pfm` and other image files during `yolo predict` call. It uses `cv2` and not `Pillow`. So, I was able to process `pfm` images with Ultralytics. However, the fact remains that Pillow does not support colored `pfm` images. [This PR](https://github.com/python-pillow/Pillow/pull/7696) added greyscale `pfm` support in Pillow and it points to [this issue](https://github.com/python-pillow/Pillow/issues/1888) for not supporting color images. [The issue](https://github.com/python-pillow/Pillow/issues/1888) is pinned on Pillow since May 2016 and thus, it seems the issue is not going to be resolved in near future. However, unless `pfm` format dominates some domains, we are not bottlenecked by it.

supervision/dataset/formats/yolo.py
```diff
@@ -167,10 +167,19 @@ def load_yolo_annotations(
...
             continue
 
-        image = cv2.imread(image_path)
+        # PIL is much faster than cv2 for checking image shape and mode: https://github.com/roboflow/supervision/issues/1554
+        image = Image.open(image_path)
         lines = read_txt_file(file_path=annotation_path, skip_empty=True)
-        h, w, _ = image.shape
+        w, h = image.size
         resolution_wh = (w, h)
+        if image.mode != "RGB":
```
SkalskiP, line 175: @patel-zeel looks like we can simplify the code here. There is no need for nested ifs.
```python
if image.mode not in ("RGB", "L"):
    raise ValueError(
        f"Images must be 'RGB' or 'grayscale', but {image_path} mode is '{image.mode}'."
    )
```

supervision/dataset/formats/yolo.py
```diff
@@ -167,10 +167,19 @@ def load_yolo_annotations(
...
+            if image.mode == "L":
+                image = image.convert("RGB")
```
SkalskiP, line 177: It seems to me that conversion to RGB is not necessary. The image is not used in the further part of the function.

supervision/dataset/formats/yolo.py
```diff
@@ -153,7 +153,7 @@ def load_yolo_annotations(
     image_paths = [
         str(path)
         for path in list_files_with_extensions(
-            directory=images_directory_path, extensions=["jpg", "jpeg", "png"]
+            directory=images_directory_path, extensions=["*"]
```
SkalskiP, line 156: There is a small but important side-effect of this change. If there are other files than images in the directory, we will also try to load them. For example, macOS puts a `.DS_Store` file in the directory. @patel-zeel I suggest putting here a list of image extensions that you have tested.

SkalskiP (changes requested):

SkalskiP:
@patel-zeel I apologize for making you wait so long for a review of this PR. I was on a long break at the turn of the year. I left some comments, please address them.

supervision/dataset/formats/yolo.py
patel-zeel: Right, that'd save us the time wasted in the conversion. I checked the other extreme as well. If we convert all images to RGB by default, that adds a bit of unnecessary overhead. So, this change will improve the speed further. Thank you for pointing it out, @SkalskiP.

supervision/dataset/formats/yolo.py
patel-zeel: I agree. On it!

patel-zeel:
No worries, @SkalskiP, I hope you had a good time during the break. Thank you for the review! I have addressed the changes.

SkalskiP:
@patel-zeel one more small comment and we should be good to go

supervision/dataset/formats/yolo.py
```diff
@@ -153,7 +153,18 @@ def load_yolo_annotations(
     image_paths = [
         str(path)
         for path in list_files_with_extensions(
-            directory=images_directory_path, extensions=["jpg", "jpeg", "png"]
+            directory=images_directory_path,
+            extensions=[
```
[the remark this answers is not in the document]
patel-zeel: I kept them in a single line @SkalskiP. Pre-commit (I think ruff) forced them on separate lines. Do we want to override that?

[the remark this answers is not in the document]
SkalskiP: eeeh that's what I thought so

SkalskiP (approved):

patel-zeel:
Thank you for the review and merge, @SkalskiP! 

I'd like to mention that this PR not only extends the support for more image extensions but also significantly speeds up the data loading with the `sv.DetectionDataset.from_yolo` function. The speed up is around 14x (or 1300%) as demonstrated in [this colab](https://colab.research.google.com/drive/1CiethT59rzFYZyXZI-nsHjZbjP0ZEI_q?usp=sharing).

SkalskiP:
Hi @patel-zeel 👋🏻 yes I know! in fact this is the main angle I plan to use when promoting this change.

PR #1637, opened 2024-10-31
feat(ByteTrack): Allow ByteTrack to track detection without class ids

Kadermiyanyedi:
ByteTrack was tested and confirmed to work without needing class_ids. After reviewing the ByteTrack documentation, also noted that class_ids were not essential, so I removed class_ids. The system functions correctly without them.

You can test the functionality with the provided [Colab Notebook](https://colab.research.google.com/drive/1SXZndmumMotpHobM3kTefuv2pkggyy_R?usp=sharing) which demonstrates ByteTrack handling cases with and without class_id.

#1582

### DIFF
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -1609,7 +1609,7 @@ class TraceAnnotator(BaseAnnotator):
         position: Position = Position.CENTER,
         trace_length: int = 30,
         thickness: int = 2,
-        color_lookup: ColorLookup = ColorLookup.INDEX,
+        color_lookup: ColorLookup = ColorLookup.CLASS,
     ):
         """
         Args:
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -1609,7 +1609,7 @@ class TraceAnnotator(BaseAnnotator):
         position: Position = Position.CENTER,
         trace_length: int = 30,
         thickness: int = 2,
-        color_lookup: ColorLookup = ColorLookup.CLASS,
+        color_lookup: ColorLookup = ColorLookup.INDEX,
     ):
         """
         Args:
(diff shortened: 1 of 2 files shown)


supervision/annotators/core.py
```diff
@@ -1609,7 +1609,7 @@ def __init__(
         position: Position = Position.CENTER,
         trace_length: int = 30,
         thickness: int = 2,
-        color_lookup: ColorLookup = ColorLookup.CLASS,
+        color_lookup: ColorLookup = ColorLookup.INDEX,
```
LinasKo, line 1612: Class IDs are not necessary, but when they are present (in most cases), it's worth showing the trails corresponding to class colors. Let's keep this as `.CLASS`.

LinasKo (changes requested):

supervision/annotators/core.py
Kadermiyanyedi: But we removed class ids argument, so now we don't have it. If you want to keep color lookup bytetrack should work like my previous PR

LinasKo: Not all situations that use a tracker will use `TraceAnnotator`. We enabled tracker for those people, but let's not break the defaults of `TraceAnnotator` - let's not break the defaults.

`resolve_color` will work as before as it takes entire detections, regardless if color was passed to ByteTrack. In 90% of the cases, detections will have class ID. In cases where they don't, `resolve_color_idx` will tell users why it's failing and how to fix it.

At least that's how I remember it

LinasKo:
I've tested the trace annotator with CLASS as the color selector, and it works as well as before.

Thank you for the contribution @Kadermiyanyedi! I'm merging this in!

LinasKo (approved):

PR #1656, opened 2024-11-06
feat(metrics): ✨ initial mean average recall metrics added

### LINKED ISSUE #1583: Metrics: Mean Average Recall (mAR)
# Metrics: Mean Average Recall (mAR)

> [!TIP]
> [Hacktoberfest](https://hacktoberfest.com/) is calling! Whether it's your first PR or your 50th, you’re helping shape the future of open source. Help us build the most reliable and user-friendly computer vision library out there! 🌱

We'd like to expand our suite of metrics with a new one - Mean Average Recall (mAR). This would involve creating it, its accompanying results class, and briefly testing it.

Note it is different from `mAP` in that it don't use the precision-recall curve, but recall-IoU. This affects `MeanAverageRecallResult`:

* The will not be `mAR@50`, `mAR@75`, etc - only the global main `mAR` value.
* Other frameworks threshold by max detections. I suggest we add this later if required.
* mAP currently implements a single averaging method, and F1 has 3 different ones (micro, macro, weighted). The choice between those is left to PR author - we can always add more later.
* The metric can report mAR values per-class. Do tell if that is difficult.
* I think we should have `1.0` as the default if no value are provided. `-1` is a consideration too, which should bring the same change to `mAP`.

Feel free to change the above if it feels better.

---

Helpful links:
* [Contribution guide](https://supervision.roboflow.com/develop/contributing/#how-to-contribute-changes)
* Metrics:
  * mAP metric: [docs](https://supervision.roboflow.com/develop/metrics/mean_average_precision/), [code](https://github.com/roboflow/supervision/blob/d6aa72c0f2b158b838145a81ed5995db6a1e9015/supervision/metrics/mean_average_precision.py#L25)
  * F1 Score: [docs](https://supervision.roboflow.com/develop/metrics/f1_score/), [code](https://github.com/roboflow/supervision/blob/d6aa72c0f2b158b838145a81ed5995db6a1e9015/supervision/metrics/f1_score.py#L25)
* [Supervision Cheatsheet](https://roboflow.github.io/cheatsheet-supervision/)
* [Colab Starter Template](https://colab.research.google.com/drive/1rin7WrS-UvVIe-_Gfxmu-yVslGphOq89#scrollTo=pjmCrNre2g58)
* [Prior metrics test Colab](https://colab.research.google.com/drive/1qSMDDpImc9arTgQv-qvxlTA87KRRegYN)

onuralpszr:
### Description

Initial mean average recall, metrics added. fixed #1583 

-   [X] This change requires a documentation update

## Docs

-   [X] Docs updated? What were the changes:

### DIFF
--- a/supervision/metrics/mean_average_recall.py
+++ b/supervision/metrics/mean_average_recall.py
@@ -59,7 +59,7 @@ class MeanAverageRecall(Metric):
         self,
         metric_target: MetricTarget = MetricTarget.BOXES,
         class_agnostic: bool = False,
-        max_detections: List[int] = [1, 10, 100],  # Add max_detections parameter
+        max_detections: List[int] = [1, 10, 100],
     ):
         """
         Initialize the Mean Average Recall metric.
@@ -131,7 +131,6 @@ class MeanAverageRecall(Metric):
         """
         result = self._compute(self._predictions_list, self._targets_list)
 
-        # Compute size-specific results
         for size_category, container in [
             (ObjectSizeCategory.SMALL, "small_objects"),
             (ObjectSizeCategory.MEDIUM, "medium_objects"),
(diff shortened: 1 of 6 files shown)


LinasKo:
Thanks for matching the structure of prior metrics, It'll be much easier to review.

supervision/metrics/mean_average_recall.py
```diff
@@ -0,0 +1,452 @@
...
+                    if not self._class_agnostic
+                    else slice(None)
+                )
+
+                if not any(target_mask):
+                    continue
+
+                class_iou = iou[target_mask]
+                if any(pred_mask):
+                    class_iou = class_iou[:, pred_mask]
```
LinasKo, line 207: General comment: I don't see the sorting of predictions by confidence anywhere. I believe we should evaluate the top-1, top-10, top-100 most confident predictions.

supervision/metrics/mean_average_recall.py
```diff
@@ -0,0 +1,452 @@
...
+            size_predictions = []
+            size_targets = []
+            for predictions, targets in zip(self._predictions_list, self._targets_list):
+                size_predictions.append(
+                    self._filter_detections_by_size(predictions, size_category)
+                )
+                size_targets.append(
+                    self._filter_detections_by_size(targets, size_category)
+                )
+            setattr(result, container, self._compute(size_predictions, size_targets))
```
LinasKo, line 148: Very nice 🙂
(no changes needed, feel free to resolve)

supervision/metrics/mean_average_recall.py
```diff
@@ -0,0 +1,452 @@
...
+        all_recalls = []
+        all_class_ids = []
+        iou_thresholds = np.linspace(0.5, 0.95, 10)
+
+        for predictions, targets in zip(predictions_list, targets_list):
+            if targets.is_empty():
+                continue
+
+            prediction_contents = self._detections_content(predictions)
+            target_contents = self._detections_content(targets)
```
LinasKo, line 175: Idea for later: in all metrics, we can move these 2 lines lower by a block (to  `if self._metric_target == MetricTarget.BOXES:`), and then extract into a function `_compute_iou`.

(No changes needed now.)

LinasKo (changes requested):
Son of a gun, you actually did it from first principles 😄

I made a handful of docs-related changes, removed a bit of dead code. A few things feel slightly off about the algo, and I left the comments - happy to have a call / discussion to figure out how it should work (I might definitely be wrong).

LinasKo:
I got an implementation out, and it's very similar to Torchmetrics, typically diverging by ~0.02 or so, up to 0.1.

In hindsight, my thinking about what max_detections is was incorrect yesterday. mAR @ 1 only considers 1 detection, but does NOT drop any targets. Therefore, the recall is almost always very small. `mAR@1` <= `mAR@10` <= `mAR@100`.

To get mAR of >0.5 is pretty tough - I bet values we've seen yesterday indicate that mAR@1 is too high.

LinasKo:
My filtering is implemented in `_compute_confusion_matrix`, check `max_detections` variable.

PR #1671, opened 2024-11-15
ci: 👷 update numpy dependency to a simplified version constraint

onuralpszr:
# Description

## Type of change

YOUR_ANSWER

## Docs


pyproject.toml
```diff
@@ -44,10 +44,10 @@ classifiers = [
 [tool.poetry.dependencies]
 python = "^3.8"
 numpy = [
-    { version = ">=1.21.2,<1.23.3", python = "<=3.10" },
-    { version = ">=1.23.3", python = ">3.10" },
+    { version = ">=1.21.2,>2.1.0", python = ">3.13" },
```
LinasKo, line 47: Should this be `>3.13`? Feels like a typo, and that it should've been <3.13

onuralpszr:
@LinasKo this change for helping inference to bump supervision version to latest and based on feedback I got @PawelPeczek-Roboflow asked me to relax numpy version for py38 and py39 and with this change it can install higher version than previously. 

Side note : I also remove pandas-stubs from metrics and based on tests all fine I am moving stubs under other typing package for mypy.

LinasKo:
1. If we're removing stubbs, we can remove the line entirely, rather than commenting it out.
2. What do you mean by a separate package for types? `supervision-stubs`, that we'd publish to PyPi?
3. To continue using pandas after calling `metric_result.as_pandas()`, you'd likely need to `import pandas as pd`. That line prompts installing stubbs. Would it not be more clear to the user if we installed it together with pandas? Or is that not a problem you see / not the ideal solution?

onuralpszr:
> 1. If we're removing stubbs, we can remove the line entirely, rather than commenting it out.
> 2. What do you mean by a separate package for types? `supervision-stubs`, that we'd publish to PyPi?
> 3. To continue using pandas after calling `metric_result.as_pandas()`, you'd likely need to `import pandas as pd`. That line prompts installing stubbs. Would it not be more clear to the user if we installed it together with pandas? Or is that not a problem you see / not the ideal solution?


1 - ) I can remove entirely for sure +1
2 - ) I mean we can move other typing mypy packages for "dev" we don't publish to pypi at all
3 - ) Pandas itself enough to have it. Mypy stub package not needed. (I tested metrics and didn't see any requirement or error at all for pandas-stubs, We only need "pandas". Stubs for "mypy" only)

LinasKo:
3. Okay. I've checked it, and the messages I see are mypy-specific. Sure - let's remove it for now. If we go the mypy route, we should offer something equivalent or offer stubs of our own, that also install pandas stubs.

I say this because as a user of mypy, having just installed `supervision[metrics]` to use `pandas`, seeing that I need to install more things is annoying.

image

pyproject.toml
```diff
@@ -44,10 +44,10 @@ classifiers = [
 [tool.poetry.dependencies]
 python = "^3.8"
 numpy = [
-    { version = ">=1.21.2,<1.23.3", python = "<=3.10" },
-    { version = ">=1.23.3", python = ">3.10" },
+    { version = ">=1.21.2,<=2.1.3", python = "<3.13" },
```
LinasKo, line 47: I suggest, if we don't impose an upper-bound for `>=3.13`, that we don't impose it for `<3.13`.

LinasKo (changes requested):

LinasKo (approved):

Issue #1681, opened 2024-11-24
Pyright reportPrivateImportUsage error on supervision.Detections


anriha:
### Search before asking

- [X] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar bug report.

### Bug

After supervision added `py.typed` support, Pyright is reporting a `reportPrivateImportUsage` error when trying to use `Detections` type from the package's root import, even though it's explicitly re-exported in the package's `__init__.py`.

### Environment

Supervision: 0.25.0
Python: 3.12.7
Pyright: 1.1.382
OS: NixOS

### Minimal Reproducible Example

```python
import supervision as sv
def process_detections(detections: sv.Detections) -> None:
    pass
```
    
# This triggers the error

```python
def process_detections(detections: sv.Detections) -> None:
    pass
```

### Additional

The error can be temporarily worked around by using direct import: `from supervision.detection.core import Detections`

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

onuralpszr:
Hello 👋 @anriha,

Could you share your pyright CLI command or configuration file? I was unable to reproduce the error you mentioned. Since pyright is used by Pylance in VS Code, I didn't see the error there. To double-check, I also installed pyright via npm and ran it on a file I created using the same code you provided. The result was:

> 0 errors, 0 warnings, 0 informations.

Additionally, I added a simple pyright configuration to a toml file for testing purposes, and I got the same results.

```toml
[tool.pyright]
include = ["supervision"]
exclude = ["**/node_modules",
    "**/__pycache__",
    "**/.mypy_cache",
]
defineConstant = { DEBUG = true }

reportMissingImports = "error"
reportMissingTypeStubs = false
reportPrivateImportUsage = true

```

Unless you have an alternative approach, I will proceed to close this issue. We are also planning to fully integrate mypy to ensure the project is compatible with static typing. I should mention that pyright and mypy sometimes differ in how they interpret typing rules. Once we finalize the mypy configuration, we might need to make adjustments or add specific settings for pyright or Pylance to ensure compatibility. However, our primary guideline will be the mypy configuration.


For information here is mypy vs pyright: https://github.com/microsoft/pyright/blob/main/docs/mypy-comparison.md

cc @LinasKo

anriha:
Thank you for looking into this. I'm still getting the error with just a basic `pyright file.py` command:
```python
# file.py
import supervision as sv

[...]

PR #1682, opened 2024-11-24
fix: 🐞 add missing exports to __all__ in supervision module

onuralpszr:
Fix: #1681


LinasKo:
Non-empty `__init__` files in repo:
```
./supervision/metrics/__init__.py
./supervision/__init__.py
./supervision/validators/__init__.py
./supervision/assets/__init__.py
```

1. I believe we should add this to `metrics/__init__.py` and `assets/__init__.py` as well.
2. Instead, do you think this would be a better approach?

```
import inspect

...

# Explicitly declares public variables; required for pyright. See #1681
__all__ = [
    name
    for name in globals()
    if not name.startswith("_") and not inspect.ismodule(globals()[name])
]
```

I teste the set of variables I get with this:
```
print(__all__)
print(__all__2)
print(set(__all__) == set(__all__2))
```

3. In case you dug this far: Shouldn't the definition of exported variables happen automatically for non-protected methods (without leading `_`)?

Thanks for triaging this issue at such a late hour!

onuralpszr:
1- For adding others I agreed

2- I like your for loop but the reason I prefer list is because "no import and logicless" sometimes using for loop + import more costly compare to simple list beside adding them not gonna be big of a deal anyway

LinasKo (approved):

PR #1691, opened 2024-11-26
Mask datatype specify in docs

Erol444:
Specified that mask is in bool datatype. Used int8 (gpt generated code) and was debugging this issue for 3 hours until I looked into source code of mask annotator.

When using int8 mask (instead of bool dtype), mask annotator will take crazy long to process (like 10sec on MacBook M2), and result won't be correct.

## Type of change

-   [x] Documentation update

Docs update

## Docs


LinasKo:
Hi @Erol444,

That's a good catch. Thank for sending a PR our way!

LinasKo (approved):

PR #1696, opened 2024-11-29
docs: 📝 extend supervision installation methods

onuralpszr:
# Description - Doc changes

This change introduce alternative installation methods of supervision for show audience other tooling too.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -33,8 +33,9 @@ We write your reusable computer vision tools. Whether you need to load your data
 You can install `supervision` in a
 [**Python>=3.8**](https://www.python.org/) environment.
 
-!!! example "pip install (recommended)"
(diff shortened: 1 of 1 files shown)


LinasKo:
Love it, this is great.

Do you think we should mention `uv add supervision` as well? My impression was that it's moving towards using `add` over `pip install`.

onuralpszr:
Let's keep both because they have different use case, "add" is for project like rye/poetry "uv pip" just directly installs no matter. We can mention difference between add or pip way If you like just simple one line 

```
         ```bash
        uv pip install supervision
        ```
        
        For uv projects

        ```bash
        uv add supervision
        ```
```

What do you think?

LinasKo (approved):

PR #1697, opened 2024-11-29
How To Benchmark Models

LinasKo:
# Description

Adding a How-To guide for model benchmarking, along with a [Colab version](https://colab.research.google.com/drive/1HoOY9pZoVwGiRMmLHtir0qT6Uj45w6Ps#scrollTo=eLnLRcRotv6D).

This is the most extensive guide to date, but I expect we'll shorten it somewhat & update it soon with new functionality (Annotator, class remapping function).

Temporarily relies on a specific branch in `inference`.

## Type of change

Ran with both Inference and Ultralytics, 

Temporarily relies on a specific branch in `inference`, checked out the looks with `mkdocs`.

## Docs

### DIFF
--- a/docs/how_to/benchmark_a_model.md
+++ b/docs/how_to/benchmark_a_model.md
@@ -3,9 +3,9 @@ comments: true
 status: new
 ---
 
-# Benchmark a Model
+![Corgi Example](https://media.roboflow.com/supervision/image-examples/how-to/benchmark-models/corgi-sorted-2.png)
 
-## Overview
+# Benchmark a Model
 
 Have you ever trained multiple detection models and wondered which one performs best on your specific use case? Or maybe you've downloaded a pre-trained model and want to verify its performance on your dataset? Model benchmarking is essential for making informed decisions about which model to deploy in production.
 
@@ -23,6 +23,8 @@ This guide will show an easy way to benchmark your results using `supervision`. 
 10. [Bonus: Model Leaderboard](#model-leaderboard)
 
 This guide will use an instance segmentation model, but it applies to object detection, instance segmentation, and oriented bounding box models (OBB) too.
+
+A condensed version of this guide is available as a [Colab Notebook](https://colab.research.google.com/drive/1HoOY9pZoVwGiRMmLHtir0qT6Uj45w6Ps?usp=sharing).
 
 ## Loading a Dataset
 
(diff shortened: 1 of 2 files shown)


SkalskiP:
Hi @LinasKo 👋🏻 

It's a bit late, as the PR has already been merged, but I have a few thoughts. Please make the changes before the release.

Good first iteration! I get the impression that you wanted to say too much, and as a result, the tutorial becomes chaotic. Keep it simple. Don't introduce too many side topics. Less is more. Simplify and organize.

1. We should treat object detection as the default example. Instance segmentation is probably an order of magnitude less common as a use case than object detection.

2. The part where you talk about which libraries we'll be using definitely shouldn't be in the "Loading a Dataset" section. In previous "How to" tutorials, we usually had a single sentence in the introduction where we mentioned which libraries we'd be using and linked to them.

Screenshot 2024-12-02 at 12 47 05

3. I think the tutorial should show how to benchmark a model on datasets in COCO, YOLO, and Pascal VOC formats. The fact that `supervision` is format-agnostic in this regard is a big asset, and we should highlight that.

This code snippet should have tabs for all three formats.

Screenshot 2024-12-02 at 13 23 21

Before running the model we should have a section showing how to iterate over dataset in all three formats.

Screenshot 2024-12-02 at 13 24 26

4. We don't need that extra table of contents. It's not consistent with the other "How to" guides that we have. It delays the user getting to the actual tutorial. Most importantly, the table of contents is already on the right side.

Screenshot 2024-12-02 at 12 27 13

5. Use concise action phrases as paragraph titles. For example, use "Load Model", "Run Model" instead of "Loading a model", "Running a model". This is consistent with the format we used in previous "How to" guides.

6. When using callouts like the ones below, place them within MKDocs Material Tip sections.

Screenshot 2024-12-02 at 13 04 07

7. I think just one tab dedicated to inference is enough. In this tutorial, you're loading a model trained on COCO, and let's stick to that. If you want to mention that inference can load any model from Roboflow Universe, do it using an MKDocs Material Tip below the code snippet.

Screenshot 2024-12-02 at 13 10 29

8. Paragraphs like this don't add much information. In this situation, I think it's better to omit the paragraph entirely.

Screenshot 2024-12-02 at 13 13 10

9. In our "How to" guides, we've been linking directly to specific classes/utils in the docs.

Screenshot 2024-12-02 at 13 20 12

To be consistent with what we have so far, I would write this as:

Screenshot 2024-12-02 at 13 20 36

"Use [`sv.BoxAnnotator`](https://supervision.roboflow.com/latest/detection/annotators/#supervision.annotators.core.BoxAnnotator) for object detection and [`sv.OrientedBoxAnnotator`](https://supervision.roboflow.com/latest/detection/annotators/#supervision.annotators.core.OrientedBoxAnnotator) for OBB."

10. Transformer is a gigantic library. We should strive to become the default benchmarking tool for detection and segmentation models there.

Screenshot 2024-12-02 at 13 25 57

PR #1588, opened 2024-10-10
Replaces github hosted images to roboflow hosting

joaomarcoscrs:
# Description

Changes some image links in the SAHI cookbook to use roboflow's image storage. It was breaking:

image
image

## Type of change

Please delete options that are not relevant.

-   [x] Bug fix (non-breaking change which fixes an issue)

## How has this change been tested, please provide a testcase or example of how you tested the change?

Go to the documentation page of the SAHI Cookbook and see that it gets no error rendering the images there.

## Any specific deployment considerations

For example, documentation changes, usability, usage/costs, secrets, etc.

## Docs

-   [x] Docs updated?

PR #645, opened 2023-12-05
Add a `sv.Detections.from_gcp_vision()` data loader

stellasphere:
# Description
Adds a `from_gcp_vision()` method to the Detections API in order to load results from Google Cloud Platform's Cloud Vision API into supervision.

Isolated and adapted the GCP Vision dataloader commit from #587. Removed the need for a classes dictionary argument and added docstrings.

## Type of change
-   [x] New feature (non-breaking change which adds functionality)

## How has this change been tested, please provide a testcase or example of how you tested the change?
Tested locally + [on this Colab](https://colab.research.google.com/drive/182ekhN5JbUDUOjzoxjrCey4s-vSGa81F?usp=sharing)

## Any specific deployment considerations
N/A

## Docs
-   [x] Docs updated? What were the changes:
Added docstrings for the new method

PR #1722, opened 2024-12-09
Improving github actions security

bigbitbus:
Adding some improvements to GH workflows.

1. Default permission restrictions
2. Action version pinning
3. Timeout limits
4. Python dependency version pinning in scripts
5. Adding basic CODEOWNERS
6. More eror handling
7. Concurrency controls (same action cannot be running in parallel)

# Description

We want to improve the GH workflows security.

## Type of change

Testing in progress...this comment will be updated

## Docs

### DIFF
--- a/.github/workflows/clear-cache.yml
+++ b/.github/workflows/clear-cache.yml
@@ -2,32 +2,41 @@
 name: Clear cache
 
 on:
-    schedule:
-      - cron: '0 0 1 * *'
-    workflow_dispatch:
+  schedule:
+    - cron: '0 0 1 * *'  # Run at midnight on the first day of every month
+  workflow_dispatch:
 
+# Restrict permissions by default
 permissions:
-  actions: write
+  actions: write  # Required for cache management
 
 jobs:
   clear-cache:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - name: Clear cache
-        uses: actions/github-script@v7
+        uses: actions/github-script@d7906e4ad0b1822421a7e6a35d5ca353c962f410 # v6.4.1
         with:
           script: |
-            console.log("About to clear")
+            console.log("Starting cache cleanup...")
             const caches = await github.rest.actions.getActionsCacheList({
               owner: context.repo.owner,
               repo: context.repo.repo,
             })
+            
+            let deletedCount = 0
             for (const cache of caches.data.actions_caches) {
-              console.log(cache)
-              github.rest.actions.deleteActionsCacheById({
-                owner: context.repo.owner,
-                repo: context.repo.repo,
-                cache_id: cache.id,
-              })
+              console.log(`Deleting cache: ${cache.key} (${cache.size_in_bytes} bytes)`)
+              try {
+                await github.rest.actions.deleteActionsCacheById({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  cache_id: cache.id,
+                })
+                deletedCount++
+              } catch (error) {
+                console.error(`Failed to delete cache ${cache.key}: ${error.message}`)
+              }
             }
-            console.log("Clear completed")
+            console.log(`Cache cleanup completed. Deleted ${deletedCount} caches.`)
--- a/.github/workflows/notebook-bot.yml
+++ b/.github/workflows/notebook-bot.yml
@@ -4,24 +4,28 @@ on:
   pull_request_target:
     types: [opened, reopened]
 
-permissions:
-  contents: read
+# Restrict permissions by default
+permissions: {}
 
 jobs:
   comment-welcome:
     permissions:
       contents: read
-      pull-requests: write
-
+      pull-requests: write  # Required for commenting on PRs
+    
     runs-on: ubuntu-latest
+    timeout-minutes: 10
+    
     steps:
     - name: Fetch pull request branch
-      uses: actions/checkout@v4
+      uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0
       with:
         repository: ${{ github.event.pull_request.head.repo.full_name }}
         ref: ${{ github.event.pull_request.head.sha }}
+        
     - name: Fetch base develop branch
       run: git fetch -u "$GITHUB_SERVER_URL/$GITHUB_REPOSITORY" develop:develop
+      
     - name: Create message
       env:
         HEAD_REPOSITORY: ${{ github.event.pull_request.head.repo.full_name }}
@@ -49,16 +53,20 @@ jobs:
           msg+="If commits are added to the pull request, synchronize your local branch: <code>git pull origin $HEAD_REF</code>\n"
         fi
         echo "MESSAGE=$msg" >> $GITHUB_ENV
-    - name: Post comment
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        ISSUE_URL: ${{ github.event.pull_request.issue_url }}
-      run: |
-        # Env var defined in previous step. Escape string for JSON.
-        body="$(echo -n -e $MESSAGE | python3 -c 'import json,sys; print(json.dumps(sys.stdin.read()))')"
-        # Add comment to pull request.
-        curl -X POST \
-          -H "Accept: application/vnd.github.v3+json" \
-          -H "Authorization: token $GITHUB_TOKEN" \
-          "${ISSUE_URL}/comments" \
-          --data "{\"body\": $body}"
+
+    - name: Find Comment
+      uses: peter-evans/find-comment@a54c31d7fa095754bfef525c0c8e5e5674c4b4b1 # v2.4.0
+      id: find-comment
+      with:
+        issue-number: ${{ github.event.pull_request.number }}
+        body-includes: Preview and run these notebook edits
+        comment-author: github-actions[bot]
+
+    - name: Add or update comment
+      uses: peter-evans/create-or-update-comment@23ff15729ef2fc348714a3bb66d2f655ca9066f2 # v3.1.0
+      if: env.MESSAGE != ''
+      with:
+        comment-id: ${{ steps.find-comment.outputs.comment-id }}
+        issue-number: ${{ github.event.pull_request.number }}
+        body: ${{ env.MESSAGE }}
+        edit-mode: replace
--- a/.github/workflows/publish-dev-docs.yml
+++ b/.github/workflows/publish-dev-docs.yml
@@ -6,48 +6,48 @@ on:
       - develop
   workflow_dispatch:
 
+# Ensure only one concurrent deployment
 concurrency:
   group: ${{ github.workflow }}-${{ github.event_name == 'push' && github.ref}}
   cancel-in-progress: true
 
+# Restrict permissions by default
 permissions:
-  contents: write
-  pages: write
-  pull-requests: write
-
+  contents: write  # Required for committing to gh-pages
+  pages: write    # Required for deploying to Pages
+  pull-requests: write  # Required for PR comments
 
 jobs:
   deploy:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     steps:
       - name: 🔄 Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0
         with:
           fetch-depth: 0
+          
       - name: 🐍 Set up Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@b64ffcaf5b410884ad320a9cfac8866006a109aa # v5.0.0
         with:
           python-version: '3.10'
-      - name: 📦 Install mkdocs-material
-        run: pip install "mkdocs-material"
-      - name: 📦 Install mkdocstrings[python]
-        run: pip install "mkdocstrings[python]"
-      - name: 📦 Install mkdocs-material[imaging]
-        run: pip install "mkdocs-material[imaging]"
-      - name: 📦 Install mike
-        run: pip install "mike"
-      - name: 📦 Install mkdocs-git-revision-date-localized-plugin
-        run: pip install "mkdocs-git-revision-date-localized-plugin"
-      - name: 📦 Install JupyterLab
-        run: pip install jupyterlab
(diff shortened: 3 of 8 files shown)


.github/workflows/notebook-bot.yml
```diff
@@ -4,24 +4,28 @@ on:
   pull_request_target:
```
LinasKo, line 4: Alex mentioned we should replace this with `pull_request` whenever possible. Got any insight on that perchance?

.github/workflows/publish-test.yml
```diff
@@ -1,30 +1,36 @@
...
 jobs:
   build-and-publish-pre-release-pypi:
     name: Build and publish to PyPI
     runs-on: ubuntu-latest
     environment: test
+    timeout-minutes: 15
     permissions:
-      id-token: write
+      id-token: write  # Required for PyPI publishing
+      contents: read   # Required for checkout
```
LinasKo, line 22: Out of curiousity, is this explicitly clarifying the permissions level we already had? Or were we granting `write` accidentally?

.github/workflows/publish-dev-docs.yml
```diff
@@ -6,48 +6,48 @@ on:
...
-      - name: 📦 Install mike
-        run: pip install "mike"
-      - name: 📦 Install mkdocs-git-revision-date-localized-plugin
-        run: pip install "mkdocs-git-revision-date-localized-plugin"
-      - name: 📦 Install JupyterLab
-        run: pip install jupyterlab
-      - name: 📦 Install mkdocs-jupyter
-        run: pip install mkdocs-jupyter
-      - name: 📦 Install mkdocs-git-committers-plugin-2
-        run: pip install mkdocs-git-committers-plugin-2
```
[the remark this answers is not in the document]
onuralpszr: Just readability but we can combine as well

LinasKo (changes requested):
Note: The comments mention CODEOWNERS, but I can't see it among the changes.

.github/workflows/publish-dev-docs.yml
```diff
@@ -6,48 +6,48 @@ on:
...
+      - name: 📦 Install dependencies
+        run: |
+          pip install "mkdocs-material==9.5.2" \
+                    "mkdocstrings[python]==0.24.0" \
+                    "mkdocs-material[imaging]==9.5.2" \
+                    "mike==2.0.0" \
+                    "mkdocs-git-revision-date-localized-plugin==1.2.1" \
+                    "jupyterlab==4.0.9" \
+                    "mkdocs-jupyter==0.24.6" \
+                    "mkdocs-git-committers-plugin-2==2.2.2"
```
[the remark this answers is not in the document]
onuralpszr: This version lock means, our local always most update and this ci change stay old after certain point If I don't update ourselves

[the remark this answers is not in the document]
onuralpszr: For same version we can use same old always latest (which is %99 percent always latest and same as local)

.github/workflows/publish-test.yml
bigbitbus: Its better to explicitly clarify permission levels; the default permission levels depend on the GITHUB_TOKEN being used (need to check that in the repo). More info here: https://graphite.dev/guides/github-actions-permissions

.github/workflows/notebook-bot.yml
bigbitbus: Will check with @alexnorell

.github/workflows/publish-dev-docs.yml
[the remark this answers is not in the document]
alexnorell: This needs to be a requirements.txt file so that Dependabot can automatically suggest new versions to update to.

.github/workflows/notebook-bot.yml
alexnorell: Yeah, this needs to be refactored to `pull_request`

.github/workflows/publish-dev-docs.yml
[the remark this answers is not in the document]
onuralpszr: We can poetry install directly instead of?doubling file and bot pr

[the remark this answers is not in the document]
alexnorell: Yes. A poetry group makes even more sense. DIdn't remember which package definition this project used.

.github/workflows/clear-cache.yml
```diff
@@ -2,32 +2,41 @@
...
+  actions: write  # Required for cache management
 
 jobs:
   clear-cache:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - name: Clear cache
-        uses: actions/github-script@v7
+        uses: actions/github-script@d7906e4ad0b1822421a7e6a35d5ca353c962f410 # v6.4.1
```
bigbitbus, line 19: Fix the shas to the actual version, not the one generated by AI

.github/workflows/publish-dev-docs.yml
```diff
@@ -6,48 +6,48 @@ on:
...
+
       - name: ⚙️ Configure git for github-actions
         run: |
           git config --global user.name "github-actions[bot]"
           git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
       - name: 🚀 Deploy MkDoc-Material with mike
         run: |
-          MKDOCS_GIT_COMMITTERS_APIKEY=${{ secrets.GITHUB_TOKEN }} mike deploy --push develop
+          MKDOCS_GIT_COMMITTERS_APIKEY=${{ secrets.GITHUB_TOKEN }} mike deploy --push --update-aliases develop latest
```
bigbitbus, line 53: As discussed revert this line.

LinasKo:
timeouts are all different, arbitrarily

.github/workflows/notebook-bot.yml
```diff
@@ -4,24 +4,28 @@ on:
...
-      uses: actions/checkout@v4
+      uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0
       with:
         repository: ${{ github.event.pull_request.head.repo.full_name }}
         ref: ${{ github.event.pull_request.head.sha }}
+
     - name: Fetch base develop branch
       run: git fetch -u "$GITHUB_SERVER_URL/$GITHUB_REPOSITORY" develop:develop
+
     - name: Create message
```
LinasKo, line 29: ❗ @bigbitbus, @alexnorell, Let's take a different approach and remove this action entirely (none of the steps are required).

* The comment informing about ReviewNB is written by reviewnb-bot ([example](https://github.com/roboflow/supervision/pull/645#issuecomment-2509554234))
* We were using this action temporarily, before we got the ReviewNB permissions approved. https://github.com/roboflow/supervision/pull/1588#issuecomment-2405844414

LinasKo (changes requested):

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: 📦 Install dependencies
+        run: |
+          poetry install --with=docs
+
+      - name: ⚙️ Configure git for github-actions
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+      - name: 🚀 Deploy Development Docs
```
alexnorell, line 50: @LinasKo @onuralpszr I moved this to a single workflow to make it easier to maintain in the future.

Something that I just noticed. workflow dispatch won't work right now. Is this still something you want?

.github/workflows/poetry-test.yml
```diff
@@ -38,6 +38,3 @@ jobs:
 
       - name: 🧪 Run the Import test
         run: poetry run python -c "import supervision; from supervision import assets; from supervision import metrics; print(supervision.__version__)"
-
```
alexnorell, line 41: Removed this since it is a duplicate of the pytest run.

.github/workflows/publish-test.yml
```diff
@@ -1,43 +1,51 @@
...
 
+      - name: 📜 Setup Poetry
+        uses: abatilo/actions-poetry@e78f54a89cb052fff327414dd9ff010b5d2b4dbd # v3.0.1
+
       - name: 🏗️ Build source and wheel distributions
         run: |
-          python -m pip install --upgrade build twine
-          python -m build
+          poetry install --with=build
+          poetry build
```
alexnorell, line 39: Does the build with poetry and checks the build outputs using twine.

.github/workflows/publish.yml
```diff
@@ -1,40 +1,49 @@
...
           python-version: ${{ matrix.python-version }}
 
+      - name: 📜 Setup Poetry
+        uses: abatilo/actions-poetry@e78f54a89cb052fff327414dd9ff010b5d2b4dbd # v3.0.1
+
       - name: 🏗️ Build source and wheel distributions
         run: |
-          python -m pip install --upgrade build twine
-          python -m build
+          poetry install --with=build
```
alexnorell, line 36: Same as above. Build the package with poetry and test with twine

.github/workflows/test.yml
```diff
@@ -4,26 +4,41 @@ on:
...
+        uses: abatilo/actions-poetry@e78f54a89cb052fff327414dd9ff010b5d2b4dbd # v3.0.1
 
       - name: 📦 Install dependencies
         run: |
-          python -m pip install --upgrade pip
-          pip install .
-          pip install pytest
+          poetry install --with=test
 
-      - name: 🧪 PyTest
```
alexnorell, line 28: This now runs pytest from an action and posts the results as a job summary.

pyproject.toml
```diff
@@ -133,6 +136,8 @@ types-requests = "^2.32.0.20240712"
 types-tqdm = "^4.66.0.20240417"
 pandas-stubs = ">=2.0.0.230412"
 
+[tool.poetry.group.build.dependencies]
```
alexnorell, line 139: Created new group for build dependencies

pyproject.toml
```diff
@@ -85,15 +85,18 @@ tqdm = ">=4.62.3"
...
+    { version = ">=2.0.0", python = "<3.13", optional = true },
+    { version = ">=2.2.3", python = ">=3.13", optional = true },
 ]
 
 opencv-python = ">=4.5.5.64"
 
 [tool.poetry.extras]
 metrics = ["pandas"]
 
+[tool.poetry.group.test.dependencies]
```
alexnorell, line 97: Created new group for test dependencies

.github/workflows/publish.yml
```diff
@@ -1,40 +1,49 @@
...
+          poetry build
           twine check --strict dist/*
 
       - name: 🚀 Publish to PyPi
-        uses: pypa/gh-action-pypi-publish@release/v1.10
+        uses: pypa/gh-action-pypi-publish@67339c736fd9354cd4f8cb0b744f2b82a74b5c70 # v1.12.3
+        with:
+          attestations: true
```
alexnorell, line 43: @LinasKo enabled attestations

.github/workflows/publish-docs.yml
LinasKo: My assumption is - we do want to enable manual dispatch of actions whenever possible.
(Also, it's nice to finally find how that's implemented)

.github/workflows/poetry-test.yml
LinasKo: @onuralpszr, was this the test that revealed the numpy import error a few weeks back? If so, did pytest not discover it?

onuralpszr: Let's remove old test and use poetry and your question answer is no it can't because i used pip to install

onuralpszr: What I am trying to say 2 things, we have new poetry/test actions which also check our "dependencies" and do "pytest" in case of "toml" file changes, we also have just a pytest by not using poetry at all. I said we can remove "old"  pytest/pip only one and use poetry + pytest (new action I recently I added). Second thing I notice in we are now using "poetry action" usually we do install via just pip but now we have full action, I was curious that why do we have that too. Because for me it is more complex and more dep

LinasKo: In your second point, you're referring to this change, I believe.

```diff
- python -m pip install --upgrade pip
- pip install .
- pip install pytest
+ poetry install --with=test
```

If so, I am in favor of the new approach. When developing locally, we usually install with `poetry install --with dev` - that's what our docs advise. Even if using `poetry` introduces an extra layer, it's more consistent with our guidelines.

.github/workflows/clear-cache.yml
```diff
@@ -1,33 +1,42 @@
...
 
 jobs:
   clear-cache:
+    name: Clear cache
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - name: Clear cache
-        uses: actions/github-script@v7
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
```
LinasKo, line 19: @onuralpszr, we had a word, and you expressed concern this approach would be less maintainable.

My question is: what steps would we need to go through if a dependabot highlights an issue? Does it really involve some complication in figuring out a working version, if we meticulously leave comments about the version previously used?

My understanding is that pinning to a hash guarantees that package was not modified.

I believe the threat model we're considering is a package version being removed from a cache and reuploaded with modified code. (correct me if I'm wrong, @alexnorell, @bigbitbus)

alexnorell: Actions managed by git tags. Git tags are not immutable. You can delete a git tag, and republish the tag to point to a new commit.

To solve for this limitation, github actions are also able to reference specific commit hashes for an action. This approach is specifically called out in by GitHub in their guide on how to use actions securely: https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

alexnorell: For the update overhead, dependabot will automatically suggest changes and bump both this hash and the comment that reflects what it is pointing to.

LinasKo: Excellent answer, thank you!

LinasKo (approved):
This covers all the points we discussed.

Thank you for the good work, @alexnorell, @bigbitbus!

Issue #1707, opened 2024-12-03
Implement metrics comparison table & plotting


LinasKo:
The metrics system allows the users to compute a metrics result - a class with values for a specific metrics run.

When comparing multiple models, a natural next step is to aggregate the results into a single table and/or plot them on a single chart.

Let's make this step easy!

I propose two new functions:
```python
def aggregate_metric_results(metrics_results: List[MetricResult], *,  include_object_sizes=False) -> pd.DataFrame:
   """
   Raises when different types of metrics results are passed in
   """

def plot_aggregate_metric_results(metrics_results: List[MetricResult], *, include_object_sizes=False) -> None:
    ...


class MetricResult(ABC):
   @abstractmethod
   def to_pandas():
      raise NotImplementedError()

   def plot():
      raise NotImplementedError()
```

---

Several questions we need to address:
1. Would it be better for `plot_aggregate_metric_results` to take in a `pd.DataFrame`? This way, the user can apply their own sorting or preprocessing, but we're not guaranteed to have the fields we need.
2. Could the naming be improved?

---

Suggested by @David-rn, [discussion](https://github.com/roboflow/supervision/discussions/1706#discussioncomment-11445217)

LinasKo:
Note: Please share a Google Colab with minimal code to test the new feature. We know it's additional work, but it will speed up the review process. You may use the [Starter Template](https://colab.research.google.com/drive/1rin7WrS-UvVIe-_Gfxmu-yVslGphOq89). The reviewer must test each change. Setting up a local environment to do this is time-consuming. Please ensure that Google Colab can be accessed without any issues (make it public). Thank you! :pray::skin-tone-2:

David-rn:
Regarding the first question and considering [the guide](https://supervision.roboflow.com/develop/how_to/benchmark_a_model/#f1-score) that led to this discussion, I think that maybe it is a better option using a `List[MetricResult]` as input as first suggested, since the user would already have used supervision classes (`Detections`, `MetricResult`). 

If you think it's a good idea, I can give it a try these days and share a Colab to evaluate whether it is feasible or there are more things to consider.

LinasKo:
That'd be most helpful @David-rn ! I'm assigning it to you.

David-rn:
I have written both functions and I came across some things: 

[...]

PR #1731, opened 2024-12-11
Aggregate Metric Results

David-rn:
# Description

This PR implements issue #1707. Currently, the metric system allows you to compute a metrics result. However, it is not possible to easily compare the results from several models in the same table or plot. In this PR, it is added two new functions `aggregate_metric_results` and `plot_aggregate_metric_results` that solves the comparison problem. 

Besides, it has been added a base class `MetricResult` from which the different results are inherited. 

## Type of change

 New feature (non-breaking change which adds functionality)

It has been tested using Google Colab and it can be found [here](https://colab.research.google.com/drive/1j5QoVmFB3WXqfOZJ8Co8r22ksl76CHzs?usp=sharing). Inside the Colab it is possible to find the different use cases considered. 

N/A

### DIFF
--- a/supervision/metrics/aggregate_metric_results.py
+++ b/supervision/metrics/aggregate_metric_results.py
@@ -13,7 +13,22 @@ def aggregate_metric_results(
     include_object_sizes=False,
 ) -> "pd.DataFrame":
     """
-    Raises when different types of metrics results are passed in
+    Convert a list of results to a pandas DataFrame.
+
+    Args:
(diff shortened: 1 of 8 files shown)


Borda:
@David-rn, I am sorry that this PR has gone stale, mainly because we and atm #2498 were aiming at the same feature and went much further, so we will proceed with that one. However, I really appreciate your efforts and will try to give credit to both of you 🦝

PR #1736, opened 2024-12-13
Fix release docs workflow file

alexnorell:
Fix release docs workflow. Copy and paste error added in an extra `)` that caused the workflow to not run.


onuralpszr (approved):
Small typo I see approved

PR #1737, opened 2024-12-13
Update docs checkout to use fetch-depth: 0

alexnorell:
Switch the checkout back to using fetch-depth 0 as suggested by the docs build process.


onuralpszr (approved):
Lgtm

PR #1739, opened 2024-12-14
fix: allow negative offsets in move_masks function (#1715)

### LINKED ISSUE #1715: `move_masks` only supports movement in positive direction
If you compare the code of `move_masks`, `move_detections` and `move_oriented_boxes`, you'll find that only mask code restricts offset direction:

```python
    if offset[0] < 0 or offset[1] < 0:
        raise ValueError(f"Offset values must be non-negative integers. Got: {offset}")
```

It should be possible to move masks in either direction, even if it results in cropping.

To complete this:
- [ ] Change the code so masks can be moved with negative offset
- [ ] Create a unit test suite for `move_masks`

---

It would help us immensely and speed up the review process if you could create a [Colab](https://colab.research.google.com/) showcasing the changes, but for this task it is optional. You may use the [Starter Template](https://colab.research.google.com/drive/1rin7WrS-UvVIe-_Gfxmu-yVslGphOq89?usp=sharing).

Anirudh2112:
# Description
This PR fixes the restriction on negative offsets in the `move_masks` function. Previously, the function would raise a ValueError for negative offset values, but similar functions like `move_detections` and `move_oriented_boxes` support negative offsets. This change makes the behavior consistent across all movement functions.

Fixes #1715

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

Added comprehensive unit tests covering:
- Positive offset (existing behavior)
- Negative offset (new behavior)
- Zero offset
- Partial out of bounds (positive)
- Partial out of bounds (negative)
- Multiple masks with different patterns

Example test case:
```python
masks = np.array([[[True, True], [True, True]]])
offset = np.array([-1, -1])
resolution_wh = (3, 3)
result = move_masks(masks=masks, offset=offset, resolution_wh=resolution_wh)


SkalskiP (approved):

SkalskiP:
Hi @Anirudh2112; thank you so much for your interest in Supervision and this specific task. I took the liberty of making a few changes to the solution you designed because I noticed it wasn’t working correctly with mixed offsets, for example, `[-2, 2]`.

Anirudh2112:
@SkalskiP, thank you for reviewing and improving the solution! This was a great learning experience for me in terms of handling edge cases and comprehensive testing. I'll be sure to consider such mixed cases in future contributions.

SkalskiP:
@Anirudh2112 great to hear that; I can't wait to see your future contributions!

PR #1748, opened 2024-12-17
Fix polygon trigger false positives when bounding boxes are clipped

YoungjaeDev:
# Description

**Fixed an issue where objects partially overlapping with polygons were incorrectly detected. Previously, when a polygon was positioned on the left side of the screen and a person on the right, with the person's bounding box slightly overlapping the polygon, the clip_boxes() function would forcefully clip the boxes to fit within the frame. This caused objects outside the frame to be incorrectly identified as being inside the polygon.**

### Changes

- Modified the trigger() function to use original bounding box coordinates (pre-clipping) for polygon intersection checks
- This reduces false positives and ensures more accurate counting while maintaining logical consistency

image

The original bounding box is large, and its bottom center point does not overlap with the polygon. However, during the clipping process, since the bottom-left corner intersects with the polygon, the box is forcefully shrunk. This causes the center point to shift inside the polygon, resulting in false detections.

Thank you

## Type of change

YOUR_ANSWER

## Docs

### DIFF
--- a/supervision/detection/tools/polygon_zone.py
+++ b/supervision/detection/tools/polygon_zone.py
@@ -1,4 +1,3 @@
-from dataclasses import replace
 from typing import Iterable, Optional
 
 import cv2
@@ -6,7 +5,7 @@ import numpy as np
 import numpy.typing as npt
 
 from supervision import Detections
-from supervision.detection.utils import clip_boxes, polygon_to_mask
+from supervision.detection.utils import polygon_to_mask
 from supervision.draw.color import Color
 from supervision.draw.utils import draw_filled_polygon, draw_polygon, draw_text
 from supervision.geometry.core import Position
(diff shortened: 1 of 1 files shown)


SkalskiP:
Hi @YoungjaeDev 👋🏻 I'm not sure I understand correctly. Did your bounding boxes have coordinates that extended beyond the video frame?

YoungjaeDev:
@SkalskiP 

Hi,

In the video, the `trigger_anchor` settings are `BOTTOM_LEFT` and `BOTTOM_CENTER`.

However, as shown in the image, the bounding box of the person had only the left corner inside the polygon, yet the count still increased.

Upon inspecting the `zone_count` code, it seems that clipping the detections to the polygon’s `x-max` and `y-max` causes the bounding box width to narrow. Consequently, the `BOTTOM_CENTER` anchor ends up inside the polygon, triggering a count.

In the PR I submitted, the code is as follows:

```python
original_anchors = np.array([
    np.ceil(detections.get_anchors_coordinates(anchor)).astype(int)
    for anchor in self.triggering_anchors
])

original_anchors_clamped = np.clip(
    original_anchors,
    a_min=[0, 0],
    a_max=[self.mask.shape[1] - 1, self.mask.shape[0] - 1]
)

is_in_zone_original = self.mask[
    original_anchors_clamped[:, :, 1],
    original_anchors_clamped[:, :, 0]
].transpose().astype(bool)
```

This code calculates the anchors based on the original detection coordinates and clamps them to the polygon mask’s dimensions. This approach avoids distorting the bounding box and accurately checks whether the anchor is within the polygon.

YoungjaeDev:
Please understand that there are many masks on the image to protect privacy.

SkalskiP:
Hi @YoungjaeDev 👋🏻 sorry for the late reply. I had some time off at the turn of the year and just got back. Do I understand correctly that the coordinates of your bounding boxes exceed the dimensions of the image? Otherwise, clipping would not have a negative effect for you.

soumik12345:
Hi @YoungjaeDev, we're closing this PR since there has been no recent activity. Feel free to reopen the PR if there's any new update.

PR #1770, opened 2025-01-08
docs/insiders

alexnorell:


onuralpszr:
@alexnorell  could you please hold this PR and also please do not add as direct doc dep to as insiders (If you do this no-one can make release and secondly, I am going to change poetry as well so please hold.

PR #1774, opened 2025-01-09
Support for IOS Matching Metric. Introduced the `mask_non_max_merge` function for handling non-maximum merging of masks

SunHao-AI:
# Description
When I used the yolo11-seg model for slice inference, I found that the merging did not work well. After analysis, it is found that for slender objects, its mask area only accounts for a small part, and the box area may be very large, so it is necessary to use mask for iou calculation. aaa
When I implemented the mask_non_max_merge function, I found that the merge effect was still not satisfactory, as shown in the figure：
NMM-IOU10
I found that the sahi library implemented the non-maximum merge algorithm of the IOS version, so I made changes to the code.
NMM-IOS10

## Type of change

```python
import cv2
import numpy as np
from PIL import Image
from ultralytics import YOLO

import supervision as sv

model = YOLO("yolon11-seg.pt")
image = cv2.imread(<your_image_path>)


def callback(image_slice: np.ndarray) -> sv.Detections:
    results = model(image_slice)[0]
    return sv.Detections.from_ultralytics(results)

# slicer = sv.InferenceSlicer(callback=callback, slice_wh=(960, 960), overlap_ratio_wh=None, overlap_wh=(50, 50), overlap_filter="non_max_merge", iou_threshold=0.1, match_metric="IOU")
slicer = sv.InferenceSlicer(callback=callback, slice_wh=(960, 960), overlap_ratio_wh=None, overlap_wh=(50, 50), overlap_filter="non_max_merge", iou_threshold=0.1, match_metric="IOS")
# slicer = sv.InferenceSlicer(callback=callback, slice_wh=(960, 960), overlap_ratio_wh=None, overlap_wh=(50, 50), overlap_filter="non_max_suppression", iou_threshold=0.1, match_metric="IOU")
# slicer = sv.InferenceSlicer(callback=callback, slice_wh=(960, 960), overlap_ratio_wh=None, overlap_wh=(50, 50), overlap_filter="non_max_suppression", iou_threshold=0.1, match_metric="IOS")
detections = slicer(image)

box_annotator = sv.BoxAnnotator()
labels = [f"{class_name} {confidence:.2f}" for class_name, confidence in zip(detections["class_name"], detections.confidence)]

annotated_image = box_annotator.annotate(scene=image, detections=detections)
annotated_image = mask_annotator.annotate(scene=annotated_image, detections=detections)
annotated_image = label_annotator.annotate(scene=annotated_image, detections=detections, labels=labels)
Image.fromarray(annotated_image[..., ::-1]).save("test.jpg")
```

## Docs

### DIFF
--- a/supervision/detection/core.py
+++ b/supervision/detection/core.py
@@ -1586,8 +1586,8 @@ class Detections:
             class_agnostic (bool): Whether to perform class-agnostic
                 non-maximum suppression. If True, the class_id of each detection
                 will be ignored. Defaults to False.
-            overlap_metric (OverlapMetric): Metric used for matching detections
-                in slices.
+            overlap_metric (OverlapMetric): Metric used for measuring overlap between
+                detections in slices.
 
         Returns:
             Detections: A new Detections object containing the subset of detections
@@ -1624,13 +1624,13 @@ class Detections:
                 predictions=predictions,
                 masks=self.mask,
                 iou_threshold=threshold,
-                match_metric=overlap_metric,
+                overlap_metric=overlap_metric,
             )
         else:
             indices = box_non_max_suppression(
                 predictions=predictions,
                 iou_threshold=threshold,
-                match_metric=overlap_metric,
+                overlap_metric=overlap_metric,
             )
 
         return self[indices]
@@ -1650,8 +1650,8 @@ class Detections:
             class_agnostic (bool): Whether to perform class-agnostic
                 non-maximum merging. If True, the class_id of each detection
                 will be ignored. Defaults to False.
-            overlap_metric (OverlapMetric): Metric used for matching detections in slices.
-                "IOU" or "IOS". Defaults "IOU".
+            overlap_metric (OverlapMetric): Metric used for measuring overlap between
+                detections in slices.
 
         Returns:
             Detections: A new Detections object containing the subset of detections
@@ -1690,13 +1690,13 @@ class Detections:
                 predictions=predictions,
                 masks=self.mask,
                 iou_threshold=threshold,
-                match_metric=overlap_metric,
+                overlap_metric=overlap_metric,
             )
         else:
             merge_groups = box_non_max_merge(
                 predictions=predictions,
                 iou_threshold=threshold,
-                match_metric=overlap_metric,
+                overlap_metric=overlap_metric,
             )
 
         result = []
@@ -1803,7 +1803,9 @@ def merge_inner_detection_object_pair(
 
 
 def merge_inner_detections_objects(
-    detections: List[Detections], threshold=0.5, match_metric: str = "IOU"
+    detections: List[Detections],
+    threshold=0.5,
+    overlap_metric: OverlapMetric = OverlapMetric.IOU,
 ) -> Detections:
     """
     Given N detections each of length 1 (exactly one object inside), combine them into a
@@ -1816,9 +1818,11 @@ def merge_inner_detections_objects(
     detections_1 = detections[0]
     for detections_2 in detections[1:]:
         if detections_1.mask is not None and detections_2.mask is not None:
-            iou = mask_iou_batch(detections_1.mask, detections_2.mask, match_metric)[0]
+            iou = mask_iou_batch(detections_1.mask, detections_2.mask, overlap_metric)[
+                0
+            ]
         else:
-            iou = box_iou_batch(detections_1.xyxy, detections_2.xyxy, match_metric)[0]
+            iou = box_iou_batch(detections_1.xyxy, detections_2.xyxy, overlap_metric)[0]
         if iou < threshold:
             break
         detections_1 = merge_inner_detection_object_pair(detections_1, detections_2)


SkalskiP:
Hi @SunHao-AI 👋🏻 thank you so much for your interest in supervision. This looks like a useful feature. I will try to get back to you with a PR review as soon as possible.

supervision/detection/utils.py
```diff
@@ -102,19 +129,36 @@ def _mask_iou_batch_split(
...
+    if match_metric.upper() == "IOU":
+        union_area = masks_true_area[:, None] + masks_detection_area - intersection_area
+        ious = np.divide(
+            intersection_area,
+            union_area,
+            out=np.zeros_like(intersection_area, dtype=float),
+            where=union_area != 0,
+        )
+    elif match_metric.upper() == "IOS":
+        small_area = np.minimum(intersection_area, masks_detection_area)
```
soumik12345, line 142: This will always result in `small_area = intersection_area` since intersection area is always <= any individual mask area. This makes the IOS calculation become `intersection_area / intersection_area = 1.0` for any overlapping masks, which is incorrect.

This should be fixed by the following sugegstion:
```suggestion
        small_area = np.minimum(masks_true_area[:, None], masks_detection_area)
```

I tested the fix with the following snippet:

```python
import cv2
import supervision as sv
from ultralytics import YOLO

model = YOLO("yolov8l-seg")
image = cv2.imread("./gettyimages-569175741-1.jpg")
results = model(image)[0]
detections = sv.Detections.from_ultralytics(results).with_nmm(match_metric="IOS")

box_annotator = sv.BoxAnnotator()

annotated_image = box_annotator.annotate(scene=image, detections=detections)
annotated_image = mask_annotator.annotate(scene=annotated_image, detections=detections)
annotated_image = label_annotator.annotate(scene=annotated_image, detections=detections)
```
Without the fix, this is how the annotations look 👇
annotated_image

With the fix, this is how the annotations look 👇
annotated_image

soumik12345 (changes requested):
Hi @SunHao-AI, thanks for the PR!
I found a bug in your implementation, can you please apply the fix?

supervision/detection/utils.py
SunHao-AI: My code was indeed wrong. Your modification is correct.

SunHao-AI: I have corrected the code and restructured the function "group_overlapping_masks",
```python
def group_overlapping_masks(
    predictions: np.ndarray,
    masks: np.ndarray,
    iou_threshold: float = 0.5,
    match_metric: str = "IOU",
) -> List[List[int]]:
    """
    Apply greedy version of non-maximum merging to avoid detecting too many

    Args:
        predictions (np.ndarray): An array of shape `(n, 5)` containing
            the bounding boxes coordinates in format `[x1, y1, x2, y2]`
            and the confidence scores.
        masks (np.ndarray): A 3D array of binary masks corresponding to the predictions.
        iou_threshold (float): The intersection-over-union threshold
            to use for non-maximum suppression. Defaults to 0.5.
        match_metric (str): Metric used for matching detections in slices.
            "IOU" or "IOS". Defaults "IOU".

    Returns:
        List[List[int]]: Groups of prediction indices be merged.
            Each group may have 1 or more elements.
    """
    merge_groups: List[List[int]] = []

    scores = predictions[:, 4]
    order = scores.argsort()

    while len(order) > 0:
        idx = int(order[-1])

        order = order[:-1]
        if len(order) == 0:
            merge_groups.append([idx])
            break

        merge_candidate = masks[idx][None, ...]
        candidate_groups = [idx]
        while len(order) > 0:
            ious = mask_iou_batch(masks[order], merge_candidate, match_metric)
            above_threshold: np.ndarray = ious.flatten() >= iou_threshold
            if not above_threshold.any():
                break

            above_idx = order[above_threshold]
            merge_candidate = np.logical_or.reduce(np.concatenate([masks[above_idx], merge_candidate]), axis=0, keepdims=True)
            candidate_groups.extend(np.flip(above_idx).tolist())
            order = order[~above_threshold]

        merge_groups.append(candidate_groups)
    return merge_groups
```
 and added the function "merge_inner_detections_objects_without_iou". 
```python
def merge_inner_detections_objects_without_iou(
    detections: List[Detections]
) -> Detections:
    """
    Given N detections each of length 1 (exactly one object inside), combine them into a
    single detection object of length 1. The contained inner object will be the merged
    result of all the input detections.

    For example, this lets you merge N boxes into one big box, N masks into one mask,
    etc.
    """
    return reduce(merge_inner_detection_object_pair, detections)
```
This is because the conditions such as IOU have already been considered when calculating the merge groups. Therefore, this condition does not need to be considered during the merging process (and there might be a situation, for example: threshold = 0.5, merge_group = [0,1,2,3,4], the IOU between 0 and [1,2,3,4] is [0.6,0.4,0.7,0.6], and after merging, some objects will be lost).

supervision/detection/core.py
```diff
@@ -1579,6 +1585,8 @@ def with_nms(
             class_agnostic (bool): Whether to perform class-agnostic
                 non-maximum suppression. If True, the class_id of each detection
                 will be ignored. Defaults to False.
+            match_metric (str): Metric used for matching detections in slices.
+                "IOU" or "IOS". Defaults "IOU".
```
SkalskiP, line 1589: This part of docstring should better explain what we are choosing between

supervision/detection/core.py
```diff
@@ -1566,7 +1569,10 @@ def box_area(self) -> np.ndarray:
         return (self.xyxy[:, 3] - self.xyxy[:, 1]) * (self.xyxy[:, 2] - self.xyxy[:, 0])
 
     def with_nms(
-        self, threshold: float = 0.5, class_agnostic: bool = False
+        self,
+        threshold: float = 0.5,
+        class_agnostic: bool = False,
+        match_metric: str = "IOU",
```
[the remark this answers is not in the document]
SkalskiP: I would even go as far as providing math formulas for `IOU` and `IOS`

SkalskiP (changes requested):

supervision/detection/utils.py
```diff
@@ -11,6 +14,42 @@
...
+        if isinstance(value, str):
+            value = value.lower()
+            try:
+                return cls(value)
+            except ValueError:
+                raise ValueError(f"Invalid value: {value}. Must be one of {cls.list()}")
+        raise ValueError(
+            f"Invalid value type: {type(value)}. Must be an instance of "
+            f"{cls.__name__} or str."
+        )
```
SkalskiP, line 50: this code should be in `supervision.detection.overlap_filter`

SkalskiP (approved):

soumik12345 (approved):

Issue #1778, opened 2025-01-20
Improving object detection benchmarking process for unrelated datasets


patel-zeel:
### Search before asking

- [x] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

##### Context
If a model is trained on a dataset, benchmarking with `supervision`'s evaluation API is straightforward.

However, if the dataset was not used to train the model (unrelated dataset), one should follow [the "how-to" guide on Benchmarking a model](https://supervision.roboflow.com/develop/how_to/benchmark_a_model/). 

What if we can improve this with a simpler API?

### Use case

In cases where the dataset was not used to train the model, people would need to: 1) remap the class names to match with the dataset; 2) transform the class IDs to match with the dataset. Here is what a boilerplate code would look like:

```py
class_mapping = {"dog": "animal", "cat": "animal", "eagle": "bird"}
for _, image, target in dataset:
    result = model.infer(image)[0]
    detection = sv.Detections.from_inference(result)
    
    # check if all mapped values are within the dataset classes
    if not all([value in dataset.classes for value in class_mapping.values()]):
        raise ValueError("All mapped values must be in dataset classes")
    
    # remap class names
    detection['class_name'] = list(map(lambda name: class_mapping[name] if name in class_maping else name, detection['class_name']))

    # remove predicted classes not in the dataset
    detection = detection[np.isin(detection['class_name'], dataset.classes)]

    # remap Class IDs based on Class names
    detection.class_id = np.array([dataset.classes.index(name) for name in detection['class_name']])
```

Proposed API for the above transformation:

```py
class_mapping = {"dog": "animal", "cat": "animal", "eagle": "bird"}
for _, image, target in dataset:
    result = model.infer(image)[0]
    detection = sv.Detections.from_inference(result)
    detection = detection.transform(dataset, class_mapping=class_mapping)
```

Where the `transform` method implements the same logic described in the boilerplate.

### Are you willing to submit a PR?

- [x] Yes I'd like to help by submitting a PR!

Yadav108:
Hey @patel-zeel, solid proposal! Before you start on the PR, a few quick design questions:

1. Naming: What about detection.align_to_dataset(dataset, class_mapping=...) instead of transform()? More explicit about what it does.
2. Optional mapping: If class_mapping=None, should we assume direct class name matching, or require it to be explicit?
3. Unmapped classes: If a detection has a class not in the mapping dict, keep it as-is or drop it?
[...]

PR #1779, opened 2025-01-23
Add transform method to Detections class and documentation for the transform method

nahcol10:
Add transform method to Detections class and documentation for the transform method

- Added the `transform` method to the `Detections` class to support remapping class names, filtering predictions, and remapping class IDs to match a dataset's classes.
- Added comprehensive documentation for the `transform` method in `docs/detections.md`, including a description, parameters, returns, raises, and an example usage.

### DIFF
--- a/docs/detection/detection.md
+++ b/docs/detection/detection.md
@@ -1,0 +1,48 @@
+### `transform`
+
+Transform detections to match the dataset's class names and IDs.
+
+This method performs the following steps:
+
+1. **Remaps class names** using the provided `class_mapping` dictionary.
+2. **Filters out predictions** that are not present in the dataset's classes.
+3. **Remaps class IDs** to match the dataset's class IDs.
+
+#### Parameters
+
+- **dataset**: The dataset object containing class names and IDs.
+- **class_mapping** (`Optional[Dict[str, str]]`): A dictionary to map model class names to dataset class names. If `None`, no remapping is performed.
+
+#### Returns
+
+- **Detections**: A new `Detections` object with transformed class names and IDs.
+
+#### Raises
+
+- **ValueError**: If the dataset does not contain the required class names.
+
+#### Example
+
+```python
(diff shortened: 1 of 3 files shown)


SkalskiP:
Hi @nahcol10 👋🏻 Thank you very much for your interest in supervision. Could you please clarify when the user would like to use the `transform` method?

nahcol10:
> Hi @nahcol10  👋🏻 Thank you very much for your interest in supervision. Could you please clarify when the user would like to use the `transform` method?

Hi @SkalskiP  👋🏻, thanks for your question!

The transform method is particularly useful when you need to remap detection class names and update class IDs to align with a specific dataset. This can be helpful in scenarios such as:

    Standardizing class names: When integrating detections from multiple sources, you can ensure consistency by mapping different class names to a unified set.

    Renaming classes: For example, mapping "dog" → "animal" or "eagle" → "bird" based on a predefined mapping.

    Filtering detections: Ensuring only valid detections (those with class names present in the dataset) are retained, improving consistency and reducing noise.

patel-zeel:
@nahcol10, probably you are trying to address #1778.

nahcol10:
> @nahcol10, are you trying to address #1778?
> 
> @SkalskiP Such a method may ease the process of benchmarking VLMs on Roboflow datasets because:
> 
>     1. VLMs by design can not predict `class_id` so we need to assign appropriate `class_id` to predictions to enable evaluation API and compute metrics such as mAP, Precision or Recall.
> 
>     2. By default, we can remove all predicted classes which are not present in the dataset. Optionally, we can have an argument in `transform` method to keep them if desired.
> 
>     3. In cases where we want to apply small modifications to predicted classes, `class_mapping` would come handy.

Yes, this method is intended to address #1778 by ensuring VLM predictions have appropriate class_id assignments, filtering out irrelevant predictions, and allowing for class modifications using class_mapping

PR #1820, opened 2025-04-15
feat(annotators): enhance label annotators with frame boundary adjust…

hidara2000:
# 🚀 Enhance label annotators with frame boundary adjustments and new base class

## Description

This PR adds the ability to ensure labels stay within frame boundaries through a new `ensure_in_frame` parameter. When enabled, this functionality guarantees that text labels for bounding boxes near image edges remain visible by adjusting their position to fit within the frame.

The key improvements include:
- ✅ Text labels near edges now properly positioned within frame boundaries
- ✅ Implemented as an optional parameter (default: `False` to maintain backward compatibility)
- ✅ Works alongside existing `smart_position` functionality with complementary behavior

*While there may be occasional label overlaps in very busy frames when both `smart_position` and `ensure_in_frame` are enabled, running the smart positioning algorithm first typically yields better results overall.*

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

## How has this change been tested?

I tested this change with various image scenarios that have bounding boxes positioned near frame edges. The implementation was verified by:

1. Comparing output images with and without the `ensure_in_frame` parameter enabled
2. Testing cases with multiple objects near edges to ensure proper positioning
3. Validating behavior when used in combination with `smart_position`

Example test code:
```python
import cv2
import numpy as np
import supervision as sv
from supervision.annotators.core import LabelAnnotator, BoxAnnotator
from PIL import Image, ImageDraw
import os
from typing import Optional, Tuple, List


def generate_mock_yolo_output(image_shape: Tuple[int, int, int]) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
    """
    Generates mock bounding box detections, confidence scores, and class predictions
    for a given image shape.  The function creates a set of detections, including
    one that covers the whole image.

    Args:
        image_shape (Tuple[int, int, int]): The shape of the image (height, width, channels).
            This is used to determine the boundaries for the generated bounding boxes.

    Returns:
        Tuple[np.ndarray, np.ndarray, np.ndarray]: A tuple containing:
            - A NumPy array of bounding boxes (N, 4), where N is the number of detections.
              Each box is defined as [xmin, ymin, xmax, ymax].
            - A NumPy array of confidence scores (N,).
            - A NumPy array of class labels (N,).
    """
    image_height, image_width, _ = image_shape
    num_detections = 100
    
    # Generate random bounding boxes
    xmin = np.random.randint(0, image_width, num_detections)
    ymin = np.random.randint(0, image_height, num_detections)
    xmax = np.random.randint(xmin + 20, image_width + 50, num_detections)
    ymax = np.random.randint(ymin + 20, image_height + 50, num_detections)
    bounding_boxes = np.stack([xmin, ymin, xmax, ymax], axis=1).astype(np.float32)

    # Add a box that covers the whole image
    full_image_box = np.array([0, 0, image_width, image_height], dtype=np.float32).reshape(1, 4)
    bounding_boxes = np.concatenate([bounding_boxes, full_image_box], axis=0)
    num_detections += 1

    # Generate random confidence scores
    confidence_scores = np.random.uniform(0.5, 0.95, num_detections).astype(np.float32)
    confidence_scores[-1] = 0.99  # High confidence for the full image box

    # Generate random class labels
    class_labels = np.random.randint(0, 2, num_detections).astype(np.int32)
    class_labels[-1] = 0  # Assign a class to the full image box

    return bounding_boxes, confidence_scores, class_labels



def process_image_with_supervision(
    image: np.ndarray,
    display_image: bool = True,
    text_position: sv.Position = sv.Position.TOP_LEFT,
    smart_position: bool = False,
    detections: Optional[sv.Detections] = None,
) -> None:
    """
    Processes an image by simulating YOLO detection and using Supervision to annotate it.
    The function generates two annotated images (with and without `ensure_in_frame`)
    and stacks them vertically, adding headers and white boundaries for clarity.

    Args:
        image (np.ndarray): The input image as a NumPy array in BGR format.
        display_image (bool, optional): Flag to control whether to display the image.
            If True, it attempts to display the image. If False, it saves the
            image to a file. Defaults to True.
        text_position (sv.Position, optional): The position of the text label
            relative to the bounding box.  Defaults to sv.Position.TOP_LEFT.
        smart_position (bool, optional): Flag to enable smart position adjustment of labels
            to keep them within the image frame. Defaults to False.
        detections (sv.Detections, optional): Pre-calculated detections.
            If provided, the function uses these detections instead of generating new ones.
            Defaults to None.

    Returns:
        None (displays or saves the stacked annotated image).
    """
    # 1. Simulate YOLO model output or use provided
    if detections is None:
        bounding_boxes, confidence_scores, class_labels = generate_mock_yolo_output(image.shape)
        detections = sv.Detections(
            xyxy=bounding_boxes,
            confidence=confidence_scores,
            class_id=class_labels,
        )

    # 2. Create annotators
    box_annotator = BoxAnnotator(thickness=2)
    class_names = ["car", "person"]

    label_annotator_in_frame = LabelAnnotator(
        text_scale=0.5,
        text_thickness=1,
        text_padding=5,
        ensure_in_frame=True,
        text_position=text_position,
        smart_position=smart_position,
    )
    label_annotator_out_of_frame = LabelAnnotator(
        text_scale=0.5,
        text_thickness=1,
        text_padding=5,
        ensure_in_frame=False,
        text_position=text_position,
        smart_position=smart_position,
    )

    # 4. Annotate the image with the detections using both annotators.
    annotated_image_in_frame = box_annotator.annotate(image.copy(), detections=detections)
    labels_in_frame = [
        f"{class_names[int(class_id)]} {confidence:.2f}"  # Corrected f-string
        for _, _, confidence, class_id, *_ in detections
    ]
    annotated_image_in_frame = label_annotator_in_frame.annotate(
        annotated_image_in_frame, detections=detections, labels=labels_in_frame
    )

    annotated_image_out_of_frame = box_annotator.annotate(image.copy(), detections=detections)
    labels_out_of_frame = [
        f"{class_names[int(class_id)]} {confidence:.2f}"  # Corrected f-string
        for _, _, confidence, class_id, *_ in detections
    ]
    annotated_image_out_of_frame = label_annotator_out_of_frame.annotate(
        annotated_image_out_of_frame, detections=detections, labels=labels_out_of_frame
    )

    # 5. Add white boundaries around the images
    border_width = 3
    annotated_image_in_frame = cv2.copyMakeBorder(
        annotated_image_in_frame,
        border_width,
        border_width,
        border_width,
        border_width,
        cv2.BORDER_CONSTANT,
        value=(255, 255, 255),
    )
    annotated_image_out_of_frame = cv2.copyMakeBorder(
        annotated_image_out_of_frame,
        border_width,
        border_width,
        border_width,
        border_width,
        cv2.BORDER_CONSTANT,
        value=(255, 255, 255),
    )

    # 6. Add headers to each image
    header_height = 30
    header_color = (255, 255, 255)
    text_color = (0, 0, 0)
    font = cv2.FONT_HERSHEY_SIMPLEX
    font_scale = 0.7
    font_thickness = 2

    # Create header images for each annotated image
    header_image_in_frame = np.zeros(
        (header_height, annotated_image_in_frame.shape[1], 3), dtype=np.uint8
    )
    header_image_in_frame[:] = header_color
    text_size_in_frame = cv2.getTextSize("Enabled", font, font_scale, font_thickness)[0]
    text_x_in_frame = annotated_image_in_frame.shape[1] - text_size_in_frame[0] - 10
    text_y_in_frame = (header_height + text_size_in_frame[1]) // 2
    cv2.putText(
        header_image_in_frame,
        "Enabled",
        (text_x_in_frame, text_y_in_frame),
        font,
        font_scale,
        text_color,
        font_thickness,
        cv2.LINE_AA,
    )

    header_image_out_of_frame = np.zeros(
        (header_height, annotated_image_out_of_frame.shape[1], 3), dtype=np.uint8
    )
    header_image_out_of_frame[:] = header_color
    text_size_out_of_frame = cv2.getTextSize("Not Enabled", font, font_scale, font_thickness)[0]
    text_x_out_of_frame = (header_image_out_of_frame.shape[1] - text_size_out_of_frame[0]) // 2
    text_y_out_of_frame = (header_height + text_size_out_of_frame[1]) // 2
    cv2.putText(
        header_image_out_of_frame,
        "Not Enabled",
        (text_x_out_of_frame, text_y_out_of_frame),
        font,
        font_scale,
        text_color,
        font_thickness,
        cv2.LINE_AA,
    )

    # Stack the headers and the images
    annotated_image_in_frame_with_header = np.vstack(
        (header_image_in_frame, annotated_image_in_frame)
    )
    annotated_image_out_of_frame_with_header = np.vstack(
        (header_image_out_of_frame, annotated_image_out_of_frame)
    )

    # 7. Stack the two images vertically
    stacked_image = np.vstack(
        (annotated_image_in_frame_with_header, annotated_image_out_of_frame_with_header)
    )

    # Add position text to the top-left corner
    cv2.putText(
        stacked_image,
        str(text_position) + f", smart_pos={smart_position}",
        (10, 20),
        cv2.FONT_HERSHEY_SIMPLEX,
        0.7,
        (0, 0, 0),
        2,
        cv2.LINE_AA,
    )

    # 8. Display the annotated image.
    if display_image:
        try:
            pil_image = Image.fromarray(cv2.cvtColor(stacked_image, cv2.COLOR_BGR2RGB))
            pil_image.show()
            pil_image.close()
        except OSError as e:
            print(f"Error displaying image: {e}. Saving image instead.")
            cv2.imwrite(f"annotated_image_{text_position}_smart_{smart_position}.jpg", stacked_image)
    else:
        cv2.imwrite(f"annotated_image_{text_position}_smart_{smart_position}.jpg", stacked_image)
        print(f"Annotated image saved to annotated_image_{text_position}_smart_{smart_position}.jpg")



def main(image_path: str = "example.jpg") -> None:
    """
    Main function to run the image processing and annotation with different label positions
    and smart position settings.

    Args:
        image_path (str, optional): Path to the image file. Defaults to "example.jpg".
    """
    # Create a dummy image
    image = np.zeros((600, 800, 3), dtype=np.uint8)
    cv2.imwrite(image_path, image)

    # 1. Generate Detections once - Moved inside process_image_with_supervision
    # mock_bounding_boxes, mock_confidence_scores, mock_class_labels = generate_mock_yolo_output(image.shape)
    # detections = sv.Detections(
    #      xyxy=mock_bounding_boxes,
    #      confidence=confidence_scores,
    #      class_id=mock_class_labels,
    # )

    # 2. Loop through positions with smart_position=False
    positions = [
        sv.Position.TOP_LEFT,
        sv.Position.CENTER_LEFT,
        sv.Position.BOTTOM_RIGHT,
        sv.Position.CENTER_RIGHT,
    ]
    for position in positions:
        print(f"Processing image with text position: {position}, smart_position=False")
        process_image_with_supervision(image, display_image=False, text_position=position, smart_position=False)  # Removed detections

    # 3. Loop through positions with smart_position=True, using the same detections
    for position in positions:
        print(f"Processing image with text position: {position}, smart_position=True")
        process_image_with_supervision(image, display_image=False, text_position=position, smart_position=True)  # Removed detections

    os.remove(image_path)



if __name__ == "__main__":
    main()

```
image
    
image
    
image
    
image

    

No special deployment considerations are needed. This feature is implemented as an optional parameter that defaults to `False`, ensuring backward compatibility with existing code.

## Docs

- [ ] Docs updated? What were the changes:
No changes to docs as functionality is similar to `smart_position` and the only entry for this in the docs was in the changelog. I can update the documentation to include this new parameter in the appropriate class references if desired, just let me know where and the format.

### DIFF
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -139,10 +139,10 @@ class _BaseLabelAnnotator(BaseAnnotator):
             # Update the properties with the spread out boxes
             adjusted_properties[:, :4] = spread_boxes
 
-            # Additional check to ensure boxes are still within frame after spreading
-            adjusted_properties[:, :4] = snap_boxes(
-                adjusted_properties[:, :4], resolution_wh
-            )
+        # Additional check to ensure boxes are still within frame after spreading
+        adjusted_properties[:, :4] = snap_boxes(
+            adjusted_properties[:, :4], resolution_wh
+        )
 
         return adjusted_properties
 
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -1397,7 +1397,7 @@ class RichLabelAnnotator(_BaseLabelAnnotator):
     @ensure_pil_image_for_annotation
     def annotate(
         self,
-        scene: Image.Image,
+        scene: ImageType
         detections: Detections,
         labels: Optional[List[str]] = None,
         custom_color_lookup: Optional[np.ndarray] = None,
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -1110,9 +1110,7 @@ class LabelAnnotator(_BaseLabelAnnotator):
         validate_labels(labels, detections)
 
         labels = get_labels_text(detections, labels)
-        label_properties = self._get_label_properties(
-            detections, labels
-        )
+        label_properties = self._get_label_properties(detections, labels)
 
         if self.smart_position:
             xyxy = label_properties[:, :4]
@@ -1176,10 +1174,12 @@ class LabelAnnotator(_BaseLabelAnnotator):
                 position=self.text_anchor,
             )
 
-            label_properties.append([
-                *text_background_xyxy,
-                total_height,
-            ])
+            label_properties.append(
+                [
+                    *text_background_xyxy,
+                    total_height,
+                ]
+            )
         return np.array(label_properties).reshape(-1, 5)
 
     def _draw_labels(
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -1,6 +1,6 @@
 from functools import lru_cache
 from math import sqrt
-from typing import Callable, List, Optional, Tuple, Union
+from typing import List, Optional, Tuple, Union
 
 import cv2
 import numpy as np
@@ -11,10 +11,14 @@ from supervision.annotators.base import BaseAnnotator, ImageType
 from supervision.annotators.utils import (
     ColorLookup,
     Trace,
+    get_labels_text,
     resolve_color,
     resolve_text_background_xyxy,
+    snap_boxes,
+    validate_labels,
+    wrap_text,
 )
-from supervision.config import CLASS_NAME_DATA_FIELD, ORIENTED_BOX_COORDINATES
+from supervision.config import ORIENTED_BOX_COORDINATES
 from supervision.detection.core import Detections
 from supervision.detection.utils import (
     clip_boxes,
@@ -74,23 +78,21 @@ class _BaseLabelAnnotator(BaseAnnotator):
 
         Args:
             color (Union[Color, ColorPalette], optional): The color to use for the label
-                                background. Defaults to ColorPalette.DEFAULT.
+                                background.
             text_color (Union[Color, ColorPalette], optional): The color to use for the
-                                label text. Defaults to Color.WHITE.
+                                label text.
             text_padding (int, optional): The padding around the label text, in pixels.
-                                Defaults to 10.
             text_position (Position, optional): The position of the text relative to the
-                                detection bounding box. Defaults to Position.TOP_LEFT.
+                                detection bounding box.
             color_lookup (ColorLookup, optional): The method used to determine the color
-                                of the label. Defaults to ColorLookup.CLASS.
+                                of the label
             border_radius (int, optional): The radius of the label background corners,
-                                in pixels. Defaults to 0.
+                                in pixels.
             smart_position (bool, optional): Whether to intelligently adjust the label
                                 position to avoid overlapping with other elements.
-                                Defaults to False.
             max_line_length (Optional[int], optional): Maximum number of characters per
                                 line before wrapping the text. None means no wrapping.
-                                Defaults to None.
+
         """
         self.color: Union[Color, ColorPalette] = color
         self.text_color: Union[Color, ColorPalette] = text_color
@@ -101,137 +103,21 @@ class _BaseLabelAnnotator(BaseAnnotator):
         self.smart_position = smart_position
         self.max_line_length: Optional[int] = max_line_length
 
-    def _validate_labels(self, labels: Optional[List[str]], detections: Detections):
-        """
-        Validates that the number of provided labels matches the number of detections.
-
-        Args:
-            labels (Optional[List[str]]): A list of labels, one for each detection. Can
-                                          be None.
-            detections (Detections): The detections to be labeled.
-
-        Raises:
-            ValueError: If `labels` is not None and its length does not match the number
-            of detections.
-        """
-        if labels is not None and len(labels) != len(detections):
-            raise ValueError(
-                f"The number of labels ({len(labels)}) does not match the "
-                f"number of detections ({len(detections)}). Each detection "
-                f"should have exactly 1 label."
-            )
-
-    @staticmethod
-    def _get_labels_text(
-        detections: Detections, custom_labels: Optional[List[str]]
-    ) -> List[str]:
-        """
-        Retrieves the text labels for the detections.
-
-        If `custom_labels` are provided, they are used. Otherwise, the labels are
-        extracted from the `detections` object, prioritizing the 'class_name' field,
-        then the `class_id`, and finally using the detection index as a string.
-
-        Args:
-            detections (Detections): The detections to get labels for.
-            custom_labels (Optional[List[str]]): An optional list of custom labels.
-
-        Returns:
-            List[str]: A list of text labels for each detection.
-        """
-        if custom_labels is not None:
-            return [str(label) for label in custom_labels]
-
-        labels = []
-        for idx in range(len(detections)):
-            if CLASS_NAME_DATA_FIELD in detections.data:
-                labels.append(detections.data[CLASS_NAME_DATA_FIELD][idx])
-            elif detections.class_id is not None:
-                labels.append(str(detections.class_id[idx]))
-            else:
-                labels.append(str(idx))
-        return labels
-
-    def _ensure_box_in_frame(
-        self,
-        box: np.ndarray,
-        frame_width: int,
-        frame_height: int,
-        check_flip_label: bool = False,
-        text_anchor: Optional[Position] = None,
-    ) -> np.ndarray:
-        """
-        Adjusts a box to ensure it stays within frame boundaries.
-
-        Args:
-            box: Box coordinates [x1, y1, x2, y2]
-            frame_width: Width of the frame
-            frame_height: Height of the frame
-            check_flip_label: Whether to check if label should be flipped
-            text_anchor: Text anchor position for flip check
-
-        Returns:
-            Adjusted box coordinates
-        """
-        x1, y1, x2, y2 = box
-
-        # Adjust x-coordinate to stay within frame
-        if x1 < 0:
-            shift = -x1
-            x1 += shift
-            x2 += shift
-        elif x2 > frame_width:
-            shift = frame_width - x2
-            x1 += shift
-            x2 += shift
-
-        # Adjust y-coordinate to stay within frame
-        if y1 < 0:
-            shift = -y1
-            y1 += shift
-            y2 += shift
-        elif y2 > frame_height:
-            shift = frame_height - y2
-            y1 += shift
-            y2 += shift
-
-            # Check if label should be flipped to above the box
-            if check_flip_label and text_anchor is not None:
-                box_height = y2 - y1
-
-                # Check anchor position to see if we can flip it
-                anchor_y = self._get_anchor_y_for_adjustment(
-                    np.array([y1, y2]), text_anchor
-                )
-
-                # If we're at the bottom, try moving to the top
-                if anchor_y >= y2 - 5:  # Near bottom edge
-                    # Check if there's room at the top
-                    if y1 - box_height >= 0:
-                        y2 = y1
-                        y1 = y2 - box_height
-
-        return np.array([x1, y1, x2, y2])
-
     def _adjust_labels_in_frame(
         self,
-        frame_width: int,
-        frame_height: int,
+        resolution_wh: Tuple[int, int],
         labels: List[str],
         label_properties: np.ndarray,
-        get_text_width_fn: Callable[[str], int],
     ) -> np.ndarray:
         """
         Adjusts the position of labels to ensure they stay within the frame boundaries.
 
         Args:
             frame_width (int): The width of the frame.
-            frame_height (int): The height of the frame.
+            resolution_wh (int, int): The width and height of the frame.
             labels (List[str]): The list of text labels.
             label_properties (np.ndarray): An array of label properties, where each row
                             contains [x1, y1, x2, y2, text_height, ...].
-            get_text_width_fn (Callable[[str], int]): A function that takes a label
-                            string and returns its width in pixels.
 
         Returns:
             np.ndarray: The adjusted label properties.
@@ -239,15 +125,10 @@ class _BaseLabelAnnotator(BaseAnnotator):
         adjusted_properties = label_properties.copy()
 
         # First, make sure the boxes don't go outside the frame
-        for i in range(len(labels)):
-            # Adjust box to stay within frame
-            adjusted_properties[i, :4] = self._ensure_box_in_frame(
-                adjusted_properties[i, :4],
-                frame_width,
-                frame_height,
-                check_flip_label=True,
-                text_anchor=self.text_anchor,
-            )
+        adjusted_properties[:, :4] = snap_boxes(
+            adjusted_properties[:, :4],
+            resolution_wh,
+        )
 
         # Apply the spread out algorithm to avoid box overlaps
         if len(labels) > 1:
@@ -259,86 +140,11 @@ class _BaseLabelAnnotator(BaseAnnotator):
             adjusted_properties[:, :4] = spread_boxes
 
             # Additional check to ensure boxes are still within frame after spreading
-            for i in range(len(labels)):
-                # Adjust box to stay within frame (without flipping)
-                adjusted_properties[i, :4] = self._ensure_box_in_frame(
-                    adjusted_properties[i, :4], frame_width, frame_height
-                )
-
-        return adjusted_properties
-
-    @staticmethod
-    def _get_anchor_y_for_adjustment(bbox_y: np.ndarray, anchor: Position) -> float:
-        """
-        Calculates the anchor y-coordinate for label adjustment based on the text anchor
-        position.
-
-        Args:
-            bbox_y (np.ndarray): An array containing the y1 and y2 coordinates of the
-                                bounding box.
-            anchor (Position): The desired text anchor position.
-
-        Returns:
-            float: The anchor y-coordinate.
-        """
-        y1, y2 = bbox_y
-        if anchor in [Position.TOP_LEFT, Position.TOP_CENTER, Position.TOP_RIGHT]:
-            return y1
-        elif anchor in [
-            Position.BOTTOM_LEFT,
-            Position.BOTTOM_CENTER,
-            Position.BOTTOM_RIGHT,
-        ]:
-            return y2
-        else:  # CENTER, CENTER_LEFT, CENTER_RIGHT
-            return (y1 + y2) / 2
-
-    def _wrap_text(self, text: str) -> List[str]:
-        """
-        Wraps text to the specified maximum line length, respecting existing newlines.
-        Uses the textwrap library for robust text wrapping.
-
-        Args:
-            text (str): The text to wrap.
-
-        Returns:
-            List[str]: A list of text lines after wrapping.
-        """
-        import textwrap
-
-        if not text:
-            return [""]
-
-        if self.max_line_length is None:
-            return text.splitlines() or [""]
-
-        # Split the text by existing newlines first
-        paragraphs = text.split("\n")
-        all_lines = []
-
-        for paragraph in paragraphs:
-            if not paragraph:
-                # Keep empty lines
-                all_lines.append("")
-                continue
-
-            # Wrap each paragraph separately
-            wrapped = textwrap.wrap(
-                paragraph,
-                width=self.max_line_length,
-                break_long_words=True,
-                replace_whitespace=False,
-                drop_whitespace=True,
+            adjusted_properties[:, :4] = snap_boxes(
+                adjusted_properties[:, :4], resolution_wh
             )
 
-            # Add the wrapped lines for this paragraph
-            if wrapped:
-                all_lines.extend(wrapped)
-            else:
-                # If wrap returns an empty list (e.g., for whitespace-only input)
-                all_lines.append("")
-
-        return all_lines if all_lines else [""]
+        return adjusted_properties
 
 
 class BoxAnnotator(BaseAnnotator):
@@ -1301,12 +1107,12 @@ class LabelAnnotator(_BaseLabelAnnotator):
         custom_color_lookup: Optional[np.ndarray] = None,
     ) -> np.ndarray:
         assert isinstance(scene, np.ndarray)
-        self._validate_labels(labels, detections)
+        validate_labels(labels, detections)
 
-        labels = self._get_labels_text(detections, labels)
+        labels = get_labels_text(detections, labels)
         label_properties = self._get_label_properties(
-            scene, detections, labels
-        )  # Pass scene
+            detections, labels
+        )
 
         if self.smart_position:
             xyxy = label_properties[:, :4]
@@ -1314,13 +1120,9 @@ class LabelAnnotator(_BaseLabelAnnotator):
             label_properties[:, :4] = xyxy
 
             label_properties = self._adjust_labels_in_frame(
-                scene.shape[1],
-                scene.shape[0],
+                (scene.shape[1], scene.shape[0]),
                 labels,
                 label_properties,
-                lambda text: cv2.getTextSize(
-                    text, CV2_FONT, self.text_scale, self.text_thickness
-                )[0][0],
             )
 
         self._draw_labels(
@@ -1335,7 +1137,6 @@ class LabelAnnotator(_BaseLabelAnnotator):
 
     def _get_label_properties(
         self,
-        scene: np.ndarray,  # Receive scene
         detections: Detections,
         labels: List[str],
     ) -> np.ndarray:
@@ -1345,7 +1146,7 @@ class LabelAnnotator(_BaseLabelAnnotator):
         ).astype(int)
 
         for label, center_coords in zip(labels, anchors_coordinates):
-            wrapped_lines = self._wrap_text(label)
+            wrapped_lines = wrap_text(label, self.max_line_length)
             line_heights = []
             line_widths = []
 
@@ -1375,12 +1176,10 @@ class LabelAnnotator(_BaseLabelAnnotator):
                 position=self.text_anchor,
             )
 
-            label_properties.append(
-                [
-                    *text_background_xyxy,
-                    total_height,
-                ]
-            )
+            label_properties.append([
+                *text_background_xyxy,
+                total_height,
+            ])
         return np.array(label_properties).reshape(-1, 5)
 
     def _draw_labels(
@@ -1427,7 +1226,7 @@ class LabelAnnotator(_BaseLabelAnnotator):
             )
 
             # Handle multiline text
-            wrapped_lines = self._wrap_text(labels[idx])
+            wrapped_lines = wrap_text(labels[idx], self.max_line_length)
             current_y = box_xyxy[1] + self.text_padding  # Start y position
 
             for line in wrapped_lines:
@@ -1550,10 +1349,10 @@ class RichLabelAnnotator(_BaseLabelAnnotator):
         custom_color_lookup: Optional[np.ndarray] = None,
     ) -> ImageType:
         assert isinstance(scene, Image.Image)
-        self._validate_labels(labels, detections)
+        validate_labels(labels, detections)
 
         draw = ImageDraw.Draw(scene)
-        labels = self._get_labels_text(detections, labels)
+        labels = get_labels_text(detections, labels)
         label_properties = self._get_label_properties(draw, detections, labels)
 
         if self.smart_position:
@@ -1562,14 +1361,9 @@ class RichLabelAnnotator(_BaseLabelAnnotator):
             label_properties[:, :4] = xyxy
 
             label_properties = self._adjust_labels_in_frame(
-                scene.width,
-                scene.height,
+                (scene.width, scene.height),
                 labels,
                 label_properties,
-                lambda text: int(
-                    draw.textbbox((0, 0), text, font=self.font)[2]
-                    - draw.textbbox((0, 0), text, font=self.font)[0]
-                ),
             )
 
         self._draw_labels(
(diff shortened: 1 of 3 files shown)


supervision/annotators/core.py
```diff
@@ -976,94 +1169,53 @@ def __init__(
...
+            border_radius=border_radius,
+            smart_position=smart_position,
+            ensure_in_frame=ensure_in_frame,
+        )
 
     @ensure_cv2_image_for_annotation
     def annotate(
         self,
-        scene: ImageType,
+        scene: np.ndarray,  # Ensure scene is initially a NumPy array here
```
onuralpszr, line 1190:
```suggestion
        scene: ImageType,  # Ensure scene is initially a NumPy array here
```

Please use ImageType we want to support both Pillow and Opencv (numpy arrays)

supervision/annotators/core.py
```diff
@@ -1258,80 +1378,30 @@ def __init__(
...
+            border_radius=border_radius,
+            smart_position=smart_position,
+            ensure_in_frame=ensure_in_frame,
+        )
 
     @ensure_pil_image_for_annotation
     def annotate(
         self,
-        scene: ImageType,
+        scene: Image.Image,
```
onuralpszr, line 1400:
```suggestion
        scene: ImageType
```

supervision/annotators/core.py
```diff
@@ -1258,80 +1378,30 @@ def __init__(
...
-            rich_label_annotator = sv.RichLabelAnnotator(font_path="path/to/font.ttf")
-            annotated_frame = label_annotator.annotate(
-                scene=image.copy(),
-                detections=detections,
-                labels=labels
-            )
-            ```
-
-        """
+    ) -> Image.Image:
```
onuralpszr, line 1404:
```suggestion
    ) -> ImageType:
```

onuralpszr (commented):
Hello @hidara2000 thank you for this awesome PR 

I made my first initials quick comments about certain change, Let me also test as well.

SkalskiP:
Hi @hidara2000 👋🏻 Huge thanks for deciding to submit a PR to introduce this change! I have a couple of points I'd like to discuss before I dive deeper into the PR review:

Wouldn't it be a better approach to keep the `smart_position` flag and simply add this extra behavior when `smart_position=True`? I understand that these two features could be seen as separate operations, but I'm still leaning towards maintaining a simple API:

- `smart_position=False` - raw, unprocessed label positions
- `smart_position=True` - we do everything we can to make them as visible as possible

For some time now, I've wanted to add support for multiline labels / label wrapping. Considering you're completely rewriting both label annotators, would you be willing to add support for multiline labels / label wrapping as part of this PR?

Screenshot 2025-03-31 at 12 22 16

hidara2000:
> Hi @hidara2000 👋🏻 Huge thanks for deciding to submit a PR to introduce this change! I have a couple of points I'd like to discuss before I dive deeper into the PR review:
> 
> Wouldn't it be a better approach to keep the `smart_position` flag and simply add this extra behavior when `smart_position=True`? I understand that these two features could be seen as separate operations, but I'm still leaning towards maintaining a simple API:
> 
> * `smart_position=False` - raw, unprocessed label positions
> * `smart_position=True` - we do everything we can to make them as visible as possible
> 
> For some time now, I've wanted to add support for multiline labels / label wrapping. Considering you're completely rewriting both label annotators, would you be willing to add support for multiline labels / label wrapping as part of this PR?
> 
> Screenshot 2025-03-31 at 12 22 16

Cheers for the response Piotr.

1. Great idea with the merging of smart_position and ensure_in_frame, Easily done.
2. Multiline. I suppose the first step would be to add support for \n and ensure padding is correct for each subsequent line. Easily done.
3. As for wrapping, not sure what I'd do here. 
    a. Simplest would be to use the inbuilt textwrap library where max_length is sent as a parameter eg.
``` python
import textwrap
...
wrapped_lines = textwrap.wrap(text, width=max_length , break_long_words=False)
return "\n".join(wrapped_lines)
...  
```    
thoughts?

I wanted to suggest a feature that I think would significantly improve the annotation experience. Currently, when you set formatting options (like colour palettes) for an annotator, those settings are applied to all the labels or classes that annotator works with.

What I'd love to see is the ability to associate different annotators with specific labels or classes. This way, you could define unique formatting, color palettes, and other visual cues for each class. Currently if I set a colour it is set for all classes. If I don't then the

This level of customization would be incredibly useful for various reasons, including:

Improved clarity: Quickly identify different object types based on their distinct visual appearance.
Enhanced collaboration: Different teams or annotators could have standardized visual representations for specific classes.
Better data analysis: Consistent color-coding could aid in the visual inspection and analysis of annotated data.
I understand that implementing this in the cleanest way might require a significant rewrite of the Detections class. However, I also thought a slightly less elegant but potentially quicker solution could involve allowing users to pass a dictionary of dataclasses as a parameter when initializing an annotation class. This dictionary could map labels to specific formatting configurations.

I'm curious to hear what others think about this idea and if anyone has any alternative suggestions for how this could be implemented. Thanks!




> Hi @hidara2000 👋🏻 Huge thanks for deciding to submit a PR to introduce this change! I have a couple of points I'd like to discuss before I dive deeper into the PR review:
> 
> Wouldn't it be a better approach to keep the `smart_position` flag and simply add this extra behavior when `smart_position=True`? I understand that these two features could be seen as separate operations, but I'm still leaning towards maintaining a simple API:
> 
> * `smart_position=False` - raw, unprocessed label positions
> * `smart_position=True` - we do everything we can to make them as visible as possible
> 
> For some time now, I've wanted to add support for multiline labels / label wrapping. Considering you're completely rewriting both label annotators, would you be willing to add support for multiline labels / label wrapping as part of this PR?
> 
> Screenshot 2025-03-31 at 12 22 16

Hi @SkalskiP 
I think this is easily doable. Let me have a look.

hidara2000:
Just an idea?

To maintain backwards compatibility whilst allowing users to set text and background colours per index, class, or track, could we extend the current architecture with a colour mapping capability?

Here's a simple approach:

1. **Create a Colour Mapping Class**:
    Create a class that maps indices/classes/tracks to specific colours.

2. **Extend the Current Parameters**:
    Allow the annotator classes to accept either the current colour options or the new mapping.

Here's how we could implement it:

```python
class ColorMap:
    """
    A class for mapping specific indices, classes, or track IDs to colours.

    This allows customising colours for specific elements whilst maintaining
    The default colour palette for others.
    """

    def __init__(self,
                 mapping: Dict[int, Union[Color, Tuple[int, int, int]]] = None,
                 default: Union[Color, ColorPalette] = ColorPalette.DEFAULT):
        """
        Initialize a colour map.

        Args:
            mapping: Dictionary mapping indices to Color objects or RGB tuples
            default: Default colour or colour palette to use for unmapped indices
        """
        self.mapping = {}
        self.default = default

        if mapping:
            for idx, colour in mapping.items():
                if isinstance(colour, tuple) and len(colour) == 3:
                    self.mapping[idx] = Color(*colour)
                else:
                    self.mapping[idx] = colour

    def get_colour(self, idx: int) -> Color:
        """Get colour for the given index, or use default if not found."""
        if idx in self.mapping:
            return self.mapping[idx]

        # Fall back to default colour/palette
        if isinstance(self.default, ColorPalette):
            return self.default.get(idx)
        return self.default
```

Then modify the`resolve_color` function to handle this new type:

```python
def resolve_color(
    colour: Union[Color, ColorPalette, ColorMap],
    detections: Detections,
    detection_idx: int,
    colour_lookup: Union[ColorLookup, np.ndarray] = ColorLookup.CLASS,
) -> Color:
    idx = resolve_color_idx(
        detections=detections,
        detection_idx=detection_idx,
        colour_lookup=colour_lookup,
    )

    if isinstance(colour, ColorMap):
        return colour.get_colour(idx)

    return get_colour_by_index(colour=colour, idx=idx)
```

Finally, update our annotator classes to accept `ColorMap` for both `colour` and `text_colour` parameters:

```python
class LabelAnnotator(_BaseLabelAnnotator):
    def __init__(
        self,
        colour: Union[Color, ColorPalette, ColorMap] = ColorPalette.DEFAULT,
        text_colour: Union[Color, ColorPalette, ColorMap] = Color.WHITE,
        # ...other parameters...
    ):
        # ...existing initialization...
```

This approach:
1. Maintains complete backwards compatibility
2. Gives users fine-grained control over specific indices/classes/tracks
3. Falls back to default behaviour for unmapped indices
4. Allows users to override default colour choices in case they are too similar to image background

Usage would be:

```python
# Map specific classes to colours (class ID 0 = red, class ID 1 = blue)
colour_map = ColorMap({0: Color.RED, 1: Color.BLUE})
text_colour_map = ColorMap({0: Color.WHITE, 1: Color.YELLOW})

# Use in annotator
annotator = LabelAnnotator(
    colour=colour_map,
    text_colour=text_colour_map
)
```

supervision/annotators/core.py
```diff
@@ -40,6 +40,307 @@
...
+        Wraps text to the specified maximum line length, respecting existing newlines.
+        Uses the textwrap library for robust text wrapping.
+
+        Args:
+            text (str): The text to wrap.
+
+        Returns:
+            List[str]: A list of text lines after wrapping.
+        """
+        import textwrap
```
SkalskiP, line 307: Let’s move this import to the top of the file instead of placing it here.

supervision/annotators/core.py
```diff
@@ -40,6 +40,307 @@
...
+        elif anchor in [
+            Position.BOTTOM_LEFT,
+            Position.BOTTOM_CENTER,
+            Position.BOTTOM_RIGHT,
+        ]:
+            return y2
+        else:  # CENTER, CENTER_LEFT, CENTER_RIGHT
+            return (y1 + y2) / 2
+
+    def _wrap_text(self, text: str) -> List[str]:
```
SkalskiP, line 296: I’d prefer this not to be a private class method—let’s move it to `supervision/annotators/utils.py` instead.

supervision/annotators/core.py
```diff
@@ -40,6 +40,307 @@
...
+
+        Returns:
+            np.ndarray: The adjusted label properties.
+        """
+        adjusted_properties = label_properties.copy()
+
+        # First, make sure the boxes don't go outside the frame
+        for i in range(len(labels)):
+            # Adjust box to stay within frame
+            adjusted_properties[i, :4] = self._ensure_box_in_frame(
```
SkalskiP, line 244: Given that we are getting rid of flipping for now, do we need to call `_ensure_box_in_frame` (`snap_boxes`) twice?

supervision/detection/utils.py
```diff
@@ -1274,36 +1277,59 @@ def cross_product(anchors: np.ndarray, vector: Vector) -> np.ndarray:
 def spread_out_boxes(
     xyxy: np.ndarray,
     max_iterations: int = 100,
+    force_scale: float = 10.0,
+    consider_size: bool = True,
```
SkalskiP, line 1281: @hidara2000 I'm curious—what was the reason for introducing those two arguments? I'm a bit concerned they might lead to unstable label positions during video processing, where small changes in initial position could cause disproportionately large shifts in the final output.

SkalskiP (changes requested):

SkalskiP:
Hi @hidara2000, sorry it took me a while to get back to you. I'm currently juggling work across 3–4 repositories, so my time is a bit stretched. I’ve now gone through your PR carefully and you’ve done an excellent job—really impressive work! Don’t be discouraged by the number of comments I left—they’re all meant to help polish things up. Once we merge this PR, it’ll take Supervision’s text annotators to the next level!

supervision/detection/utils.py
hidara2000: @SkalskiP 
The reason for introducing force_scale and consider_size was primarily to offer more granular control over how the spreading algorithm resolves overlaps in static images (during testing). force_scale allows tuning the overall repulsion strength, and consider_size was an attempt to see if factoring in the label box dimensions could lead to a more visually pleasing distribution, especially in complex overlap scenarios.

You're absolutely right to be concerned about video stability. Iterative algorithms like spread_out_boxes can be sensitive to small frame-to-frame variations in detection positions. Parameters like force_scale (especially if set high) and consider_size can amplify these small variations into larger, potentially noticeable jumps or jitter in the label positions across consecutive video frames.

Given this valid concern and the potential for these parameters to introduce instability, I've reverted the spread_out_boxes function in the PR back to the original version that doesn't have these parameters.

I'm still interested to know your thoughts though – do you think there's a viable way to use the version of the function (below) with force_scale and consider_size without causing instability (perhaps with very conservative default values)? Or do the added parameters introduce unnecessary complexity that would require users to tune them during class instantiation, which might not be ideal for a general-purpose annotator?

Looking forward to your feedback!

``` python
def alternative_spread_out_boxes(
    xyxy: np.ndarray,
    max_iterations: int = 50,  # Moderate default iterations
    force_scale: float = 5.0,  # Moderate default force scale
    consider_size: bool = False, # Default to False for better video stability
    min_force_magnitude: float = 2.0 # Make minimum force tunable
) -> np.ndarray:
    """
    Spread out boxes that overlap with each other, optimized for a balance
    between overlap resolution and video stability.

    Args:
        xyxy: Numpy array of shape (N, 4) where N is the number of boxes.
        max_iterations: Maximum number of iterations to run the algorithm for.
                        Lower values may improve performance and stability
                        but could leave some overlaps unresolved.
        force_scale: Scale factor for the repulsion forces. Lower values result
                     in less aggressive spreading, which can improve video stability.
        consider_size: Whether to consider box size when calculating forces.
                       Setting to True might yield better static layouts but can
                       increase jitter in video due to fluctuating box sizes.
                       Defaults to False for better video stability.
        min_force_magnitude: Minimum magnitude for calculated force vectors.
                             Ensures slight overlaps still result in movement.

    Returns:
        np.ndarray: A numpy array of shape (N, 4) with adjusted box positions.
    """
    if len(xyxy) == 0:
        return xyxy

    # Add a small padding to ensure boxes that are just touching are considered for overlap
    xyxy_padded = pad_boxes(xyxy, px=1)

    # Calculate box areas if we're considering size (only done once)
    size_factors = np.ones(len(xyxy_padded))
    if consider_size:
        box_areas = (xyxy_padded[:, 2] - xyxy_padded[:, 0]) * (
            xyxy_padded[:, 3] - xyxy_padded[:, 1]
        )
        # Calculate the size factors (normalize by mean size), handle empty box_areas
        if len(box_areas) > 0 and np.mean(np.sqrt(box_areas)) != 0:
             size_factors = np.sqrt(box_areas) / np.mean(np.sqrt(box_areas))
             # Clip to avoid extreme values influencing forces too much
             size_factors = np.clip(size_factors, 0.5, 2.0)


    for _ in range(max_iterations):
        # Calculate IoU between all pairs of boxes (NxN matrix)
        iou = box_iou_batch(xyxy_padded, xyxy_padded)
        np.fill_diagonal(iou, 0)  # Eliminate self-interactions (a box doesn't overlap with itself)

        # If there are no overlaps, we are done
        if np.all(iou == 0):
            break

        overlap_mask = iou > 0

        # Calculate centers of the boxes (Nx2)
        centers = (xyxy_padded[:, :2] + xyxy_padded[:, 2:]) / 2

        # Calculate vectors pointing from each box center to every other box center (NxNx2)
        delta_centers = centers[:, np.newaxis, :] - centers[np.newaxis, :, :]
        # Only consider deltas for overlapping boxes
        delta_centers *= overlap_mask[:, :, np.newaxis]

        # Sum the delta vectors for each box to get the total push direction (Nx2)
        delta_sum = np.sum(delta_centers, axis=1)

        # Normalize the sum of deltas to get direction vectors (unit vectors)
        delta_magnitude = np.linalg.norm(delta_sum, axis=1, keepdims=True)
        direction_vectors = np.divide(
            delta_sum,
            delta_magnitude,
            out=np.zeros_like(delta_sum), # Use zeros where magnitude is zero to avoid NaNs
            where=delta_magnitude != 0,
        )

        # Calculate the base force magnitude based on total overlap (sum of IoUs)
        base_force_magnitude = np.sum(iou, axis=1)
        force_vectors = base_force_magnitude[:, np.newaxis] * direction_vectors

        # Apply size-based scaling if enabled
        if consider_size:
             force_vectors *= size_factors[:, np.newaxis]

        # Apply the general force scale
        force_vectors *= force_scale

        # Ensure minimum force for small overlaps to guarantee separation
        current_force_magnitudes = np.linalg.norm(force_vectors, axis=1, keepdims=True)
        small_force_mask = (current_force_magnitudes > 0) & (current_force_magnitudes < min_force_magnitude)

        if np.any(small_force_mask):
             # Rescale small force vectors to have the minimum magnitude
             force_directions_for_small = force_vectors / np.where(
                 current_force_magnitudes > 0, current_force_magnitudes, 1
             )
             force_vectors = np.where(
                 small_force_mask, force_directions_for_small * min_force_magnitude, force_vectors
             )

        # Convert displacement vectors to integers for pixel-based movement
        force_vectors = force_vectors.astype(int)

        # Apply forces to update box positions (shift both corners by the same vector)
        xyxy_padded[:, [0, 1]] += force_vectors
        xyxy_padded[:, [2, 3]] += force_vectors

    # Remove the padding before returning
    return pad_boxes(xyxy_padded, px=-1)
```

hidara2000:
> Hi @hidara2000, sorry it took me a while to get back to you. I'm currently juggling work across 3–4 repositories, so my time is a bit stretched. I’ve now gone through your PR carefully and you’ve done an excellent job—really impressive work! Don’t be discouraged by the number of comments I left—they’re all meant to help polish things up. Once we merge this PR, it’ll take Supervision’s text annotators to the next level!

I appreciate you going through it, and I agree with all the comments. Changes made as per advice and results from test below. 
<br>
     
    
image    
image    
image    
image

supervision/annotators/core.py
hidara2000: 🤦

supervision/annotators/core.py
hidara2000: Done

supervision/annotators/core.py
```diff
@@ -40,6 +40,307 @@
...
+            )
+
+            # Add the wrapped lines for this paragraph
+            if wrapped:
+                all_lines.extend(wrapped)
+            else:
+                # If wrap returns an empty list (e.g., for whitespace-only input)
+                all_lines.append("")
+
+        return all_lines if all_lines else [""]
```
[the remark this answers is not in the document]
hidara2000: Done

supervision/annotators/core.py
```diff
@@ -40,6 +40,307 @@
...
+            color_lookup (ColorLookup, optional): The method used to determine the color
+                                of the label. Defaults to ColorLookup.CLASS.
+            border_radius (int, optional): The radius of the label background corners,
+                                in pixels. Defaults to 0.
+            smart_position (bool, optional): Whether to intelligently adjust the label
+                                position to avoid overlapping with other elements.
+                                Defaults to False.
+            max_line_length (Optional[int], optional): Maximum number of characters per
+                                line before wrapping the text. None means no wrapping.
+                                Defaults to None.
```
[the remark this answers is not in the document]
hidara2000: Done

supervision/annotators/core.py
```diff
@@ -40,6 +40,307 @@
...
+                labels.append(str(detections.class_id[idx]))
+            else:
+                labels.append(str(idx))
+        return labels
+
+    def _ensure_box_in_frame(
+        self,
+        box: np.ndarray,
+        frame_width: int,
+        frame_height: int,
```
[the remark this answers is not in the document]
hidara2000: Done. Might be worth double checking that I understood you properly here

supervision/annotators/core.py
```diff
@@ -40,6 +40,307 @@
...
+
+        # Adjust y-coordinate to stay within frame
+        if y1 < 0:
+            shift = -y1
+            y1 += shift
+            y2 += shift
+        elif y2 > frame_height:
+            shift = frame_height - y2
+            y1 += shift
+            y2 += shift
```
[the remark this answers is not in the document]
hidara2000: done

supervision/annotators/core.py
```diff
@@ -40,6 +40,307 @@
...
+                    if y1 - box_height >= 0:
+                        y2 = y1
+                        y1 = y2 - box_height
+
+        return np.array([x1, y1, x2, y2])
+
+    def _adjust_labels_in_frame(
+        self,
+        frame_width: int,
+        frame_height: int,
```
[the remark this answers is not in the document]
hidara2000: easy

supervision/annotators/core.py
```diff
@@ -40,6 +40,307 @@
...
+        self.color: Union[Color, ColorPalette] = color
+        self.text_color: Union[Color, ColorPalette] = text_color
+        self.text_padding: int = text_padding
+        self.text_anchor: Position = text_position
+        self.color_lookup: ColorLookup = color_lookup
+        self.border_radius: int = border_radius
+        self.smart_position = smart_position
+        self.max_line_length: Optional[int] = max_line_length
+
+    def _validate_labels(self, labels: Optional[List[str]], detections: Detections):
```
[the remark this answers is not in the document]
hidara2000: done

supervision/annotators/core.py
```diff
@@ -40,6 +40,307 @@
...
+        """
+        if labels is not None and len(labels) != len(detections):
+            raise ValueError(
+                f"The number of labels ({len(labels)}) does not match the "
+                f"number of detections ({len(detections)}). Each detection "
+                f"should have exactly 1 label."
+            )
+
+    @staticmethod
+    def _get_labels_text(
```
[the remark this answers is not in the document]
hidara2000: done

supervision/annotators/core.py
hidara2000: fixed

supervision/annotators/core.py
```diff
@@ -40,6 +40,307 @@
...
+                anchor_y = self._get_anchor_y_for_adjustment(
+                    np.array([y1, y2]), text_anchor
+                )
+
+                # If we're at the bottom, try moving to the top
+                if anchor_y >= y2 - 5:  # Near bottom edge
+                    # Check if there's room at the top
+                    if y1 - box_height >= 0:
+                        y2 = y1
+                        y1 = y2 - box_height
```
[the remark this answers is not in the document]
hidara2000: done

supervision/annotators/core.py
```diff
@@ -1293,57 +1348,24 @@ def annotate(
...
         if self.smart_position:
             xyxy = label_properties[:, :4]
             xyxy = spread_out_boxes(xyxy)
             label_properties[:, :4] = xyxy
 
+            label_properties = self._adjust_labels_in_frame(
+                (scene.width, scene.height),
+                labels,
+                label_properties,
+            )
```
soumik12345, line 1367: The `spread_out_boxes` function is being called twice, once in L1360 and again inside the `self._adjust_labels_in_frame` function, and I think the 1st time its being called can be avoided; this is because, in `self._adjust_labels_in_frame` the boxes are first snapped to the frame and then `spread_out_boxes` is called again on the snapped boxes which effectively discards the effect of the 1st call.

soumik12345 (changes requested):
Hi @hidara2000, great work on the PR!
I left one minor feedback regarding a potential redundant function call.

supervision/annotators/core.py
hidara2000: ## FIX Summary: Remove Redundant `spread_out_boxes` Calls

### Problem
The codebase contained redundant calls to `spread_out_boxes()` that were immediately overwritten by subsequent operations, resulting in unnecessary processing.

### Analysis
The redundant pattern occurred in two locations:
- **Line 1360**: `spread_out_boxes()` called before `_adjust_labels_in_frame()`
- **Line 1117**: Same redundant pattern

The issue is that `_adjust_labels_in_frame()` already contains the complete label adjustment workflow:
1. Snaps boxes to frame boundaries (line 128)
2. Spreads out boxes to avoid overlaps (line 138)
3. Re-snaps boxes after spreading (line 143)

The initial `spread_out_boxes()` calls were pointless because:
- Their results were overwritten by the snapping operation inside `_adjust_labels_in_frame()`
- The actual spreading that affects the final output happens inside `_adjust_labels_in_frame()`

### Changes Made
Removed redundant code at two locations:

**Lines 1115-1120:**
- Removed: `xyxy = label_properties[:, :4]`
- Removed: `xyxy = spread_out_boxes(xyxy)`
- Removed: `label_properties[:, :4] = xyxy`
- Replaced with direct call to `_adjust_labels_in_frame()`

**Lines 1354-1359:**
- Removed: `xyxy = label_properties[:, :4]`
- Removed: `xyxy = spread_out_boxes(xyxy)`
- Removed: `label_properties[:, :4] = xyxy`
- Replaced with direct call to `_adjust_labels_in_frame()`

### Impact
- Eliminates redundant processing
- Maintains identical functionality
- Improves code clarity by removing misleading operations

SkalskiP (approved):

PR #1834, opened 2025-04-28
Fix/mAP

rafaelpadilla:
# Description

The current implementation of **Mean Average Precision (mAP)** in `supervision.metrics` produces results that diverge from both **pycocotools** and published benchmarks (e.g., the [Roboflow leaderboard](https://leaderboard.roboflow.com/)).

This PR **aligns supervision's mAP implementation with pycocotools**, the official COCO evaluation tool, ensuring reliable, standardized metrics. 🚀

## Key points 🏆
* **Self-contained**: No external dependencies are needed to compute the metric. 
* **Seamless integration**: Metrics are computed directly in `supervision/metrics/mean_average_precision.py` and works directly with `supervision.Detections` objects.
* **Unchanged public API**: Existing user code continues to work with the same interface. **No impact for users!** 🔥
```
import supervision as sv
from supervision.metrics import MeanAveragePrecision
preds   = sv.Detections(...)
targets = sv.Detections(...)
metric  = MeanAveragePrecision()
result  = metric.update(preds, targets).compute()
print(result.map50_95)

```

* **Same result object**: `MeanAveragePrecisionResult` class still exposes the same properties:
```
print("Overall:")
print(f"map50_95: {result.map50_95}")
print(f"map50:     {result.map50}")
print(f"map75:     {result.map75}")
print(f"map50_95: {result.small_objects.map50_95}")
print(f"map50_95: {result.medium_objects.map50_95}")
print(f"map50_95: {result.large_objects.map50_95}")
```
* **Numerical parity**: Evaluations match pycocotools within < 2e-5. ✅
    * A Colab notebook was created to compute mAP results for RT-Detr models **purely with pycocotools** [here](https://colab.research.google.com/drive/1seSGLmmJvQJ4JbXkjt_nY32UuLUXayRa?usp=sharing).
    * Another Colab notebook was created to compute mAP results for RT-Detr models using **this proposed branch** [here](https://colab.research.google.com/drive/1fFjMC9_K4z4VTYC7mFlsmIR4YyV_Dhgx?usp=sharing).
    * A detailed results table comparing the results of both Colab notebooks (pure pycocotools vs. this PR) is [here](https://docs.google.com/spreadsheets/d/1w8MpCgJuv6vZQfNTbmzNgitKvJwynuLpB9Ioqxkm-SI/edit?gid=0), showing a maximum deviation of 2e-5. 🎯

## Type of change

### DIFF
--- a/supervision/dataset/formats/coco.py
+++ b/supervision/dataset/formats/coco.py
@@ -105,6 +105,7 @@ def coco_annotations_to_detections(
     xyxy = np.asarray(xyxy)
     xyxy[:, 2:4] += xyxy[:, 0:2]
 
+    data = dict()
     if use_iscrowd:
         iscrowd = [
             image_annotation["iscrowd"] for image_annotation in image_annotations
@@ -113,10 +114,6 @@ def coco_annotations_to_detections(
         data = dict(
             iscrowd=np.asarray(iscrowd, dtype=int), area=np.asarray(area, dtype=float)
         )
-    else:
-        iscrowd = [0] * len(image_annotations)
-        area = None
-        data = dict()
 
     if with_masks:
         mask = coco_annotations_to_masks(
@@ -182,14 +179,15 @@ def get_coco_class_index_mapping(annotations_path: str) -> Dict[int, int]:
     class_mapping = build_coco_class_index_mapping(
         coco_categories=coco_data["categories"], target_classes=classes
     )
-    return class_mapping
+    inv_class_mapping = {v: k for k, v in class_mapping.items()}
+    return inv_class_mapping
 
 
 def load_coco_annotations(
     images_directory_path: str,
     annotations_path: str,
     force_masks: bool = False,
-    use_iscrowd: bool = False,
+    # use_iscrowd: bool = True,
 ) -> Tuple[List[str], List[str], Dict[str, Detections]]:
     coco_data = read_json_file(file_path=annotations_path)
     classes = coco_categories_to_classes(coco_categories=coco_data["categories"])
@@ -219,7 +217,7 @@ def load_coco_annotations(
             image_annotations=image_annotations,
             resolution_wh=(image_width, image_height),
             with_masks=force_masks,
-            use_iscrowd=use_iscrowd,
+            use_iscrowd=True,
         )
 
         annotation = map_detections_class_id(
--- a/supervision/dataset/formats/coco.py
+++ b/supervision/dataset/formats/coco.py
@@ -94,7 +94,6 @@ def coco_annotations_to_detections(
     resolution_wh: Tuple[int, int],
     with_masks: bool,
     use_iscrowd: bool = False,
-    use_precomputed_area: bool = False,
 ) -> Detections:
     if not image_annotations:
         return Detections.empty()
@@ -110,19 +109,13 @@ def coco_annotations_to_detections(
         iscrowd = [
             image_annotation["iscrowd"] for image_annotation in image_annotations
         ]
-    else:
-        iscrowd = [0] * len(image_annotations)
-
-    if use_precomputed_area:
         area = [image_annotation["area"] for image_annotation in image_annotations]
-    else:
-        area = None
-
-    if use_iscrowd or use_precomputed_area:
         data = dict(
             iscrowd=np.asarray(iscrowd, dtype=int), area=np.asarray(area, dtype=float)
         )
     else:
+        iscrowd = [0] * len(image_annotations)
+        area = None
         data = dict()
 
     if with_masks:
@@ -197,7 +190,6 @@ def load_coco_annotations(
     annotations_path: str,
     force_masks: bool = False,
     use_iscrowd: bool = False,
-    use_precomputed_area: bool = False,
 ) -> Tuple[List[str], List[str], Dict[str, Detections]]:
     coco_data = read_json_file(file_path=annotations_path)
     classes = coco_categories_to_classes(coco_categories=coco_data["categories"])
@@ -228,7 +220,6 @@ def load_coco_annotations(
             resolution_wh=(image_width, image_height),
             with_masks=force_masks,
             use_iscrowd=use_iscrowd,
-            use_precomputed_area=use_precomputed_area,
         )
 
         annotation = map_detections_class_id(
--- a/supervision/dataset/formats/coco.py
+++ b/supervision/dataset/formats/coco.py
@@ -90,7 +90,11 @@ def coco_annotations_to_masks(
 
 
 def coco_annotations_to_detections(
-    image_annotations: List[dict], resolution_wh: Tuple[int, int], with_masks: bool
+    image_annotations: List[dict],
+    resolution_wh: Tuple[int, int],
+    with_masks: bool,
+    use_iscrowd: bool = False,
+    use_precomputed_area: bool = False,
 ) -> Detections:
     if not image_annotations:
         return Detections.empty()
@@ -102,15 +106,32 @@ def coco_annotations_to_detections(
     xyxy = np.asarray(xyxy)
     xyxy[:, 2:4] += xyxy[:, 0:2]
 
+    if use_iscrowd:
+        iscrowd = [
+            image_annotation["iscrowd"] for image_annotation in image_annotations
+        ]
+    else:
+        iscrowd = [0] * len(image_annotations)
+
+    if use_precomputed_area:
+        area = [image_annotation["area"] for image_annotation in image_annotations]
+    else:
+        area = None
+
+    data = dict(
+        iscrowd=np.asarray(iscrowd, dtype=int), area=np.asarray(area, dtype=float)
+    )
+
     if with_masks:
         mask = coco_annotations_to_masks(
             image_annotations=image_annotations, resolution_wh=resolution_wh
         )
-        return Detections(
-            class_id=np.asarray(class_ids, dtype=int), xyxy=xyxy, mask=mask
-        )
+    else:
+        mask = None
 
-    return Detections(xyxy=xyxy, class_id=np.asarray(class_ids, dtype=int))
+    return Detections(
+        class_id=np.asarray(class_ids, dtype=int), xyxy=xyxy, mask=mask, data=data
+    )
 
 
 def detections_to_coco_annotations(
@@ -159,16 +180,29 @@ def detections_to_coco_annotations(
     return coco_annotations, annotation_id
 
 
+def get_coco_class_index_mapping(annotations_path: str) -> Dict[int, int]:
+    coco_data = read_json_file(annotations_path)
+    classes = coco_categories_to_classes(coco_categories=coco_data["categories"])
+    class_mapping = build_coco_class_index_mapping(
+        coco_categories=coco_data["categories"], target_classes=classes
+    )
+    return class_mapping
+
+
 def load_coco_annotations(
     images_directory_path: str,
     annotations_path: str,
     force_masks: bool = False,
+    use_iscrowd: bool = False,
+    use_precomputed_area: bool = False,
 ) -> Tuple[List[str], List[str], Dict[str, Detections]]:
     coco_data = read_json_file(file_path=annotations_path)
     classes = coco_categories_to_classes(coco_categories=coco_data["categories"])
+
     class_index_mapping = build_coco_class_index_mapping(
         coco_categories=coco_data["categories"], target_classes=classes
     )
+
     coco_images = coco_data["images"]
     coco_annotations_groups = group_coco_annotations_by_image_id(
         coco_annotations=coco_data["annotations"]
@@ -190,7 +224,10 @@ def load_coco_annotations(
             image_annotations=image_annotations,
             resolution_wh=(image_width, image_height),
             with_masks=force_masks,
+            use_iscrowd=use_iscrowd,
+            use_precomputed_area=use_precomputed_area,
         )
+
         annotation = map_detections_class_id(
             source_to_target_mapping=class_index_mapping,
             detections=annotation,
--- a/supervision/dataset/formats/coco.py
+++ b/supervision/dataset/formats/coco.py
@@ -118,9 +118,12 @@ def coco_annotations_to_detections(
     else:
         area = None
 
-    data = dict(
-        iscrowd=np.asarray(iscrowd, dtype=int), area=np.asarray(area, dtype=float)
-    )
+    if use_iscrowd or use_precomputed_area:
+        data = dict(
+            iscrowd=np.asarray(iscrowd, dtype=int), area=np.asarray(area, dtype=float)
+        )
+    else:
+        data = dict()
 
     if with_masks:
         mask = coco_annotations_to_masks(
--- a/supervision/detection/utils.py
+++ b/supervision/detection/utils.py
@@ -1369,7 +1369,9 @@ def _jaccard(box_a: List[float], box_b: List[float], is_crowd: bool) -> float:
 
 
 def iou_with_jaccard(
-    dt: List[List[float]], gt: List[List[float]], is_crowd: List[bool]
+    boxes_true: List[List[float]],
+    boxes_detection: List[List[float]],
+    is_crowd: List[bool],
 ) -> np.ndarray:
     """
     Calculate the intersection over union (IoU) between detection bounding boxes (dt)
@@ -1377,9 +1379,9 @@ def iou_with_jaccard(
     Reference: https://github.com/rafaelpadilla/review_object_detection_metrics
 
     Args:
-        dt (List[List[float]]): List of detection bounding boxes in the \
+        boxes_true (List[List[float]]): List of ground-truth bounding boxes in the \
             format [x, y, width, height].
-        gt (List[List[float]]): List of ground-truth bounding boxes in the \
+        boxes_detection (List[List[float]]): List of detection bounding boxes in the \
             format [x, y, width, height].
         is_crowd (List[bool]): List indicating if each ground-truth bounding box \
             is a crowd region or not.
@@ -1387,11 +1389,13 @@ def iou_with_jaccard(
     Returns:
         np.ndarray: Array of IoU values of shape (len(dt), len(gt)).
     """
-    assert len(is_crowd) == len(gt), "iou(iscrowd=) must have the same length as gt"
-    if len(dt) == 0 or len(gt) == 0:
+    assert len(is_crowd) == len(boxes_true), (
+        "iou(iscrowd=) must have the same length as boxes_true"
+    )
+    if len(boxes_detection) == 0 or len(boxes_true) == 0:
         return np.array([])
-    ious = np.zeros((len(dt), len(gt)), dtype=np.float64)
-    for g_idx, g in enumerate(gt):
-        for d_idx, d in enumerate(dt):
+    ious = np.zeros((len(boxes_detection), len(boxes_true)), dtype=np.float64)
+    for g_idx, g in enumerate(boxes_true):
+        for d_idx, d in enumerate(boxes_detection):
             ious[d_idx, g_idx] = _jaccard(d, g, is_crowd[g_idx])
     return ious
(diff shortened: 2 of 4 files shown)


supervision/metrics/mean_average_precision.py
```diff
@@ -626,3 +229,1253 @@ def plot(self):
 
         plt.tight_layout()
         plt.show()
+
+    def to_pycocotools_output(self) -> str:
```
SkalskiP, line 233: `MeanAveragePrecisionResult` currently have `__str__`. I think we can replace `__str__` implementation with this.

supervision/metrics/mean_average_precision.py
```diff
@@ -626,3 +229,1253 @@ def plot(self):
...
+        self,
+        metric_target: MetricTarget = MetricTarget.BOXES,
+        class_agnostic: bool = False,
+        class_mapping: Optional[Dict[int, int]] = None,
+        image_indices: Optional[List[int]] = None,
+    ):
+        """
+        Initialize the Mean Average Precision metric.
+
+        Args:
```
SkalskiP, line 1263: Update docstring to reflect newly added arguments.

supervision/metrics/mean_average_precision.py
```diff
@@ -626,3 +229,1253 @@ def plot(self):
...
+                prediction.class_id[:] = -1
+            for target in targets:
+                target.class_id[:] = -1
+
+        self._predictions_list.extend(predictions)
+        self._targets_list.extend(targets)
+
+        return self
+
+    def _compute_polygon_area(self, coords: List[float]) -> float:
```
SkalskiP, line 1322: Looks like this function is unused.

supervision/dataset/formats/coco.py
```diff
@@ -102,15 +106,35 @@ def coco_annotations_to_detections(
...
+        ]
+    else:
+        iscrowd = [0] * len(image_annotations)
+
+    if use_precomputed_area:
+        area = [image_annotation["area"] for image_annotation in image_annotations]
+    else:
+        area = None
+
+    if use_iscrowd or use_precomputed_area:
```
SkalskiP, line 121: Does it make sense to have two separate flags if, in the end, they are connected with `or` ?

supervision/dataset/formats/coco.py
```diff
@@ -159,16 +183,29 @@ def detections_to_coco_annotations(
     return coco_annotations, annotation_id
 
 
+def get_coco_class_index_mapping(annotations_path: str) -> Dict[int, int]:
```
SkalskiP, line 186: This function is not used.

SkalskiP (changes requested):

supervision/metrics/mean_average_precision.py
```diff
@@ -626,3 +229,1253 @@ def plot(self):
...
+            f" Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] \
+                = {self.small_objects.map50_95:.3f}\n"
+            f" Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] \
+                = {self.medium_objects.map50_95:.3f}\n"
+            f" Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] \
+                = {self.large_objects.map50_95:.3f}"
+        )
+
+
+class EvaluationDataset:
```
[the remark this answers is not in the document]
rafaelpadilla: The idea of having an intermediate format provided by the `EvaluationDataset` object is that we keep the implementation of `COCOEvaluator` as close as possible to pycocotools. 

`Detections` objects do not have properties that facilitate the search of objects and images (like `image id`) in a large dataset. The `EvaluationDataset` class facilitates the process, as it contains functions to find detections and images easily, just like in pycocotools.

In the future, if the object `Detections` is changed so that it can differentiate "predictions" from "annotations", and includes image ids, I think it would make sense to change.

The `EvaluationDataset` facilitates maintainability, comprehension, comparability with pycocotools (super important imho), and facilitates the expansion of the code to other metrics like average recall.

supervision/metrics/mean_average_precision.py
rafaelpadilla: My bad. Thanks for noticing it :smile:

supervision/metrics/mean_average_precision.py
rafaelpadilla: Removed!

supervision/dataset/formats/coco.py
rafaelpadilla: You're correct — this function is not currently invoked in the main pipeline. However, it is essential for evaluating certain models whose class ID schemes differ from those used in the COCO dataset.

Specifically, some models use sequential class IDs (e.g., 0 to 79 for 80 classes), whereas COCO's official annotations intentionally skip some IDs. You can see a detailed breakdown of these skipped IDs in [this spread sheet](https://docs.google.com/spreadsheets/d/e/2PACX-1vTTq7tjBi4RXPKS4_mCXEM4PF6i3YpRG6X1hvJjS_7OQjN4-D6DEBUg8OzGDat9fklk9iF3zdxrWDNK/pubhtml).

To address this mismatch, this function is super useful. A practical example of this mapping is used in
[this colab notebook](https://colab.research.google.com/drive/1fFjMC9_K4z4VTYC7mFlsmIR4YyV_Dhgx?usp=sharing), where `get_coco_class_index_mapping` is applied to reproduce results consistent with the [roboflow/model-leaderboard](https://github.com/roboflow/model-leaderboard)

supervision/dataset/formats/coco.py
rafaelpadilla: Thanks for bringing this up — you're absolutely right. Initially, I aimed to keep each part independent to provide more flexibility for external users. However, after reviewing it more closely, I agree that a single flag is sufficient and cleaner. I've updated the code accordingly to keep only one flag.

supervision/metrics/mean_average_precision.py
rafaelpadilla: Done!

supervision/metrics/mean_average_precision.py
```diff
@@ -626,3 +229,1253 @@ def plot(self):
...
+
+    # Calculate areas
+    Aa = max(x2a - xa, 0.0) * max(y2a - ya, 0.0)
+    Ab = max(x2b - xb, 0.0) * max(y2b - yb, 0.0)
+    Ai = max(x2i - xi, 0.0) * max(y2i - yi, 0.0)
+
+    if is_crowd:
+        return Ai / (Aa + EPS)
+
+    return Ai / (Aa + Ab - Ai + EPS)
```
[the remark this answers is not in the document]
rafaelpadilla: Done!

rafaelpadilla:
Hey @SkalskiP ,

Thank you for your review! 🙌
I've addressed all the comments and updated the code accordingly.
Let me know if everything looks good on your side.

supervision/metrics/mean_average_precision.py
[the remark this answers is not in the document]
rafaelpadilla: The main benefit of introducing an `EvaluationDataset` intermediate is alignment with `pycocotools`. It mirrors their internal structure and makes our implementation easier to understand, debug, and compare against their results — which is crucial for metric consistency and correctness.

Also, `Detections` doesn't currently expose image IDs or a clean way to separate predictions from ground truth. `EvaluationDataset` fills that gap by organizing data in a way that's much closer to what `pycocotools` expects, making it easier to implement metrics like mAP and potentially recall in the future.

If the `Detections` class evolves to include this structure natively, we could definitely revisit. But for now, this abstraction helps keep the logic cleaner, more testable, and easier to extend without bloating the evaluator.

supervision/dataset/formats/coco.py
```diff
@@ -102,15 +105,29 @@ def coco_annotations_to_detections(
...
+        iscrowd = [
+            image_annotation["iscrowd"] for image_annotation in image_annotations
+        ]
+        area = [image_annotation["area"] for image_annotation in image_annotations]
+        data = dict(
+            iscrowd=np.asarray(iscrowd, dtype=int), area=np.asarray(area, dtype=float)
+        )
+    else:
+        iscrowd = [0] * len(image_annotations)
+        area = None
```
SkalskiP, line 118: It looks like we’re not actually using `iscrowd` or `area`; we’re just returning empty dictionaries. I think it makes sense to remove both `iscrowd` and `area`.

supervision/metrics/mean_average_precision.py
```diff
@@ -626,3 +208,1144 @@ def plot(self):
...
+        # Truncate the predictions if there are more predictions than the max detections
+        # to evaluate
+        if len(dt) > self.params.max_dets[-1]:
+            dt = dt[0 : self.params.max_dets[-1]]
+
+        gt_boxes = [g["bbox"] for g in gt]
+        dt_boxes = [d["bbox"] for d in dt]
+
+        # Get the iscrowd flag for each gt
+        is_crowd = [int(o["iscrowd"]) for o in gt]
```
SkalskiP, line 624: When you load a dataset that doesn’t have `iscrowd`, this line causes the entire code to crash. We want users to be able to run evaluation on their own datasets, which might be in YOLO format or even COCO format without `iscrowd`. The code shouldn’t crash in this case.

SkalskiP:
Hi @rafaelpadilla, I just tested YOLO11 by comparing the results obtained with the updated `supervision` to those from `pycocotools`, and we’re seeing about a ~1% discrepancy.

SkalskiP:
Hi @rafaelpadilla, short update: I spent some time today diving a bit deeper. I independently benchmarked YOLO11 using both the new `supervision` and `pycocotools`, and I’m getting identical results between those two methods. However, these results differ from what our ML team report in the paper. I reached out to our ML team, to learn more.

Here are Google Colabs I prepared:

- [YOLO11 benchmark with fixed supervision](https://colab.research.google.com/drive/1ZPsO6d0R0Btqs2iApyvr-ieAeWejBHco?usp=sharing)
- [YOLO11 benchmark with pycocotools](https://colab.research.google.com/drive/13hrbP7P302l0qmHxhKxfO9T9m6A-FH8J?usp=sharing)

I'll keep you posted. For now, let's focus on other comments in this PR.

supervision/metrics/mean_average_precision.py
rafaelpadilla: You're absolutely right — thanks for pointing that out! 
:heavy_check_mark:  I've updated the evaluation logic so that if `iscrowd` is missing from the dataset, it now defaults to `iscrowd=0`. 
This ensures compatibility with datasets in YOLO format or COCO variants that don’t include the `iscrowd` field, and prevents the code from crashing in those cases.

supervision/dataset/formats/coco.py
rafaelpadilla: I agree that `use_iscrowd` might be a bit misleading when you first see it. However, the evaluation logic needs to account for the `iscrowd` property when it’s present in the dataset. This is necessary to produce results consistent with pycocotools.

Also, the `use_iscrowd` flag (which defaults to `False`) is essential for maintaining compatibility with existing tests, particularly test/dataset/formats/test_coco.py::test_coco_annotations_to_detections.

In short, the `use_iscrowd` flag serves two purposes:
* It includes the `iscrowd` property from the dataset in the Detection objects, which is critical for matching pycocotools behavior.
* It ensures backward compatibility with existing tests.

To keep things simple and closer to the original logic, I kept the `use_iscrowd` flag only in the `coco_annotations_to_detections` function. I hope that addresses your concern, but happy to adjust if you have other suggestions.

supervision/dataset/formats/coco.py
rafaelpadilla: I have updated the code to return the reversed mapping. Thanks for the suggestion!

rafaelpadilla:
Hey Piotr,

Thanks again for your review!

I've addressed all your comments and incorporated your suggestions throughout the PR.

Regarding results consistency:
* I re-ran the tests using your YOLO Colab [notebook](https://colab.research.google.com/drive/1ZPsO6d0R0Btqs2iApyvr-ieAeWejBHco?usp=sharing) and confirmed that the results remain identical :rocket: 
* I also re-tested with RT-DETR using my [notebook](https://colab.research.google.com/drive/1fFjMC9_K4z4VTYC7mFlsmIR4YyV_Dhgx?usp=sharing#scrollTo=_tL9JuQeF-Nu)  and the results are consistent as well. :rocket: 
* To make things easier to compare, I’ve consolidated both sets of results (YOLO and RT-DETR) into [this Google Sheet](https://docs.google.com/spreadsheets/d/1w8MpCgJuv6vZQfNTbmzNgitKvJwynuLpB9Ioqxkm-SI/edit?usp=sharing).

All points that could be implemented without impacting metric accuracy have been addressed. For the others, I provided context explaining the rationale behind the current implementation and pycocotools.

I believe this PR will make Supervision a simple and reliable tool for testing object detection models.

Let me know if anything else comes up.

rafaelpadilla:
Sorry, I think I accidentally closed this PR. 
Reopening it

supervision/dataset/formats/coco.py
```diff
@@ -159,16 +173,29 @@ def detections_to_coco_annotations(
...
+    )
+    inv_class_mapping = {v: k for k, v in class_mapping.items()}
+    return inv_class_mapping
+
+
 def load_coco_annotations(
     images_directory_path: str,
     annotations_path: str,
     force_masks: bool = False,
+    # use_iscrowd: bool = True,
```
SkalskiP, line 190: I noticed that in [this commit](https://github.com/roboflow/supervision/pull/1834/commits/cff974555e9d5f2245d9fa83940abcc114a84d3f#diff-64070d7a771b2a8229f1551983f12235d058b6af2fce4a834a626fd13eced843L566), you removed use_iscrowd from the public API of DetectionDataset.from_coco. I agree with this change. However, I think we should keep the use_iscrowd parameter available at the load_coco_annotations level.

supervision/detection/utils.py
```diff
@@ -1325,3 +1325,77 @@ def spread_out_boxes(
...
+    Ab = max(x2b - xb, 0.0) * max(y2b - yb, 0.0)
+    Ai = max(x2i - xi, 0.0) * max(y2i - yi, 0.0)
+
+    if is_crowd:
+        return Ai / (Aa + EPS)
+
+    return Ai / (Aa + Ab - Ai + EPS)
+
+
+def iou_with_jaccard(
```
SkalskiP, line 1371: From your explanation over email, I understand that we cannot merge this function with box_iou_batch. In that case, let’s rename it to box_iou_batch_with_jaccard and add it to supervision/__init__.py.

supervision/detection/utils.py
```diff
@@ -1325,3 +1325,77 @@ def spread_out_boxes(
...
+        boxes_true (List[List[float]]): List of ground-truth bounding boxes in the \
+            format [x, y, width, height].
+        boxes_detection (List[List[float]]): List of detection bounding boxes in the \
+            format [x, y, width, height].
+        is_crowd (List[bool]): List indicating if each ground-truth bounding box \
+            is a crowd region or not.
+
+    Returns:
+        np.ndarray: Array of IoU values of shape (len(dt), len(gt)).
+    """
```
SkalskiP, line 1391: Please include usage example at the end of docstring. You can use xyxy_to_xywh as reference.

SkalskiP (changes requested):

SkalskiP:
Hi @rafaelpadilla,
One of the features currently missing from our mAP implementation is support for max_det. The original pycocotools implementation allows you to specify the maximum number of detections per image and automatically sorts and filters them accordingly.

supervision/detection/utils.py
rafaelpadilla: :heavy_check_mark: Done!

rafaelpadilla: :heavy_check_mark: Done!

rafaelpadilla:
Hi @SkalskiP ,

I am not aware of this parameter `max_det` in the official pycocotools. Actually I have never seen the limitation of max detections by the user.  

I was looking for this usage in their examples provided by the official page [here](https://github.com/cocodataset/cocoapi/blob/master/PythonAPI/pycocoEvalDemo.ipynb) and [here](https://github.com/cocodataset/cocoapi/blob/master/PythonAPI/pycocoDemo.ipynb) , and could not find it. Where is it passed to the evaluator?

Could you please provide an example of a pycocotools call using this `max_det` parameter?

SkalskiP:
@rafaelpadilla 

> I am not aware of this parameter `max_det` in the official pycocotools. Actually I have never seen the limitation of max detections by the user.

I actually meant `maxDet`. I used snake_case out of habit, so I wrote `max_det`. Sorry for the confusion!

> If you are referring to `maxDet` in [this](https://github.com/cocodataset/cocoapi/blob/master/PythonAPI/pycocotools/cocoeval.py#L235) function, this is not an argument passed by the user. It is actually used internally for the computation of mAP. 

I genuinely thought that `maxDet` was a configurable parameter. In that case, let's keep the functional consistency with `pycocotools`. Please ignore my previous comment.

supervision/dataset/formats/coco.py
```diff
@@ -159,16 +173,29 @@ def detections_to_coco_annotations(
...
+    )
+    inv_class_mapping = {v: k for k, v in class_mapping.items()}
+    return inv_class_mapping
+
+
 def load_coco_annotations(
     images_directory_path: str,
     annotations_path: str,
     force_masks: bool = False,
+    use_iscrowd: bool = True,
```
SkalskiP, line 190: I'ts a bit weird that by default `use_iscrowd=True` in `load_coco_annotations` but `use_iscrowd=False` in `coco_annotations_to_detections`. Let's unify that.

supervision/dataset/formats/coco.py
```diff
@@ -159,16 +173,29 @@ def detections_to_coco_annotations(
     return coco_annotations, annotation_id
 
 
+def get_coco_class_index_mapping(annotations_path: str) -> Dict[int, int]:
```
SkalskiP, line 176: This function will become part of the public supervision API.

- Just like with box_iou_batch_with_jaccard, we need to IMPORT it in `supervision/__init__.py` and include it in `__all__`.

- We also need a docstring that explains the purpose of this function, the use cases where it would be useful, and how it works.

supervision/detection/utils.py
rafaelpadilla: Got your idea now! My bad 😅
I was a bit too literal with your previous comments.
:heavy_check_mark: Fixed now.

supervision/__init__.py
```diff
@@ -229,3 +233,103 @@
     "xyxy_to_xyah",
     "xyxy_to_xywh",
 ]
+
+
+def _jaccard(box_a: List[float], box_b: List[float], is_crowd: bool) -> float:
```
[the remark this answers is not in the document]
rafaelpadilla: :heavy_check_mark: Done!

supervision/dataset/formats/coco.py
```diff
@@ -90,7 +90,10 @@ def coco_annotations_to_masks(
 
 
 def coco_annotations_to_detections(
```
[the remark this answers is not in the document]
rafaelpadilla: Nice!
:heavy_check_mark:  Done!

For every existing test case in `test_coco_annotations_to_detections`, now we have test cases with `use_iscrowd=True` and `use_iscrowd=False`.

supervision/dataset/formats/coco.py
rafaelpadilla: Agree! :+1: 
:heavy_check_mark: Unified to `use_iscrowd=True`

supervision/dataset/formats/coco.py
rafaelpadilla: :heavy_check_mark: Done!

rafaelpadilla:
Hi @SkalskiP 

I believe I have addressed all the points raised during the review. 🙌
I also expanded the test cases to improve coverage with use_iscrowd=True and use_iscrowd=False.

In this last round of updates, nothing was changed that would affect the outputs. Still, as an extra precaution, I re-ran everything to confirm that the results remain consistent with previous runs ✅

The pre-commit.ci check shows a failure that is not related to this PR.

Let me know if everything looks good on your end. I believe this is now in a solid state to be merged. 🚀
Hopefully, this update will help the community achieve more consistent and reliable results with supervision.

Thanks again for the thoughtful feedback throughout the review process! 🙏 🙂

PR #1846, opened 2025-05-15
feat(detection): add transform method to remap and filter detections

AshAnand34:
# Description

This pull request introduces a new `transform` method to the `Detections` class in `supervision/detection/core.py`, allowing remapping and filtering of detections to match a target dataset's class set. It also includes corresponding unit tests to validate the functionality.

### New `transform` method in `Detections`:

* Added the `transform` method to the `Detections` class, enabling remapping of class names using an optional `class_mapping` dictionary and filtering detections to match the classes in a target dataset. The method raises an error if the required `class_name` field is missing from the `.data` attribute.

## Type of change

### Unit tests for `transform` (in `test_transform.py`)
* Added `test_transform_remap_and_filter` to verify that the method correctly remaps class names, filters out invalid detections, and updates `class_id` to match the target dataset's class indices.
* Added `test_transform_no_class_mapping` to ensure the method works correctly when no `class_mapping` is provided, retaining only the classes present in the target dataset.
* Added `test_transform_raises_without_class_name` to confirm that the method raises a `ValueError` if the `class_name` field is missing in the `.data` attribute.

### DIFF
--- a/supervision/detection/core.py
+++ b/supervision/detection/core.py
@@ -1449,15 +1449,21 @@ class Detections:
         # Get class names for each detection
         class_names = self.data.get("class_name")
         if class_names is None:
-            raise ValueError("Detections must have 'class_name' in .data to use transform().")
+            raise ValueError(
+                "Detections must have 'class_name' in .data to use transform()."
+            )
         class_names = np.array(class_names)
         # Remap class names if mapping is provided
         if class_mapping is not None:
-            class_names = np.array([class_mapping.get(name, name) for name in class_names])
+            class_names = np.array(
+                [class_mapping.get(name, name) for name in class_names]
+            )
         # Filter out detections whose class is not in dataset.classes
         keep = np.isin(class_names, dataset.classes)
         # Remap class_id to match dataset.classes
-        new_class_id = np.array([dataset.classes.index(name) for name in class_names[keep]])
+        new_class_id = np.array(
+            [dataset.classes.index(name) for name in class_names[keep]]
+        )
         # Build new Detections object
         return Detections(
             xyxy=self.xyxy[keep],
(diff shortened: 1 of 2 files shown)


supervision/detection/core.py
```diff
@@ -1435,6 +1435,55 @@ def with_nmm(
 
         return Detections.merge(result)
 
+    def transform(self, dataset, class_mapping: Optional[dict] = None) -> Detections:
```
soumik12345, line 1438: `remap_class_indices` would be a more appropriate name for this function.

supervision/detection/core.py
```diff
@@ -1435,6 +1435,55 @@ def with_nmm(
...
+                "Detections must have 'class_name' in .data to use transform()."
+            )
+        class_names = np.array(class_names)
+        # Remap class names if mapping is provided
+        if class_mapping is not None:
+            class_names = np.array(
+                [class_mapping.get(name, name) for name in class_names]
+            )
+        # Filter out detections whose class is not in dataset.classes
+        keep = np.isin(class_names, dataset.classes)
```
soumik12345, line 1464: I think its better to remove the filtering functionality from the scope of this PR since the filter can be easily applied on the remapped `sv.Detections` object. Supervision should consist of atomic functionalities that can be chained together as per the developer's requirement.

test/detection/test_transform.py
```diff
@@ -0,0 +1,55 @@
+from types import SimpleNamespace
+
+import numpy as np
+import pytest
+
+from supervision.detection.core import Detections
+
+
+def test_transform_remap_and_filter():
```
soumik12345, line 9: This testcase is failing, please make sure that tests are passing.

soumik12345 (changes requested):
Hi @AshAnand34, thanks for the PR!
I left some comments, it would be great if you address them.

AshAnand34:
> Hi @AshAnand34, thanks for the PR!
> I left some comments, it would be great if you address them.

I am out of town at the moment. I will address them when I get back.

Borda:
> I am out of town at the moment. I will address them when I get back.

sure take your time :)

PR #1853, opened 2025-05-27
Correct confusion matrix calculation-function evaluate_detection_batch

panagiotamoraiti:
# Description

This fixes the issue where predicted bounding boxes were matched to ground truth boxes solely based on IoU, without considering class agreement during the matching process. Currently, if a predicted box has a higher IoU but the wrong class, it gets matched first, and the correct prediction with the right class but lower IoU is discarded. This leads to miscounting true positives and false positives, resulting in inaccurate confusion matrix.

The change modifies the matching logic (method evaluate_detection_batch) to incorporate both IoU and class agreement simultaneously, ensuring only predictions that match both IoU threshold and class are matched to ground truths. This results in a correct confusion matrix.

## Type of change

I had an image with 2 TP and 1 FP detections, but the confusion matrix predicted 1 TP, 2 FP and 1FN. The FP bbox with the wrong class had higher overlap so the TP was discarded. At the end also this bbox was discareded due to the wrong class id. Now my confusion matrix predicts correctly  2 TP and 1 FP detections. 

I run this in a big dataset, another script i have developed and used extensively in previous project gives the following results that now match with the confusion matrix, before i corrected them they didn't match.

Test Set:
Ground Truth Objects: 481
True Positives: 469
False Positives: 11
False Negatives: 12
--------------------------------------------------
Validation Set:
Ground Truth Objects: 1073
True Positives: 1037
False Positives: 23
False Negatives: 36
--------------------------------------------------
Train Set:
Ground Truth Objects: 3716
True Positives: 3674
False Positives: 52
False Negatives: 42
--------------------------------------------------


supervision/metrics/detection.py
```diff
@@ -299,32 +299,38 @@ def evaluate_detection_batch(
...
-            matches = np.zeros((0, 3))
+        matched_gt_idx = set()
+        matched_det_idx = set()
 
-        matched_true_idx, matched_detection_idx, _ = matches.transpose().astype(
-            np.int16
-        )
+        # For each GT, find best matching detection (highest IoU > threshold)
+        for gt_idx, gt_class in enumerate(true_classes):
+            candidate_det_idxs = np.where(iou_batch[gt_idx] > iou_threshold)[0]
```
soumik12345, line 308: The selection of the best match is happening based solely on IoU, which means a wrong-class prediction can still be chosen over a right-class one if it has a higher IoU.

supervision/metrics/detection.py
```diff
@@ -299,32 +299,38 @@ def evaluate_detection_batch(
...
+            if best_det_idx not in matched_det_idx:
+                # Count as matched regardless of class:
+                # same class → TP, different class → misclassification
+                result_matrix[gt_class, det_class] += 1
+                matched_gt_idx.add(gt_idx)
+                matched_det_idx.add(best_det_idx)
             else:
-                result_matrix[true_class_value, num_classes] += 1  # FN
+                # Detection already matched, GT is FN
+                result_matrix[gt_class, num_classes] += 1
```
soumik12345, line 328: It seems that this logic iterates through ground truth boxes and for each one finds the best-matching detection box, i.e, the one with highest IoU above the threshold, that hasn't been matched yet.

The issue with this logic is that the matching process depends on the order of the ground truth boxes in `true_classes`. So, if a single detection box has a high IoU with multiple ground truth boxes, it will be matched with the first ground truth box that is processed which can lead to inconsistent and incorrect confusion matrices, as the result will vary depending on the order of ground truths in the input data.

soumik12345 (changes requested):
Hi @panagiotamoraiti, thanks for the PR!
I have a few comments regarding your logic, it would be great if you can address them and add some unit tests.

supervision/metrics/detection.py
```diff
@@ -296,36 +314,66 @@ def evaluate_detection_batch(
         true_boxes = targets[:, :class_id_idx]
         detection_boxes = detection_batch_filtered[:, :class_id_idx]
 
+        # # Debug: Print IoU calculations
+        # print("Debug IoU calculations:")
+        # print(f"GT boxes: {true_boxes}")
+        # print(f"Detection boxes: {detection_boxes}")
+
```
soumik12345, line 321: Please remove unnecessary comments.

supervision/metrics/detection.py
```diff
@@ -296,36 +314,66 @@ def evaluate_detection_batch(
...
-                    true_class_value, detection_classes[matched_detection_idx[j]]
-                ] += 1  # TP
-            else:
-                result_matrix[true_class_value, num_classes] += 1  # FN
-
-        for i, detection_class_value in enumerate(detection_classes):
-            if not any(matched_detection_idx == i):
-                result_matrix[num_classes, detection_class_value] += 1  # FP
-
+        # print(f"IoU matrix:\n{iou_batch}")
```
soumik12345, line 326: Please remove unnecessary comments.

supervision/metrics/detection.py
```diff
@@ -296,36 +314,66 @@ def evaluate_detection_batch(
...
+            for det_idx in range(len(detection_classes)):
+                iou = iou_batch[gt_idx, det_idx]
+                if iou > iou_threshold:
+                    gt_class = true_classes[gt_idx]
+                    det_class = detection_classes[det_idx]
+                    class_match = gt_class == det_class
+                    valid_matches.append((gt_idx, det_idx, iou, class_match))
+                    # print(f"Valid match: GT[{gt_idx}] class={gt_class} vs
+                    # Det[{det_idx}] class={det_class}, IoU={iou:.3f},
+                    # class_match={class_match}")
```
soumik12345, line 340: Please remove unnecessary comments.

supervision/metrics/detection.py
```diff
@@ -296,36 +314,66 @@ def evaluate_detection_batch(
...
+
+        # Sort matches by class match first (True before False), then by IoU descending
+        # This prioritizes correct class predictions over higher IoU with wrong class
+        valid_matches.sort(key=lambda x: (x[3], x[2]), reverse=True)
+        # print(f"Sorted matches: {valid_matches}")
```
soumik12345, line 345: Please remove unnecessary comments.

supervision/metrics/detection.py
```diff
@@ -296,36 +314,66 @@ def evaluate_detection_batch(
...
+        matched_gt_idx = set()
+        matched_det_idx = set()
+
+        for gt_idx, det_idx, iou, class_match in valid_matches:
+            if gt_idx not in matched_gt_idx and det_idx not in matched_det_idx:
+                # Valid spatial match - record the class prediction
+                gt_class = true_classes[gt_idx]
+                det_class = detection_classes[det_idx]
+                # print(f"Assigning match: GT[{gt_idx}] class={gt_class} ->
+                # Det[{det_idx}] class={det_class}")
```
soumik12345, line 358: Please remove unnecessary comments.

supervision/metrics/detection.py
```diff
@@ -296,36 +314,66 @@ def evaluate_detection_batch(
...
+                # This handles both correct classification (TP) and misclassification
+                result_matrix[gt_class, det_class] += 1
+                matched_gt_idx.add(gt_idx)
+                matched_det_idx.add(det_idx)
+
+        # Count unmatched ground truth as FN
+        for gt_idx, gt_class in enumerate(true_classes):
+            if gt_idx not in matched_gt_idx:
+                # print(f"Unmatched GT[{gt_idx}] class={gt_class} -> FN")
```
soumik12345, line 367: Please remove unnecessary comments.

supervision/metrics/detection.py
```diff
@@ -296,36 +314,66 @@ def evaluate_detection_batch(
...
+                result_matrix[gt_class, num_classes] += 1
+
+        # Count unmatched detections as FP
+        for det_idx, det_class in enumerate(detection_classes):
+            if det_idx not in matched_det_idx:
+                # print(f"Unmatched Det[{det_idx}] class={det_class} -> FP")
```
soumik12345, line 373: Please remove unnecessary comments.

supervision/metrics/detection.py
```diff
@@ -296,36 +314,66 @@ def evaluate_detection_batch(
...
+                result_matrix[num_classes, det_class] += 1
+
+        # print(f"Final matrix:\n{result_matrix}")
```
soumik12345, line 376: Please remove unnecessary comments.

soumik12345 (changes requested):
Hi @panagiotamoraiti, thanks for addressing the feedback.
Please remove the commented out debug statements, and add the testcases in [`test_detection.py`](https://github.com/roboflow/supervision/blob/develop/test/metrics/test_detection.py).

@SkalskiP / @onuralpszr can you please take a look as well?

galafis:
Excellent work on fixing this critical confusion matrix bug! The issue you've identified—where IoU-based matching without class agreement leads to incorrect metrics—is fundamental to proper evaluation. Your solution with class-prioritized global matching is elegant and mathematically sound.

**Technical Assessment:**

1. **Algorithmic Correctness**: The global sorting approach `(class_match, IoU)` is optimal. It ensures class-correct matches are prioritized while maintaining IoU-based ranking within each class group.

2. **Order Independence**: The move from sequential GT processing to global match collection eliminates the order dependency issue correctly identified by @soumik12345.

3. **Hungarian Assignment Alternative**: Your greedy approach is computationally efficient and appropriate for detection evaluation. While Hungarian assignment would be optimal for bipartite matching, the greedy method with proper sorting achieves the same practical results for most real-world scenarios.

**Implementation Strengths:**
- Comprehensive test cases covering edge cases and boundary conditions
- Clear separation of TP/FP/FN logic
- Maintains COCO evaluation protocol compliance
- Performance-efficient with O(n*m) complexity

**Minor Suggestions:**
- Consider adding validation for empty detection/GT cases in the main function
- The extensive test suite should definitely be integrated into the main test file as suggested

This fix will significantly improve evaluation accuracy across the computer vision community. The mathematical rigor and extensive testing demonstrate excellent software engineering practices.

Best regards,
Gabriel

panagiotamoraiti:
@soumik12345 I added the testcases into the main test file. Sorry for the delay but i had completely forgotten for some reason this suggestion. Thanks to @galafis that reminded me this.

@galafis Regarding your second suggestion (validation for empty detection/GT cases). I don't think that this is essential because if detections are empty in confusion matrix we'll have FPs, if ground truths are empty we'll have FNs and if both are empty w'll get a confusion matrix with zeros. I think no error should occur. 
I added 3 tescases addressing these 3 cases.

panagiotamoraiti:
@SkalskiP @onuralpszr @soumik12345 Any news about this PR?

panagiotamoraiti:
@SkalskiP @onuralpszr @soumik12345 Any news about this PR?

Borda:
@panagiotamoraiti could you pls adress the failing imports in tests

panagiotamoraiti:
@Borda It seems that `mock_detections` was replaced with with `_create_detections` in a newer version of supervision library. I have updated the confusion matrix test accordingly.

supervision/metrics/detection.py
```diff
@@ -285,9 +284,28 @@ def evaluate_detection_batch(
         """
         result_matrix = np.zeros((num_classes + 1, num_classes + 1))
 
+        # Filter predictions by confidence threshold
         conf_idx = 5
         confidence = predictions[:, conf_idx]
-        detection_batch_filtered = predictions[confidence > conf_threshold]
+        detection_batch_filtered = predictions[confidence >= conf_threshold]
```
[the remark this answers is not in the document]
Borda: I beilve this is valid point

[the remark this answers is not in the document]
panagiotamoraiti: I agree

Borda:
@Matvezy mind have look, pls ^^

src/supervision/metrics/detection.py
```diff
@@ -268,9 +267,28 @@ def evaluate_detection_batch(
         """
         result_matrix = np.zeros((num_classes + 1, num_classes + 1))
 
+        # Filter predictions by confidence threshold
         conf_idx = 5
         confidence = predictions[:, conf_idx]
-        detection_batch_filtered = predictions[confidence > conf_threshold]
+        detection_batch_filtered = predictions[confidence >= conf_threshold]
```
Matvezy: Why do we witch to >= for confidence threshold comparison? I believe pycocotools does just > and we were doing > too. Is there a specific reason?

Matvezy:
@panagiotamoraiti thank you for your work here! The fix and implementation make sense to me. One last ask I have is if you can add a test that runs your code with and without a fix on a small roboflow-format dataset (test split, maybe 10-100 images)? It could be an actual downloaded dataset or also just synthetic data. This is to ensure using the library on dataset's split still works properly, which is important functionality.

panagiotamoraiti:
@Matvezy Yes, i think i can add this additional test. Should i add it in a separate file inside supervision/test/metrics ?

Borda:
@Matvezy added test with Yolo dataset and validated that without the fix tests fail

PR #1857, opened 2025-06-04
added class_dist

Vishal35198:
Added a functionality to get the class distribution - 
`def get_class_distribution(dataset: "DetectionDataset") -> Dict[str, int]:
    """
    Returns a dictionary with class names as keys and sample counts as values.
    """
    from collections import Counter
    all_classes = []
    for det in dataset.annotations:
        all_classes.extend(det['class_names'])
    return dict(Counter(all_classes))
`

### DIFF
--- a/supervision/dataset/utils.py
+++ b/supervision/dataset/utils.py
@@ -48,16 +48,19 @@ def approximate_mask_with_polygons(
         approximate_polygon(polygon=polygon, percentage=approximation_percentage)
         for polygon in polygons
     ]
-    
+
+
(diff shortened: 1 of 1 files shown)


Borda:
Hi @Vishal35198 — sorry we let this sit untouched for a year. `get_class_distribution` iterates `dataset.annotations` expecting a list of dicts with a `class_names` key, but `DetectionDataset.annotations` is actually a `dict[str, Detections]` keyed by image path — so it raises an error instead of counting anything against the real API. I'm closing it since it doesn't work as written — reopen with a version that counts over `annotation.class_id`/`dataset.classes`, and we'll gladly take another look.

PR #1860, opened 2025-06-22
Include `scipy-stubs` as optional dev dependency

jorenham:
This adds [`scipy-stubs`](https://github.com/scipy/scipy-stubs) as an optional development dependency. It can help quite a bit with IDE supports, such as autocompletion and introspection. It requires no configuration, mypy plugins, and there's no runtime impact.  So even if you don't care about typing annotations, you'll basically improve DX for free :)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,8 @@ dev = [
     "ipywidgets>=8.1.1",
     "jupytext>=1.16.1",
     "nbconvert>=7.14.2",
-    "docutils!=0.21"
+    "docutils!=0.21",
+    "scipy-stubs>=1.14.1.6",
 ]
 docs = [
     "mkdocs-material[imaging]>=9.5.5",


soumik12345:
Hi @jorenham, thanks for the PR!
Can you please clarify exactly how including `scipy-stubs` as a dev dependency improves the development experience?

jorenham:
> Hi @jorenham, thank you very much for your interest in `supervision`. I reviewed our code and found only [one place](https://github.com/roboflow/supervision/blob/a6f738d2d1bfef938535110314bd12c3149cdd6a/supervision/tracker/byte_tracker/matching.py#L6) where we use `scipy` directly. This is within the `byte_tracker` package, which we plan to deprecate in the upcoming release. For this reason, we’ve decided not to merge this PR.

Makes sense; thanks for your time :)

PR #1895, opened 2025-07-17
fix: ID=0 bug in Mean Average Precision calculation

balthazur:
# Description

Objects were getting 0.0 mAP despite perfect IoU matches due to a bug in annotation ID assignment. The COCO evaluator treats ID=0 as "no match", but supervision was assigning the first annotation ID=0, causing it to be incorrectly classified as a false positive.

### Root Cause
```python
# BUGGY: Started IDs from 0
"id": len(annotations),  # First object gets ID=0 → treated as "no match"
```

### Solution
```python  
# FIXED: Start IDs from 1
"id": len(annotations) + 1,  # Start IDs from 1 (0 means no match)
```

### Impact
- **Before**: Single objects → 0.000 mAP, Multi-objects → partial failure
- **After**: All perfect predictions → 1.000 mAP

### Testing
Added 3 regression tests covering single objects, multiple objects, and batch updates to prevent future regressions.

### Reproduction Notebooks:

supervision 0.26.0 showing the bug:
https://colab.research.google.com/drive/1naVTb7qo7rbgT_AYaD8V9CvRdhK_wwPH?usp=sharing

With fix:
https://colab.research.google.com/drive/1Kz34ec1zuDbqE7lovvTWz1REKLNsyPuM?usp=sharing

Tests, and in a reproduction notebook

## Docs

### DIFF
--- a/supervision/metrics/mean_average_precision.py
+++ b/supervision/metrics/mean_average_precision.py
@@ -1237,7 +1237,7 @@ class MeanAveragePrecision(Metric):
                     "image_id": image_id,
                     "bbox": xywh,
                     "category_id": category_id,
-                    "id": len(annotations),  # incrementally increase the id
+                    "id": len(annotations) + 1,  # incrementally increase the id
                 }
                 annotations.append(dict_annotation)
         # Category list
--- a/supervision/metrics/mean_average_precision.py
+++ b/supervision/metrics/mean_average_precision.py
@@ -1237,7 +1237,7 @@ class MeanAveragePrecision(Metric):
                     "image_id": image_id,
                     "bbox": xywh,
                     "category_id": category_id,
-                    "id": len(annotations) + 1,  # incrementally increase the id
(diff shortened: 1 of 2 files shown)


rafaelpadilla:
@balthazur @SkalskiP 
Wow, great catch! thank you for spotting this! That `id=0` definitely slipped through my fingers. In larger datasets like COCO, the difference is minimal, but it has a big impact on results of small datasets. 
Starting IDs from 1 is absolutely the right fix. Nice work identifying this :clap: 

I re-run my tests comparing the results of pycocotools ([in this colab](https://colab.research.google.com/drive/1seSGLmmJvQJ4JbXkjt_nY32UuLUXayRa?usp=sharing)) and the results from this branch ([in this colab](https://colab.research.google.com/drive/1fFjMC9_K4z4VTYC7mFlsmIR4YyV_Dhgx?usp=sharing#scrollTo=2oF3_sgWFyv6)), and results still match as expected. :heavy_check_mark: 

Great fix, @balthazur :clap:

PR #1898, opened 2025-07-18
fix: handle -1 sentinel values in mAP calculation to prevent negative results and mimic pycocotools behaviour

balthazur:
# Description

This PR fixes an issue where Mean Average Precision (mAP) calculations could return negative values when certain object size categories have no data. The problem occurred because -1 sentinel values (used to indicate "no data") were being included in the mean calculation.

Notebook demonstrating the issue: https://colab.research.google.com/drive/1enb5xJ-28XLTw65GSr1ogCZl5OY7VFfF?usp=sharing

Note that this PR branches of #1895 to include the fix for always having the first ever object being a non-match. Merging that one first will slightly reduce the diff here.

### Problem

When computing mAP for specific object sizes (small/medium/large), if a category had no ground truth objects, the precision array would contain -1 values. These sentinel values were incorrectly included when calculating the mean, resulting in negative mAP scores instead of the expected -1 (no data) indicator.

This behavior didn't match pycocotools, which correctly returns:
- `0.0` for missed detections (GT exists but no predictions)
- `-1.0` for no data scenarios (no GT regardless of predictions)

Here's a notebook reproducing showing the pycocotools behaviour: https://colab.research.google.com/drive/1dyDgHDX68SMcHbCghCZTaHbrn0E8tvKL?usp=sharing

### Solution

1. **Added proper -1 sentinel value handling in `_accumulate` method**:
   - Use `np.ma.masked_equal` to exclude -1 values from mean calculations
   - Check if all values are masked (empty array) to prevent "mean of empty slice" warnings
   - Return -1 for categories with no data

2. **Fixed empty predictions handling**:
   - Added check in `load_predictions` to handle empty prediction lists without IndexError

3. **Refactored for DRY principle**:
   - Created helper function `compute_average_precision` to avoid code duplication
   - Applied consistently across all object size categories (all/small/medium/large)

## Testing

Added comprehensive test coverage for all pycocotools scenarios:
- ✅ Perfect detections → mAP = 1.0
- ✅ Missed detections (GT exists, no prediction) → mAP = 0.0
- ✅ False positives (no GT, prediction exists) → mAP = -1.0
- ✅ No data (no GT, no prediction) → mAP = -1.0
- ✅ Mixed scenarios with different object sizes
- ✅ Empty predictions and targets

All tests pass without warnings.

## Type of change

Locally, and running tests.

Also ran Model Eval with it. 

Before:
image

After:
image

Looks reasonable.

## Docs

### DIFF
--- a/supervision/metrics/mean_average_precision.py
+++ b/supervision/metrics/mean_average_precision.py
@@ -57,7 +57,11 @@ class MeanAveragePrecisionResult:
 
     @property
     def map50_95(self) -> float:
-        return self.mAP_scores.mean()
+        valid_scores = self.mAP_scores[self.mAP_scores > -1]
+        if len(valid_scores) > 0:
+            return valid_scores.mean()
+        else:
+            return -1
 
     @property
     def map50(self) -> float:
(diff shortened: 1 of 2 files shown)


soumik12345 (approved):
LGTM ✅

SkalskiP:
It looks like I don’t have access to that Colab either.

rafaelpadilla:
Hey @balthazur , 

Thank you for your detailed explanation. 
Yeah, I got confused with the branches. I just changed to  `-b model-eval-related-fixes` (or `develop` ), and now I can reproduce the results perfectly.

Totally agree on your fixes. :clap: 

Really appreciate your attention to detail! :1st_place_medal:

PR #1912, opened 2025-07-24
`edit_distance` and `fuzzy_match_index` that will allow to match faulty class name generated by VLMs with exact class names

SkalskiP:
# Description

Utils allowing to match faulty class name generated by VLMs with exact class names.

### DIFF
--- a/supervision/detection/utils/vlms.py
+++ b/supervision/detection/utils/vlms.py
@@ -1,0 +1,106 @@
+from __future__ import annotations
+
+
+def edit_distance(string_1: str, string_2: str, case_sensitive: bool = True) -> int:
+    """
+    Calculates the minimum number of single-character edits required
+    to transform one string into another. Allowed operations are insertion,
+    deletion, and substitution.
+
+    Args:
+        string_1 (str): The source string to be transformed.
+        string_2 (str): The target string to transform into.
+        case_sensitive (bool, optional): Whether comparison should be case-sensitive.
+            Defaults to True.
+
+    Returns:
+        int: The minimum number of edits required to convert `string_1`
+        into `string_2`.
+
+    Examples:
+        ```python
+        import supervision as sv
+
+        sv.edit_distance("hello", "hello")
+        # 0
+
+        sv.edit_distance("Test", "test", case_sensitive=True)
+        # 1
+
+        sv.edit_distance("abc", "xyz")
+        # 3
+
+        sv.edit_distance("hello", "")
+        # 5
+
+        sv.edit_distance("", "")
+        # 0
+
+        sv.edit_distance("hello world", "helloworld")
+        # 1
+        ```
+    """
+    if not case_sensitive:
+        string_1 = string_1.lower()
+        string_2 = string_2.lower()
+
+    length_1 = len(string_1)
+    length_2 = len(string_2)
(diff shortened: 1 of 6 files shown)


supervision/detection/utils/vlms.py
```diff
@@ -0,0 +1,106 @@
...
+                substitution_cost = 0
+            else:
+                substitution_cost = 1
+            distance_matrix[i][j] = min(
+                distance_matrix[i - 1][j] + 1,
+                distance_matrix[i][j - 1] + 1,
+                distance_matrix[i - 1][j - 1] + substitution_cost,
+            )
+
+    return distance_matrix[length_1][length_2]
```
soumik12345, line 68: Suggestion for optimization of space complexity: The current implementation uses a DP approach with space complexity `O(len(string_1) * len(string_2))`. However, the calculation of each row in the `distance_matrix` only depends on the previous row, so we don't need to store the entire matrix, instead just keep track of only the previous and current rows. This can reduce the space complexity to `O(min(len(string_1), len(string_2)))`.

supervision/metrics/mean_average_precision.py
```diff
@@ -102,12 +102,12 @@ def __str__(self) -> str:
...
-            f"Average Precision (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] \
-                = {self.medium_objects.map50_95:.3f}\n"
-            f"Average Precision (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] \
-                = {self.large_objects.map50_95:.3f}"
+            f"Average Precision (AP) @[ IoU=0.50:0.95 | area= small | "
+            f"maxDets=100 ] = {self.small_objects.map50_95:.3f}\n"
+            f"Average Precision (AP) @[ IoU=0.50:0.95 | area=medium | "
+            f"maxDets=100 ] = {self.medium_objects.map50_95:.3f}\n"
+            f"Average Precision (AP) @[ IoU=0.50:0.95 | area= large | "
+            f"maxDets=100 ] = {self.large_objects.map50_95:.3f}"
```
soumik12345, line 110: Are these changes relevant to VLM utils?

soumik12345 (dismissed):
Hi @SkalskiP, thanks for the PR!
This seems like a really useful utility for VLM-based detections!

supervision/metrics/mean_average_precision.py
SkalskiP: This relates to an issue I noticed while working on the RF-DETR fine-tuning notebook. The last two lines were misformatted.

Screenshot 2025-07-25 at 12 58 08

supervision/detection/utils/vlms.py
```diff
@@ -0,0 +1,105 @@
+from __future__ import annotations
+
+
+def edit_distance(string_1: str, string_2: str, case_sensitive: bool = True) -> int:
```
soumik12345, line 4: nit: `levenshtein_distance` would be a more suitable name for this function ([ref](https://en.wikipedia.org/wiki/Levenshtein_distance))

soumik12345 (approved):
Hi @SkalskiP, thanks for addressing the earlier comments!
I have one minor suggestion, other than that the PR is good to be merged.

PR #1919, opened 2025-07-30
FEATURE: Spline Annotations

Ashp116:
# Description

Added a spline annotation to smooth tracking detections by fitting a spline curve through the detected points. This helps reduce noise and jitter in the tracking paths, resulting in cleaner and more visually appealing traces. The spline annotation improves the quality of annotation for object tracking by producing smoother motion paths compared to raw point connections. This feature addresses issues related to noisy and jittery tracking visualization.

I will attach 2 videos showcasing spline vs tracer annotations:

RED - `SplineAnnotator`
BLUE - `TracerAnnotations`

https://github.com/user-attachments/assets/0c829567-7f37-4871-9b42-cda5c5cc1550

https://github.com/user-attachments/assets/1099c69e-8dc0-44a9-897a-7c9cfd893a9e

## Type of change

* [x] New feature (non-breaking change which adds functionality)
* [ ] This change requires a documentation update

Use this code to test out the spline tracer

```py
import supervision as sv
from ultralytics import YOLO

model = YOLO('yolo11n.pt')
spline_annotator = sv.SplineAnnotator(thickness=2, color=sv.Color.RED, smoothing_factor=20)
trace_annotator = sv.TraceAnnotator(thickness=4, color=sv.Color.BLUE)

vid_path = VideoAssets.BASKETBALL.value # Use your own video path
download_assets(VideoAssets.BASKETBALL.value)

video_info = sv.VideoInfo.from_video_path(video_path=vid_path)
frames_generator = sv.get_video_frames_generator(source_path=vid_path)
tracker = sv.ByteTrack()

with sv.VideoSink(target_path='output-annotator-1.mp4', video_info=video_info) as sink:
    for frame in frames_generator:
        result = model(frame, verbose=False)[0]
        
        # Overlay both trace and spline
            scene=frame.copy(),
            detections=detections)
        annotated_frame = spline_annotator.annotate(
            scene=annotated_frame,
            detections=detections)
```

None. This is an additive feature without breaking changes or new dependencies.

## Docs

* [ ] Docs updated? What were the changes:

### DIFF
--- a/supervision/annotators/core.py
+++ b/supervision/annotators/core.py
@@ -1937,24 +1937,33 @@ class SplineAnnotator(BaseAnnotator):
         position: Position = Position.CENTER,
         trace_length: int = 30,
         thickness: int = 2,
+        smoothing_factor: int = 20,
+        spline_order: int = 3,
         color_lookup: ColorLookup = ColorLookup.CLASS,
     ):
         """
         Args:
-            color (Union[Color, ColorPalette]): The color to draw the trace, can be
+            color (Union[Color, ColorPalette]): The color to draw the spline, can be
                 a single color or a color palette.
-            position (Position): The position of the trace.
+            position (Position): The position of the spline.
                 Defaults to `CENTER`.
-            trace_length (int): The maximum length of the trace in terms of historical
+            trace_length (int): The maximum length of the spline in terms of historical
                 points. Defaults to `30`.
-            thickness (int): The thickness of the trace lines. Defaults to `2`.
+            thickness (int): The thickness of the spline lines. Defaults to `2`.
+            smoothing_factor (int): The smoothing factor of the spline.
+            spline_order (int): The order of the spline. Use odd numbers that are 1 <= x <= 5.
             color_lookup (ColorLookup): Strategy for mapping colors to annotations.
                 Options are `INDEX`, `CLASS`, `TRACK`.
         """
         self.color: Color | ColorPalette = color
         self.trace = Trace(max_size=trace_length, anchor=position)
         self.thickness = thickness
+        self.smoothing_factor = smoothing_factor
         self.color_lookup: ColorLookup = color_lookup
+
+        if spline_order % 2 == 0 or spline_order < 1 or spline_order > 5:
+            raise ValueError("Spline order must be an odd number between 1 and 5.")
+        self.spline_order = spline_order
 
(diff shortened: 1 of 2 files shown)


SkalskiP:
Hi @Ashp116 👋🏻 Thanks for coming back with more great feature ideas! I love to see contributors coming back to add even more! 🔥

Quick question before we dive in: are you on LinkedIn or X? I wanted to tag you in the `supervision-0.26.0` release post, but I couldn't match your GitHub account to any social profiles.

Regarding your PR, this is a fantastic idea! How about we update the `TraceAnnotator` to include a `smooth` flag? Is there any reason user would like to manipulate `smoothing_factor` and `spline_order`?

Ashp116:
Hi @SkalskiP,

Thank you for the kind words!

Regarding socials, I’m on [X](https://x.com/Ashp116), but I have a small favor to ask. I'd prefer that any public post only mention my GitHub account. However, if you'd like to tag me in a LinkedIn post specifically, feel free to use my LinkedIn profile. I just like to keep things separated.

As for adding a `smooth` flag — yes, I think that's a good approach to introduce spline annotations. I’ll go ahead and make those changes. Regarding the `smoothing_factor` and `spline_order`, I don’t expect users to need to tweak those often; I added them mainly to allow for more flexible spline configurations.

Now that I think about it, adding a `smooth: int = 0` parameter would be a better approach than introducing a separate class for spline annotations. While a `smooth: bool = True` flag isn’t a bad option either, I prefer using `smooth` as an integer. It gives users finer control over the smoothing behavior, allowing them to specify the level of smoothing directly in their annotations.

supervision/annotators/core.py
```diff
@@ -1820,6 +1821,7 @@ def __init__(
         position: Position = Position.CENTER,
         trace_length: int = 30,
         thickness: int = 2,
+        smooth: int = 0,
```
SkalskiP, line 1824: I'd prepare to keep is simple and use `bool` here.

SkalskiP (changes requested):

supervision/annotators/core.py
Ashp116: I have made this change, but might I ask, won't an `int` be better as it allows for different smoothness levels for a given trace line?

SkalskiP:
This is great! I'm merging!

PR #1924, opened 2025-07-30
FEATURE: New `Video` API

### LINKED ISSUE #1923: BUG: Audio stream not captured in `process_video`
### Search before asking

- [x] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar bug report.

### Bug

The `process_video` function currently does not record or preserve the audio stream from the `source_path`. While it correctly processes and renders the video frames, it completely disregards the audio component, resulting in an output video that has no sound.

### Environment

_No response_

```py
import supervision as sv
from ultralytics import YOLO

model = YOLO('yolo11n.pt')
trace_annotator = sv.TraceAnnotator(thickness=4, color=sv.Color.BLUE)

vid_path = VideoAssets.BASKETBALL.value #"original.mp4"
download_assets(VideoAssets.BASKETBALL.value)

video_info = sv.VideoInfo.from_video_path(video_path=vid_path)
frames_generator = sv.get_video_frames_generator(source_path=vid_path)
tracker = sv.ByteTrack()

def callback(frame, i):
    result = model(frame, verbose=False)[0]
    
        scene=frame.copy(),
        detections=detections)
    
    return annotated_frame

sv.process_video(
    source_path=vid_path,
    target_path='output-annotator-1.mp4',
    callback=callback,
    show_progress=True
)
```

### Additional

_No response_

### LINKED ISSUE #1929: Reimplement video utils
### Description

The functionalities currently available in `supervision.utils.video` should be reimplemented and consolidated within a new `Video` class. Importantly, all features supported by the old video API must remain available in the new implementation. 

- get video info (works for files, RTSP, webcams)

    ```python
    import supervision as sv
     
    # static video
    sv.Video("source.mp4").info
    
    # video stream
    sv.Video("rtsp://...").info

    # webcam
    sv.Video(0).info
    ```

- simple frame iteration (object is iterable)

    ```python
    import supervision as sv
    
    video = sv.Video("source.mp4")
    for frame in video:
        ...
    ```

- advanced frame iteration (stride, sub-clip, on-the-fly resize)

    ```python
    import supervision as sv
    
    for frame in sv.Video("source.mp4").frames(stride=5, start=100, end=500, resolution_wh=(1280, 720)):
        ...
    ```

- process the video

    ```python
    import cv2
    import supervision as sv

    def blur(frame, i):
        return cv2.GaussianBlur(frame, (11, 11), 0)
    
    sv.Video("source.mp4").save(
        "blurred.mp4",
        callback=blur,
        show_progress=True
    )
    ```

- overwrite target video parameters

    ```python
    import supervision as sv
    
    sv.Video("source.mp4").save(
        "timelapse.mp4",
        fps=60,
        callback=lambda f, i: f,
        show_progress=True
    )
    ```

- complete manual control with explicit `VideoInfo`

    ```python
    from supervision import Video, VideoInfo
    
    source = Video("source.mp4")
    target_info = VideoInfo(width=800, height=800, fps=24)
    
    with src.sink("square.mp4", info=target_info) as sink:
        for f in src.frames():
            f = cv2.resize(f, target_info.resolution_wh)
            sink.write(f)
    ```

- multi-backend support decode/encode; implement PyAV and OpenCV

    ```python
    import supervision as sv
    
    video = sv.Video("source.mkv", backend="pyav")
    
    video = sv.Video("source.mkv", backend="opencv")
    ```

    suggested minimal protocol

    ```python
    class Backend(Protocol):
        def open(self, path: str) -> Any: ...
        def info(self, handle: Any) -> VideoInfo: ...
    
        def read(self, handle: Any) -> tuple[bool, np.ndarray]: ...
        def grab(self, handle: Any) -> bool: ...
        def seek(self, handle: Any, frame_idx: int) -> None: ...
    
        def writer(self, path: str, info: VideoInfo, codec: str) -> Writer: ...
    
    class Writer(Protocol):
        def write(self, frame: np.ndarray) -> None: ...
        def close(self) -> None: ...
    ``` 

### Additional

- Please share a Google Colab with minimal code to test the new feature. We know it's additional work, but it will speed up the review process. The reviewer must test each change. Setting up a local environment to do this is time-consuming. Please ensure that Google Colab can be accessed without any issues (make it public). Thank you! 🙏🏻 
- Mark all methods from the old video API as deprecated. Find examples of already deprecated methods or classes in the current codebase and use the same approach to mark the old video API methods. A deprecation period of at least 5 releases is required.
- Reimplement the internals of the old video API using the new video API.
- Provide full unit-test coverage matching existing test style.
- Update the library docs (docstrings, mkdocs) to reflect the new API.
- Take into account the comments in https://github.com/roboflow/supervision/issues/1687.

Ashp116:
# Description

This PR introduces a new `Video` API that streamlines video processing and rendering workflows. It addresses both issues #1923 and #1929 by enabling more flexible backend support and improved audio-video synchronization.

With this update, the video processing function now supports multiple backends, including `PyAV` and `OpenCV`. Notably, `PyAV` is the only backend currently supporting audio rendering, which significantly improves output quality.

This PR requires the optional dependency [`pyAV`](https://pypi.org/project/av/) for the video rendering backend. 

Tags:
Fixes #1923 
Fixes #1929

## Type of change

* [x] Bug fix (non-breaking change which fixes an issue)

Please refer to #1923 and #1929

Ensure that [`pyAV`](https://pypi.org/project/av/) is installed in the environment to test `pyAV` backend.

## Docs

* [ ] Docs updated? What were the changes


SkalskiP:
Hi @Ashp116 👋🏻 Another great idea! Video processing is probably the oldest part of `supervision`, written over two years ago, and I’ve been wanting to update its API for a while. Would you be open to not only adding audio support but also helping me with the update?

Ashp116:
Hi @SkalskiP, yea, I would like to help update the API. I was thinking of changing how videos are written in `process_video`. The original compression is lost when annotations are added and the file is written to a `target_path`. But yea, I would like to help out with the update.

SkalskiP:
Hi @Ashp116 bardzo się cieszę, że chcesz mi pomoc! Let's goooo! 🔥 🔥 🔥

I want the functionalities currently found in `supervision.utils.video` to be reorganized around a new `Video` class. Importantly, all features previously available in the old API must still be supported in the new one. Ideally, the new API should be more consistent and expressive.

- get video info (works for files, RTSP, webcams)

    ```python
    import supervision as sv
     
    # static video
    sv.Video("source.mp4").info
    
    # video stream
    sv.Video("rtsp://...").info

    # webcam
    sv.Video(0).info
    ```

- simple frame iteration (object is iterable)

    ```python
    import supervision as sv
    
    video = sv.Video("source.mp4")
    for frame in video:
        ...
    ```

- advanced frame iteration (stride, sub-clip, on-the-fly resize)

    ```python
    import supervision as sv
    
    for frame in sv.Video("source.mp4").frames(stride=5, start=100, end=500, resolution_wh=(1280, 720)):
        ...
    ```

- process the video

    ```python
    import cv2
    import supervision as sv

    def blur(frame, i):
        return cv2.GaussianBlur(frame, (11, 11), 0)
    
    sv.Video("source.mp4").save(
        "blurred.mp4",
        callback=blur,
        show_progress=True
    )
    ```

- overwrite target video parameters

    ```python
    import supervision as sv
    
    sv.Video("source.mp4").save(
        "timelapse.mp4",
        fps=60,
        callback=lambda f, i: f,
        show_progress=True
    )
    ```

- complete manual control with explicit `VideoInfo`

    ```python
    from supervision import Video, VideoInfo
    
    source = Video("source.mp4")
    target_info = VideoInfo(width=800, height=800, fps=24)
    
    with src.sink("square.mp4", info=target_info) as sink:
        for f in src.frames():
            f = cv2.resize(f, target_info.resolution_wh)
            sink.write(f)
    ```

- multi-backend support decode/encode

    ```python
    import supervision as sv
    
    video = sv.Video("source.mkv", backend="pyav")
    
    video = sv.Video("source.mkv", backend="opencv")
    ```

    suggested minimal protocol

    ```python
    class Backend(Protocol):
        def open(self, path: str) -> Any: ...
        def info(self, handle: Any) -> VideoInfo: ...
    
        def read(self, handle: Any) -> tuple[bool, np.ndarray]: ...
        def grab(self, handle: Any) -> bool: ...
        def seek(self, handle: Any, frame_idx: int) -> None: ...
    
        def writer(self, path: str, info: VideoInfo, codec: str) -> Writer: ...
    
    class Writer(Protocol):
        def write(self, frame: np.ndarray) -> None: ...
        def close(self) -> None: ...
    ```

Ashp116:
Hi @SkalskiP,

I’ve addressed most of the features you mentioned, but I have some thoughts on a few aspects of the implementation:

* **`.save` Functionality**
  How would you handle `.save` for a video feed coming from a webcam or an RTSP stream? Currently, I have it where only video files can be saved.

* **Writer and Backend Classes**
  This is just my personal opinion, but should these classes be moved to separate scripts/modules? If we add more writers and backends in the future, keeping everything inside the main video script might become cluttered.

* **“Complete manual control with explicit `VideoInfo`” Functionality**

  ```python
  from supervision import Video, VideoInfo

  source = Video("source.mp4")
  target_info = VideoInfo(width=800, height=800, fps=24)

  with src.sink("square.mp4", info=target_info) as sink:
      for f in src.frames():
          f = cv2.resize(f, target_info.resolution_wh)
          sink.write(f)
  ```

  I’m not fully clear on what this feature is intended to do. In this snippet, the `Video` instance `source` is created but never used afterward. Is `src` supposed to be `source`? Also, is the goal to create sinks for each backend? Could you please clarify the purpose and expected usage here?

pyproject.toml
```diff
@@ -46,7 +46,8 @@ dependencies = [
     "pillow>=9.4",
     "requests>=2.26.0",
     "tqdm>=4.62.3",
-    "opencv-python>=4.5.5.64"
+    "opencv-python>=4.5.5.64",
+    "imageio-ffmpeg (>=0.6.0,<0.7.0)"
```
SkalskiP, line 50: Let's not use `imageio-ffmpeg` and use `PyAV` instead.

SkalskiP: When adding `PyAV` as dependency, make sure to make it optional. We don't want to require everyone to install it.

supervision/utils/video.py
```diff
@@ -141,6 +673,7 @@ def _validate_and_setup_video(
     return video, start, end
 
 
+@DeprecationWarning
```
SkalskiP, line 676: let's use `deprecated` from `supervision.utils.internal`. it is currently used in the codebase to mark several deprecations.

supervision/utils/video.py
```diff
@@ -192,6 +725,7 @@ def get_video_frames_generator(
     video.release()
 
 
+@DeprecationWarning
```
SkalskiP, line 728: let's use `deprecated` from `supervision.utils.internal`. it is currently used in the codebase to mark several deprecations.

supervision/utils/video.py
```diff
@@ -117,6 +648,7 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
         self.__writer.release()
 
 
+@DeprecationWarning
```
SkalskiP, line 651: No need to mark private classes / functions as deprecated. We do it only for classes / functions in public supervision API.

supervision/utils/video.py
```diff
@@ -60,6 +74,523 @@ def resolution_wh(self) -> tuple[int, int]:
...
+        self.backend.save(
+            target_path=target_path,
+            callback=callback,
+            fps=fps,
+            progress_message=progress_message,
+            show_progress=show_progress,
+        )
+
+
+@DeprecationWarning
```
SkalskiP, line 593: let's use `deprecated` from `supervision.utils.internal`. it is currently used in the codebase to mark several deprecations.

supervision/utils/video.py
```diff
@@ -141,6 +673,7 @@ def _validate_and_setup_video(
     return video, start, end
 
 
+@DeprecationWarning
 def get_video_frames_generator(
```
SkalskiP, line 677: Internals of this deprecated function should be now reimplemented using new `Video` API.

supervision/utils/video.py
SkalskiP: Internals of this deprecated class should be now reimplemented using new `Video` API.

SkalskiP (changes requested):

pyproject.toml
```diff
@@ -58,6 +58,7 @@ Documentation = "https://supervision.roboflow.com/latest/"
 metrics = [
     "pandas>=2.0.0",
 ]
+video = ["av (>=15.0.0,<16.0.0)"]
```
SkalskiP, line 61: I’d rename it from `video` to `ffmpeg`. Also, do we need to set the maximum version at this point?

supervision/utils/video.py
```diff
@@ -192,6 +197,7 @@ def get_video_frames_generator(
     video.release()
 
 
+@deprecated("Use `sv.Video.save()` with a callback for processing and saving videos.")
```
SkalskiP, line 200: We aim to keep the format of all deprecation messages consistent. Please use the message below and ensure the same formatting is applied to all deprecation messages.

```
@deprecated(
    "`process_video` is deprecated and will be removed in "
    "`supervision-0.32.0`. Use `Video().save` instead."
)
```

supervision/video/__init__.py
```diff
@@ -0,0 +1,5 @@
+from supervision.video.backend.base import BaseBackend, BaseWriter
+from supervision.video.core import Video
+from supervision.video.utils import SOURCE_TYPE, VideoInfo
+
+__all__ = ["SOURCE_TYPE", "BaseBackend", "BaseWriter", "Video", "VideoInfo"]
```
SkalskiP, line 5: I don't think we need this.

supervision/video/backend/openCV.py
```diff
@@ -0,0 +1,213 @@
+from __future__ import annotations
```
SkalskiP, line 1: Let’s make all file names snake_case, `opencv.py` instead of `openCV.py`.

supervision/video/utils.py
```diff
@@ -0,0 +1,104 @@
...
+                fps = 30  # Default to 30fps if invalid
+            fps = round(fps)
+
+            total_frames = int(video.get(cv2.CAP_PROP_FRAME_COUNT))
+            if total_frames < 0:
+                total_frames = None  # Some formats may not report frame count
+        finally:
+            video.release()
+
+        return VideoInfo(width, height, fps, total_frames)
```
SkalskiP, line 94: Shouldn't this implementation be backend-specific?

SkalskiP (changes requested):

supervision/video/backend/base.py
```diff
@@ -0,0 +1,77 @@
...
+class BaseWriter(ABC):
+    @abstractmethod
+    def __init__(
+        self,
+        filename: str,
+        fps: int,
+        frame_size: tuple[int, int],
+        codec: str | None = None,
+        backend: BaseBackend | None = None,
+        render_audio: bool = False,
```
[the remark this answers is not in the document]
Ashp116: I think we do. `pyAV`'s default compression codec is `h264`, which is way better than `opencv`'s `mp4v`. So if a user wants just to render the video frames and not the audio using `pyAV`, this allows for that. Although, I think we should set `render_audio` to `None` and the default to `True`.

pyproject.toml
Ashp116: I don't think we need the maximum version. Poetry just generated it.

Ashp116:
Hi @SkalskiP,

Thank you for reviewing the PR. I have addressed all the comments from the review. Could you please take a look at the following points?

* **`render_audio` parameter:**
  You mentioned the need for this parameter. I agree it’s necessary. Here is my [response](https://github.com/roboflow/supervision/pull/1924#discussion_r2267654159):

> I think we do. `pyAV`’s default compression codec is `h264`, which produces much better quality than OpenCV’s `mp4v`. If a user wants to render only the video frames without audio using `pyAV`, this parameter allows that. I also suggest setting `render_audio` to `None` with the default as `True`.
Overall, `ffmpeg`’s default compression leads to better video outputs.

* **`.show()` feature:**
  I think this is a useful addition, and I have implemented it. However, there are some issues: currently, `cv2.imshow` is used to render the frame with a wait time of 1 ms. This causes the display to not match the correct FPS for a given video source. There are ways to address this, but they would require adding a dependency. Could you share your thoughts on this implementation approach?

* **`pyAV` webcam bug:**
  In my previous review, I completely missed webcam support for the `pyAV` backend. I’ve added this in the current PR. Using a webcam with `pyAV` requires a different code path. Could you help test this feature on other devices? I’ve verified it works on a Windows machine.

Please let me know if you have any feedback or suggestions. Thanks

ryashry (changes requested):
Improve Hight quality video web

ryashry (approved):

Ashp116:
Hello @SkalskiP, 

I been a while, I hope things are going great! I was wondering if you can review this pr at its current state so I can make changes and add-in couple of other features to this pr?

PR #1947, opened 2025-08-09
Video class refactor

timmermansjoy:
# Description

implementing feature request: #1929

pyav (optional)

## Type of change

There will be a google colabs provided when its ready for testing as requested in #1929 and there can also be extra unit tests added if needed

Documentation should be updated to give information about how the new video class will work compared to the old stand-alone functions

## Docs

still need to be done


timmermansjoy:
@SkalskiP for your issue of #1929 Is this a good start of you feel things need to be fundamentally changed?

SkalskiP:
Hi @timmermansjoy 👋🏻 Can you accept the CLA? We won't be able to merge this PR without it.

timmermansjoy:
@SkalskiP i see, just did yea

pyproject.toml
```diff
@@ -83,6 +83,9 @@ build = [
     "wheel>=0.40,<0.46",
     "build>=0.10,<1.3"
 ]
+video = [
```
SkalskiP, line 86: I would probably rename it to `ffmpeg`.

supervision/utils/video.py
```diff
@@ -60,6 +62,7 @@ def resolution_wh(self) -> tuple[int, int]:
         return self.width, self.height
 
 
+@deprecated
```
SkalskiP, line 65: I'd expect a bit more informative message.

```python
@deprecated(
    "`VideoSink` enum is deprecated and will be removed in "
    "`supervision-0.32.0`. Use `Video().sink` instead."
)
```

That would need to be applied to all `deprecated` decorators.

supervision/video/backends/base.py
```diff
@@ -0,0 +1,118 @@
+from __future__ import annotations
+
+from collections.abc import Iterator
+from typing import Protocol, runtime_checkable
+
+import numpy as np
+
+from ..utils import VideoInfo
```
SkalskiP, line 8: This import does not follow import standards. Let's update it to `from supervision.video.core import VideoInfo`.

supervision/video/backends/base.py
```diff
@@ -0,0 +1,118 @@
...
+    below.  Anything else is considered a private implementation detail.
+    """
+
+    def __init__(self, source: str | int):
+        """Create a new backend for source.
+
+        ``source`` can be
+        * ``str`` - file path, RTSP/HTTP URL …
+        * ``int`` - webcam index (OpenCV-style)
+        """
```
SkalskiP, line 25: All docstrings must follow the Google style format compatible with MKDocs. Ensure every docstring is updated to meet these guidelines.

supervision/video/backends/base.py
```diff
@@ -0,0 +1,118 @@
...
+        """Grab the next frame without decoding pixels.
+
+        Equivalent to OpenCV's ``VideoCapture.grab``.  Useful if the user only
+        wants to skip frames quickly (stride > 1 for example).
+        """
+
+    def seek(self, frame_idx: int) -> None:
+        """Seek to frame_idx so that the next :py:meth:`read` returns it."""
+
+    # Encoding ---------------------------------------------------------------
```
SkalskiP, line 46: Remove all section headers of this kind.

supervision/video/backends/opencv.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def grab(self) -> bool:
+        """Grab the next frame without decoding pixels."""
+        return self.cap.grab()
+
+    def seek(self, frame_idx: int) -> None:
+        """Seek to frame_idx so that the next :py:meth:`read` returns it."""
+        self.cap.set(cv2.CAP_PROP_POS_FRAMES, frame_idx)
+
+    # ? Do we want to mix and match different writers to different backends?
```
SkalskiP, line 74: The writer should be consistent with the backend. For example, `OpenCVBackend` should use `OpenCVWriter`, and so on.

supervision/video/utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+    width: int
+    height: int
+    fps: int
+    precise_fps: float | None = None
+    total_frames: int | None = None
+
+    @classmethod
+    def from_video_path(cls, video_path: str, backend: str | None = None) -> VideoInfo:
+        from .core import Video
```
SkalskiP, line 41: We should avoid placing imports inside methods whenever possible.

SkalskiP (changes requested):

supervision/video/utils.py
timmermansjoy: I agree, But this one is nececary to avoid circular imports since core also uses VideoInfo and we use Video here from core making it a circular import if we would import it at the top of the file

Issue #1966, opened 2025-09-22
MeanAverageRecall does not follow COCO: mAR@K should use top-K detections per image, not globally


stop1one:
### Search before asking

- [x] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar bug report.

### Describe the bug

The current implementation of [MeanAverageRecall](https://github.com/roboflow/supervision/blob/develop/supervision/metrics/mean_average_recall.py) computes mAR@K by selecting the top-K predictions across all images in the dataset, rather than selecting the top-K predictions **per image**. According to the [COCO evaluation protocol](https://cocodataset.org/#detection-eval), mAR@K should be calculated by considering the top-K highest-confidence detections **for each image**.

> **Average Recall (AR):** AR<sup>max=K</sup> AR given K detections per image

This issue occurs because, in the concatenation step below, all detection results are merged together without keeping track of which image each detection came from. As a result, the subsequent selection of top-K predictions is performed globally across the entire dataset, rather than per image.

https://github.com/roboflow/supervision/blob/deb1c9c4f4b0cd678416a67c8a13f2ef8ed6878f/supervision/metrics/mean_average_recall.py#L222-L225

### Proposed Solution

To address this issue, I have modified the `_compute` and `_compute_average_recall_for_classes`  functions so that only the top-K detections per image are considered when calculating mAR@K, in accordance with the COCO evaluation protocol.
In both functions, instead of simply concatenating all detections, I've modified the process to first filter the statistics by confidence score before concatenating them and calculating the confusion matrix. I will submit a pull request with these changes shortly.

### Environment

- Supervision 0.26.1
- OS: Ubuntu 24.04
- Python: 3.12.3

### Are you willing to submit a PR?

- [x] Yes I'd like to help by submitting a PR!

galafis:
Excellent technical analysis @stop1one! This is a critical bug that affects the validity of mAR@K evaluations for object detection models.

You're absolutely correct that the current implementation violates the COCO protocol. The key issue is indeed in the concatenation step where `np.concatenate(*stats, 0)` merges detections globally rather than maintaining per-image boundaries for top-K selection.

**Technical Impact:**
- **Bias toward high-density images**: Images with many detections dominate the global top-K selection
- **Inconsistent with COCO benchmarks**: Results won't match official COCO evaluation tools
- **Affects model comparison**: Different models may show artificially different mAR@K scores

**Validation Approach:**
[...]

PR #1967, opened 2025-09-22
Fix MeanAverageRecall: compute mAR@K using top-K detections per image (COCO-compliant)

stop1one:
# Description

This PR fixes the calculation of mAR@K in `MeanAverageRecall` to comply with the COCO evaluation protocol.  
Previously, the implementation selected the top-K predictions globally across all images, rather than per image.
According to the [COCO evaluation protocol](https://cocodataset.org/#detection-eval), mAR@K should be calculated by considering the top-K highest-confidence detections **for each image**.

This issue is tracked in issue #1966 

To resolve this, I modified the `_compute` and `_compute_average_recall_for_classes` function so that only the top-K predictions per image are considered when calculating mAR@K.

No new dependencies are required for this change.

## Type of change

I tested the change by running the metric on a dataset with varying numbers of predictions per image and verified that, for each image, only the top-K predictions (by confidence) were used in the mAR@K calculation.

No special deployment considerations are required.

## Docs

-   [ ] Docs updated? What were the changes: N/A

### DIFF
--- a/supervision/metrics/mean_average_recall.py
+++ b/supervision/metrics/mean_average_recall.py
@@ -219,9 +219,8 @@ class MeanAverageRecall(Metric):
                 large_objects=None,
             )
 
-        concatenated_stats = [np.concatenate(items, 0) for items in zip(*stats)]
         recall_scores_per_k, recall_per_class, unique_classes = (
-            self._compute_average_recall_for_classes(*concatenated_stats)
+            self._compute_average_recall_for_classes(stats)
         )
 
         return MeanAverageRecallResult(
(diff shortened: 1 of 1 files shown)


galafis:
Parabéns pela correção! A modificação para que o mAR@K seja calculado por imagem está totalmente alinhada ao protocolo COCO, melhorando a acurácia das métricas. Essa abordagem é recomendada para benchmarks robustos e facilita comparações justas entre modelos. Ótima contribuição para toda a comunidade, obrigado! Assinado: Gabriel.

stop1one:
Thank you for the encouraging feedback and detailed notes, Gabriel 🙏  
I’ve recently resumed work on this and will add the unit test you suggested (per-image K-filtering with synthetic data) very soon.

stop1one:
I've added a simple unit test with synthetic data to validate the mAR@K calculation.

**Test setup:**
- **15 images** in total.
- Each image has **≤ bounding boxes/detections**.
- Therefore, **mAR@10 and mAR@100 should be identical**, since $K=10$ is already exceeds the maximum number of detections per image.

**Result with the original (buggy) implementation:**
```
E           AssertionError: 
E           Arrays are not almost equal to 5 decimals
E           
E           Mismatched elements: 2 / 3 (66.7%)
E           Max absolute difference among violations: 0.23173375
E           Max relative difference among violations: 0.80613893
E            ACTUAL: array([0.05573, 0.52786, 0.63622])
E            DESIRED: array([0.28746, 0.63622, 0.63622])
```

As shown above, **mAR@10 (0.52786) ≠ mAR@100 (0.63622)**, which is incorrect.  
This demonstrates that the original code was applying Top-K filtering **across the dataset**, rather than **per image**.  
The fix in this PR corrects that behaviour.

stop1one:
Hi @SkalskiP , 
I just realized I hadn’t signed the CLA earlier — that’s done now ✅  
Would you mind taking another look at this PR when you have a moment?  
Thanks again for your time!

Borda:
@Matvezy mind have look at this one ^^ 🦩

supervision/metrics/mean_average_recall.py
```diff
@@ -238,25 +237,34 @@ def _compute(
 
     def _compute_average_recall_for_classes(
         self,
-        matches: np.ndarray,
```
Borda: @stop1one, could you pls have a look at the collisions and ideally do not change the API here?

stop1one: Hi @Borda , It's been about a few months since I opened this PR, so there are quite a few conflicts to resolve. I'll rebase it and see if I can revert the API changes as you suggested.

stop1one: https://github.com/roboflow/supervision/blob/5ce6e484b8d47ad5ba1329bc838826185f9940ec/src/supervision/metrics/mean_average_recall.py#L428-L431

I've looked into the collisions and resolving it is straightforward.
However, the current concatenation process merges detection results from each image into a single array withou any boundaries, which is the root cause of the incorrect mAR@K calculation.
To fix this, **updating the function signiture is inevitable**.

I see two ways to proceed:
1. Minimize API changes at the cost of adding redundant and less efficient code.
2. Refactor the structure for better efficiency and cleaner code, even if it requires a more significant change to the signature.

Which direction would you prefer?

stop1one:
Moving this to #2136

Issue #1987, opened 2025-10-09
Polygon Zone Superposition - Clipping bug


zaknye:
### Search before asking

- [x] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar bug report.

### Bug

I've identified a bug in the `PolygonZone` class that allows a single object to occupy two ROIs even if they have no overlap. It seems the bug was introduced here:
https://github.com/roboflow/supervision/issues/1101#issuecomment-2051863063

By moving the clipping calculation from using the frame height/width to the ROI height/width you're creating two separate bounding boxes for each separate ROI. This clips the boxes to exist within the bounds of the boolean numpy array (good) but then calculates the anchor point AFTER modifying the box (bad). This allows a single object to have two different anchor points for different ROIs causing the object to exists within two zones at once.

I propose that instead of clipping the incoming BBox to within the ROI then calculating anchor points do the following:
1. Calculate the center of the box
2. Check if the center point exceeds the min/max of the shape of the `self.mask` array. If so return False as the BBox can not possible occupy the ROI
3. Finally check the point against the `self.mask` array and return the result. 

### Environment

- Supervision: 0.26.1
- Python 3.12.3  

### Minimal Reproducible Example

rois:
roi1 = np.array([[0,167],[422,79],[457,427],[102,547]], np.int32)
roi2 = np.array([[458,67],[1011,75],[997,309],[515,609]], np.int32)
roi3 = np.array([[0,683],[457,474],[474,639],[219,720]], np.int32)

detection:
Detections(xyxy=array([[441.50900269, 258.48202515, 718.4284668 , 712.74450684]]), mask=None, confidence=array([0.89746094]), class_id=array([0]), tracker_id=array([3]), data={'class_name': array(['person'], dtype='<U6')}, metadata={})

### Are you willing to submit a PR?

- [x] Yes I'd like to help by submitting a PR!

PR #1992, opened 2025-10-14
Fix/polygon zone roi crop

zaknye:
# Description

Fix Polygon zone calculation so that boxes can not exist in two mutually exclusive zones. See https://github.com/roboflow/supervision/issues/1987

- None

## Type of change

Added a test case that introduces a single box to two polygon zones and tests to see if it is an occupant in either. With correct operation using the center point the box should only reside in one and not both.

Added some document changes to clear up using the polygon zone

## Docs

Added some document changes to clear up using the polygon zone
Added a change to CONTRIBUTING.md as one of the examples of using `uv` seemed to be out of date

### DIFF
--- a/src/supervision/detection/tools/polygon_zone.py
+++ b/src/supervision/detection/tools/polygon_zone.py
@@ -33,8 +33,6 @@ class PolygonZone:
             in the zone to be considered occupants.
             (default: (sv.Position.BOTTOM_CENTER,)).
         current_count (int): The current count of detected objects within the zone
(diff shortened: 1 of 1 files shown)


Borda:
Hi @zaknye — apologies for not closing this out sooner. The underlying issue (#1987) was fixed in #1991, which was merged to `develop` on 2026-03-10. That PR takes a similar approach — calculating the anchor point before clipping — so this PR is superseded. Thank you for the contribution and for writing the test case that helped clarify the problem. Closing as superseded — please reopen if you see remaining edge cases on the next release.

PR #1997, opened 2025-10-26
feat: add threaded I/O pipeline for video processing

AnonymDevOSS:
Implements pipeline with bounded queues to overlap decode, compute and encode. Reduces I/O stalls.

# Description

Process a video using a threaded pipeline that asynchronously
reads frames, applies a callback to each, and writes the results
to an output file.

This function implements a three-stage pipeline designed to maximize
frame throughput.

- Reader thread: reads frames from disk into a bounded queue ('read_q')
  until full, then blocks. This ensures we never load more than 'prefetch'
  frames into memory at once.

- Main thread: dequeues frames, applies the 'callback(frame, idx)',
  and enqueues the processed result into 'write_q'.
  This is the compute stage. It's important to note that it's not threaded,
  so you can safely use any detectors, trackers, or other stateful objects
  without synchronization issues.

- Writer thread: dequeues frames and writes them to disk.

Both queues are bounded to enforce back-pressure:
  - The reader cannot outpace processing
  - The processor cannot outpace writing

Summary:
- It's thread-safe: because the callback runs only in the main thread,
using a single stateful detector/tracker inside callback does not require
synchronization with the reader/writer threads.

- While the main thread processes frame N, the reader is already decoding frame N+1,
  and the writer is encoding frame N-1. They operate concurrently without blocking
  each other.

## Type of change

-   [x ] New feature (non-breaking change which adds functionality)
-   [x ] This change requires a documentation update

I created a benchmark script to measure the performance impact of these changes. ([benchmark_process_video.py](https://github.com/user-attachments/files/23153706/benchmark_process_video.py); 
[full_results.txt](https://github.com/user-attachments/files/23153708/full_results.txt))

I created 3 functions to benchmark: opencv (short), opencv (long), tracker; benchmarking current process_video and new process_video_threads.

Results below, 5 executions for each case:
```
OpenCV (short)
process_video_threads       avg=  2.338s  stdev= 0.200s  min= 2.137s  max= 2.620s
process_video               avg=  3.560s  stdev= 0.404s  min= 3.219s  max= 4.249s

OpenCV (long) 
process_video_threads       avg= 18.449s  stdev= 1.426s  min=17.067s  max=20.863s
process_video               avg= 28.067s  stdev= 1.345s  min=26.261s  max=29.373s

Tracker.
process_video_threads       avg= 21.481s  stdev= 0.593s  min=20.825s  max=22.205s
process_video               avg= 24.929s  stdev= 0.368s  min=24.464s  max=25.307s
```

Initially, I explored using threads and processes to parallelize process_video (I can push some of those prototypes if needed), but this design wasn’t thread-safe for stateful callbacks (e.g. trackers) and showed little improvement in profiling; most of the total time was spent on disk I/O rather than computation.

This optimization instead focuses on improving the I/O path, yielding a more generic and safe performance gain.

### DIFF
--- a/supervision/detection/utils/boxes.py
+++ b/supervision/detection/utils/boxes.py
@@ -14,8 +14,8 @@ def box_aspect_ratio(xyxy: np.ndarray) -> np.ndarray:
     for boxes with zero height to avoid division errors.
 
     Args:
-        xyxy (`numpy.ndarray`): Array of bounding boxes in `(x_min, y_min, x_max, y_max)`
-            format with shape `(N, 4)`.
+        xyxy (`numpy.ndarray`): Array of bounding boxes in
+            `(x_min, y_min, x_max, y_max)` format with shape `(N, 4)`.
 
     Returns:
         `numpy.ndarray`: Array of aspect ratios with shape `(N,)`, where each element is


AnonymDevOSS:
I pushed it now. Sorry, I read your message a bit too late.

SkalskiP (approved):

SkalskiP:
Hi @AnonymDevOSS 👋🏻 I applied a few cosmetic updates to the docstring and variable names. Merging now.

PR #1998, opened 2025-10-26
fix: process video bug on unused else

AnonymDevOSS:
The loop never uses break, so the else clause always executes. However, since it iterates over a generator, the generator is already exhausted after the first pass... so the else block would iterate over nothing.

So, i believe the 'else' is redundant.

## Type of change

```
>>> def g():
...   for x in range(3):
...     yield x
... 
>>> gen = g()
>>> for x in gen:
...   print(x)
... else:
...   for y in gen:
...     print("y", y)
... 
0
1
2

```

## Docs

### DIFF
--- a/supervision/utils/video.py
+++ b/supervision/utils/video.py
@@ -249,10 +249,6 @@ def process_video(
         ):
             result_frame = callback(frame, index)
             sink.write_frame(frame=result_frame)
-        else:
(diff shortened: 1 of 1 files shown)


Borda:
@AnonymDevOSS, thank you for your effort and for helping to resolve this issue. Regarding the progress and still some non-trivial path to be finished, we will proseed with #2022

PR #2001, opened 2025-10-27
feat: speed up box iou batch using 2d in-place ops instead of 3d

AnonymDevOSS:
# Description

I focused on improving the performance of the IoU calculation between bounding boxes.

This change adds a new function box_iou_batch_alt, which performs the same computation as box_iou_batch but runs faster and uses less memory.

The optimization uses in-place NumPy operations and avoids creating temporary arrays.
Results are identical to the original implementation within a small numerical tolerance.

No new dependencies were added.

## Type of change

 Performance improvement (non-breaking change)

I compared the outputs of both functions (box_iou_batch and box_iou_batch_alt) using random test boxes.
A simple pytest test checks that the results are equal within rtol=1e-6 and atol=1e-6.
Both functions return the same values for IoU and IoS modes.
Benchmarks show that the new version is around 2–5× faster on typical box sizes.

I also add a benchmark.

```
(.venv) wednesday:benchmark rafelbennasar$ python benchmark_iou.py 
Benchmarking 2 functions on sizes=[(512, 1024)], 5 runs × 5 iterations each

- box_iou_batch  (512×1024)
[benchmark_iou.py](https://github.com/user-attachments/files/23167441/benchmark_iou.py)

  Run 01:   0.241 s
  Run 02:   0.232 s
  Run 03:   0.522 s
  Run 04:   0.246 s
  Run 05:   0.207 s
  Average:   0.290 s

- box_iou_batch_alt  (512×1024)
  Run 01:   0.060 s
  Run 02:   0.064 s
  Run 03:   0.066 s
  Run 04:   0.065 s
  Run 05:   0.066 s
  Average:   0.064 s


= Summary (5 runs × 5 iterations each):
box_iou_batch_alt           avg=  0.064s  stdev= 0.002s  min= 0.060s  max= 0.066s
box_iou_batch               avg=  0.290s  stdev= 0.131s  min= 0.207s  max= 0.522s


```

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -68,10 +68,10 @@ def random_boxes(
 
     Args:
         count (`int`): Number of random bounding boxes to generate.
-        image_size (`tuple[int, int]`): Image size as `(width, height)`. Defaults to `(1920, 1080)`.
-        min_box_size (`int`): Minimum side length (pixels) for generated boxes. Defaults to `20`.
-        max_box_size (`int`): Maximum side length (pixels) for generated boxes. Defaults to `200`.
-        seed (`int` or `None`): Optional random seed for reproducibility. Defaults to `None`.
+        image_size (`tuple[int, int]`): Image size as `(width, height)`.
+        min_box_size (`int`): Minimum side length (pixels) for generated boxes.
+        max_box_size (`int`): Maximum side length (pixels) for generated boxes.
+        seed (`int` or `None`): Optional random seed for reproducibility.
 
     Returns:
         (`numpy.ndarray`): Array of shape `(count, 4)` with bounding boxes as


SkalskiP:
Hi @AnonymDevOSS 👋🏻 thanks a lot for the contribution!

Please accept the CLA license so we can move forward.

SkalskiP:
Looks like you used different account to commit the code. 

Screenshot 2025-10-28 at 13 05 08

supervision/detection/utils/iou_and_nms.py
```diff
@@ -231,6 +231,96 @@ def box_area(box):
     return ious
 
 
+def box_iou_batch_alt(
```
SkalskiP: Let's rename it to `box_iou_batch` and remove old implementation of this function.

test/detection/utils/functions.py
```diff
@@ -0,0 +1,38 @@
+import random
+
+import numpy as np
+
+
+def generate_boxes(
```
SkalskiP: In `supervision/test/test_utils.py`, there are already two functions, `mock_detections` and `mock_key_points`. It seems we are duplicating some logic from `mock_detections`. Try to unify them. 

If that’s not possible, at least move `generate_boxes` to `supervision/test/test_utils.py` and make its arguments and naming conventions consistent with the existing functions. 

One quick improvement would be to replace the separate `W` and `H` arguments with a single `resolution_wh` argument, which we use throughout the codebase.

test/detection/utils/test_iou_and_nms.py
```diff
@@ -631,3 +634,14 @@ def test_mask_non_max_merge(
         sorted_result = sorted([sorted(group) for group in result])
         sorted_expected_result = sorted([sorted(group) for group in expected_result])
         assert sorted_result == sorted_expected_result
+
+
+def test_box_iou_batch_and_alt_equivalence():
```
SkalskiP: As we will rename `box_iou_batch_alt` to `box_iou_batch` that test no longer will be necessary.

AnonymDevOSS: Added additional test cases rather than using the trivial approach of comparing against the original implementation.

SkalskiP (approved):

SkalskiP:
Hi @AnonymDevOSS! I pushed a few small improvements today. I updated the docstring and example, unified the variable names, and cleaned up the unit test structure. Since we modified `box_iou_batch`, I updated `box_iou` as well. The code is tested and ready to merge. I also added my own speed benchmark.

SkalskiP (approved):

PR #2059, opened 2026-01-07
examples/traffic_analysis: improve CLI argument handling

Borda:
This pull request refactors the command-line interface (CLI) handling in both `inference_example.py` and `ultralytics_example.py` for the traffic analysis examples. The main improvement is replacing the previous `argparse`-based CLI with a more flexible approach that prefers `jsonargparse` if available, while providing a fallback to positional arguments for environments where `jsonargparse` is not installed. Additionally, the main execution logic is moved into a dedicated `main` function in each script, improving modularity and readability.

**CLI refactoring and flexibility:**

* Replaced the `argparse`-based CLI with logic that first attempts to use `jsonargparse` for argument parsing, falling back to simple positional arguments if `jsonargparse` is unavailable. This makes the scripts more robust and user-friendly in different environments. [[1]](diffhunk://#diff-30cfa50159fbe874da6f1bcefc707172f587555b7c5fefddc9e2c997bb4fc6a8L183-R247) [[2]](diffhunk://#diff-4aa62614bcc352987ce79b79d88d5b6ca290a0e451308e2cff351353619bded7L180-R229)
* Moved the script logic into a dedicated `main` function in both `inference_example.py` and `ultralytics_example.py`, improving code organization and enabling easier reuse and testing. [[1]](diffhunk://#diff-30cfa50159fbe874da6f1bcefc707172f587555b7c5fefddc9e2c997bb4fc6a8L183-R247) [[2]](diffhunk://#diff-4aa62614bcc352987ce79b79d88d5b6ca290a0e451308e2cff351353619bded7L180-R229)

**Imports and dependencies:**

* Updated imports to remove `argparse` and add `sys`, reflecting the new CLI handling approach. [[1]](diffhunk://#diff-30cfa50159fbe874da6f1bcefc707172f587555b7c5fefddc9e2c997bb4fc6a8L3-R4) [[2]](diffhunk://#diff-4aa62614bcc352987ce79b79d88d5b6ca290a0e451308e2cff351353619bded7L3-R3)

---

Part of #2090

PR #2055, opened 2026-01-07
examples/heatmap_and_track: improve CLI argument parsing

Borda:
This pull request refactors the CLI interface and entrypoint logic for the heatmap and tracking script, making it more robust and flexible. The main function is renamed and documented, and the CLI argument parsing now supports both `jsonargparse` and a fallback to positional arguments if the package is unavailable. This improves usability and maintainability of the script.

**CLI and Entrypoint Refactor:**

* Replaced `argparse` with a conditional import of `jsonargparse` for CLI parsing, with fallback to positional arguments if `jsonargparse` is unavailable. This makes the script more flexible for different environments. [[1]](diffhunk://#diff-33e1ef7081c39bcfb5e8ae015880aa18a46dc8fdd60744beb8b125ffd2533712L1-R1) [[2]](diffhunk://#diff-33e1ef7081c39bcfb5e8ae015880aa18a46dc8fdd60744beb8b125ffd2533712L104-R145)
* Renamed the main function from `heatmap_and_track` to `main` and added a comprehensive docstring describing all parameters for better clarity and maintainability. [[1]](diffhunk://#diff-33e1ef7081c39bcfb5e8ae015880aa18a46dc8fdd60744beb8b125ffd2533712L15-R15) [[2]](diffhunk://#diff-33e1ef7081c39bcfb5e8ae015880aa18a46dc8fdd60744beb8b125ffd2533712R27-R41)
* Updated the script's entrypoint to call `main` directly with parsed arguments, removing the previous manual mapping and improving code readability.

---

Part of #2090

PR #2068, opened 2026-01-11
Add jsonargparse dependency for CLI

Borda:
This pull request adds a new optional dependency group for CLI utilities in the `pyproject.toml` file. This group includes the `jsonargparse` package, which will be used for examples and other command-line interface utilities.

Dependency management:

* Added a `cli` dependency group with `jsonargparse[signatures]>=4.38.0` to support CLI utilities and examples in `pyproject.toml`.

Will impact: #2059, #2058, #2057, #2056, #2055, #2054

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,6 +58,10 @@ Documentation = "https://supervision.roboflow.com/latest/"
 metrics = [
     "pandas>=2.0.0",
 ]
+cli = [
+    # to be used for examples and other CLI utilities
+    "jsonargparse[signatures]>=4.38.0",
+]
 
 [dependency-groups]
 dev = [


pyproject.toml
```diff
@@ -58,6 +58,10 @@ Documentation = "https://supervision.roboflow.com/latest/"
 metrics = [
     "pandas>=2.0.0",
 ]
+cli = [
```
SkalskiP, line 61: why not just throw it into `dev` ?

Borda: Because dev is not extra but just a group, so not distributed with the package

Borda: > The `[dependency-groups]` table is a relatively new standard (defined in PEP 735) designed to provide a unified way to store non-runtime dependencies.
> 
> As of right now, plain pip cannot read `[dependency-groups]` directly. If you try to run `pip install .`, it will only install the base dependencies. If you try `pip install ".[dev]"`, it will look for `[project.optional-dependencies]`, not the new `[dependency-groups]`.

Borda:
Let's decide later if we extend supervison with CLI

PR #2022, opened 2025-11-27
fix: `process_video` silent hang on callback error

realh4m:
# Description

Added exception tracking in `process_video` so callback errors raised in the worker loop are surfaced after reader/writer threads finish, preventing silent hangs when the callback crashes mid-stream.

Introduced `read_finished` to avoid draining the read queue unnecessarily once EOF is reached.

- **Issue:** Fixes the silent failure where `process_video` could hang without any error when the callback raised an exception.
- **Dependencies:** None.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

## How has this change been tested, please provide a testcase or example of how you tested the change?

Verified locally with a callback that raises an exception.

## Any specific deployment considerations

No additional considerations; behavior change only affects `process_video` error handling.

## Docs

- [ ] Docs updated? What were the changes:

PR #2102, opened 2026-01-23
fix(process_video): deadlock on callback exception + improve shutdown robustness

SkalskiP:
## Description

The updated `process_video` function implements a three-stage threaded pipeline consisting of a reader, a main-thread processor, and a writer, while using bounded queues to strictly control memory consumption during video processing.

When the user-supplied callback raises an exception, the function logs the error along with the affected frame index and, by default, re-raises it after performing a clean shutdown of all components.

With the optional `skip_on_error=True` parameter, problematic frames are silently discarded instead of interrupting the entire job, which is especially useful for noisy or partially corrupted videos.

The reader thread runs as non-daemon so that `cv2.VideoCapture.release()` is reliably called even during early termination, preventing file handle or resource leaks.

Shutdown is managed through a combination of a `stop_event`, non-blocking queue operations with small timeouts, and a light sleep-based backoff that avoids both deadlocks and excessive CPU usage.

The writer thread is kept as a daemon because disk writes can usually be safely interrupted, while both threads receive sensible join timeouts with logging warnings if they fail to complete in time.

## Type of Change

🐛 Bug fix (non-breaking change which fixes an issue)

## Motivation and Context

This change fixes a critical bug in the `process_video` function where an unhandled exception thrown inside the user-provided `callback` would cause the entire processing pipeline to deadlock indefinitely.  

The main thread would exit its processing loop but then block forever on `reader_worker.join()` while the reader thread (a daemon) was stuck trying to `put()` into a full `frame_read_queue`.

## Testing

- [ ] I have added unit tests that prove my fix is effective or that my feature works

### DIFF
--- a/supervision/utils/video.py
+++ b/supervision/utils/video.py
@@ -206,27 +206,33 @@ def process_video(
     progress_message: str = "Processing video",
     skip_on_error: bool = False,
 ) -> None:
-    """Process video frames using a three-stage threaded pipeline with controlled memory usage.
+    """Process video frames using a three-stage threaded pipeline with controlled
+    memory usage.
 
-    Reads frames in a background thread, processes them via user callback in the main thread,
-    and writes results in another background thread. Uses bounded queues to limit memory.
+    Reads frames in a background thread, processes them via user callback in the main
+    thread, and writes results in another background thread. Uses bounded queues to
+    limit memory.
 
     Args:
         source_path: Path to the input video file.
         target_path: Path where the processed video will be saved.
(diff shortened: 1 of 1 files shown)


Borda:
FYI, the core deadlock this PR set out to fix has already shipped via #2022 (merged 2026-01-30) and was hardened further since: `develop` now drains the read queue after a callback exception and joins both threads with a timeout instead of blocking indefinitely. Closing this as resolved by that work.

The `skip_on_error` idea is still valuable and isn't covered by #2022 — worth a focused follow-up PR against current `develop` rather than rebasing this one, since the threading internals have moved on since January. Reopen if you'd rather land that here instead.

Issue #2108, opened 2026-01-27
[Improvement] Increase Unit Test Coverage to ~90%


Borda:
### 📝 Description

As `supervision` matures into an industry-standard utility package for Computer Vision, the reliability and stability of the codebase become paramount. Currently, our total test coverage sits at approximately **50%**, with critical modules like `metrics` and `key_points` falling well below 40%.

To ensure this package remains the go-to solution for production pipelines and research alike, I propose an initiative to systematically increase our unit test coverage to a realistic target of **~90%**.

### 🚀 Motivation

Why is this high priority?

* **Industry Standard Responsibility:** `supervision` is infrastructure. Users rely on it for critical CV workflows. A subtle bug in a metric calculation (like mAP or confusion matrix) can silently invalidate research experiments.
* **Refactoring Confidence:** As we optimize performance, a robust test suite acts as a safety net.
* **Handling Edge Cases:** Computer Vision data is notoriously messy. High coverage ensures we handle empty detections, `None` values, and malformed shapes gracefully.

### 📊 Current State

According to the latest Codecov report, we have significant gaps in core modules.

| Module | Coverage | Status | Notes |
| :--- | :--- | :--- | :--- |
| **`metrics`** | **38.36%** | 🔴 Critical | Over 1,100 lines untested. |
| **`key_points`** | **36.68%** | 🔴 Critical | Lowest percentage in the library. |
| **`utils`** | **55.06%** | 🟡 Low | |
| **`dataset`** | **57.32%** | 🟡 Low | Formats logic needs hardening. |
| **`detection`** | **68.49%** | 🟡 Moderate | High volume of missed lines (639) due to file size. |

### 🛠 Proposed Strategy

We should tackle this modularly. The immediate priority is to lift the "Red" modules out of the danger zone.

**Priority 1: The Criticals**
* **`metrics/`**: This is the highest impact area. We need to verify that mathematical calculations for accuracy, recall, and precision are correct across various edge cases.
* **`key_points/`**: Needs basic functional tests for skeleton handling and visualization.

**Priority 2: Stability Layers**
* **`dataset/`**: Ensure saving/loading functions (YOLO, COCO, Pascal VOC) are robust against file system errors and malformed inputs.
* **`utils/`**: General utility functions that are used everywhere need to be bulletproof.

### ✅ Guidelines for New Tests

For contributors looking to help:
* **Test Behavior, Not Implementation:** Focus on inputs and expected outputs.
[...]

PR #2119, opened 2026-01-31
add unit tests for precision, recall and f1_score metrics

dcpagotto:
Hey! This PR adds comprehensive unit tests for the precision, recall and f1_score metrics that were missing test coverage.

Previously these 3 metrics had zero tests while mean_average_precision was the only one with tests. This was mentioned in issue #2108 which asked for increasing test coverage to ~90%.

What I tested:
- Instance creation with default and custom parameters
- Reset functionality to clear stored data 
- Perfect matches, no overlaps, empty cases
- Single class and multiple class scenarios  
- Different IoU thresholds (0.5 vs 0.75)
- All 3 averaging methods (macro, micro, weighted)
- Confidence ranking and edge cases
- List inputs and error handling

All test values were calculated manually to make sure the math is correct. Added comments explaining the calculations so its easy to verify.

The tests follow the same style as existing tests in the repo and use pytest with parametrize for testing different configurations.

This should help get closer to the 90% coverage target mentioned in the issue.


test/metrics/test_f1_score.py
```diff
@@ -0,0 +1,511 @@
...
+    predictions = Detections(
+        xyxy=np.array([[10, 10, 50, 50]], dtype=np.float32),
+        confidence=np.array([0.9]),
+        class_id=np.array([0]),
+    )
+
+    targets = Detections(
+        xyxy=np.empty((0, 4), dtype=np.float32),
+        class_id=np.empty((0,), dtype=int),
+    )
```
Borda: seems you are using the same, so lets wrap then to a class and use fixture to init these two inputs to umprove redability

Borda (approved):

PR #2120, opened 2026-01-31
refactor(typing): Typing error has been fixed for draw and geometry files

Kadermiyanyedi:
Fix typing issues in the following files:
* supervision/draw/color.py
* supervision/draw/utils.py
* supervision/geometry/core.py

This change only addresses mypy/type-checking errors and does not modify functionality.


Kadermiyanyedi:
Rect fields (x, y, width, height) are defined as float, but are used as pixel indices in draw/utils.py file. They are explicitly cast to int here. It may be worth updating Rect (or adding a pixel-based variant) to better reflect this usage. @Borda

Borda (approved):

Issue #2088, opened 2026-01-21
[Refactor]: Improve typing and sanitize docs


Borda:
### Search before asking

- [x] I have searched the issues and discussions and found no similar feature request.

### Feature Description

I suggest we remove them from the docstrings and rely solely on the signature. Here is the breakdown of why the Python ecosystem has shifted this way:

### 1. The Modern Approach (Signature as Truth)
Since PEP 484, the standard is to treat the function signature as the **Single Source of Truth**.
Tools like **mypy**, **Pyright**, and our IDEs rely entirely on the signature to validate our code. By keeping the focus there, we ensure that the types we document are the actual types being enforced by the interpreter.

### 2. The Legacy Approach (Why to avoid duplication)
Writing types in the docstring was necessary before Python 3.5, but today it is considered a violation of **DRY (Don't Repeat Yourself)**.
* **Maintenance Burden:** It creates a "split truth." If we change a variable from `int` to `float` in the code but forget to update the English text in the docstring, our documentation becomes factually incorrect.
* **Style Standards:** The **Google Python Style Guide** now explicitly states: *"The description should include the type if and only if the function does not have a type annotation."*

### 3. Do we need both for the docs?
No. You might be worried that removing them will break our generated documentation, but tools like **Sphinx** (via `sphinx-autodoc-typehints`) or **MkDocs** now automatically pull the types from the signature and insert them into the generated HTML. We get the same result without the manual effort.

### Are you willing to submit a PR?

- [x] Yes I'd like to help by submitting a PR!

Borda:
This is multi PR effort, so some other contributors are welcome, lest just sync our work by choosing a sub-package which is not addressed yet and linking this issue to keep track of progress... 🦩

Borda:
and setting the validation in https://github.com/roboflow/supervision/pull/2091 so whenever we fix some we prune the ignore list so we ensure that we won't regress...

Borda:
let's consider switching `mypy` to `ty` when `ty` has an official hook - https://github.com/astral-sh/ty/issues/269

Kadermiyanyedi:
Hi! I’d be happy to help with this issue.
If needed, I can handle the convert mypy to ty.

Alternatively, we could break this down into a small TODO list and people can work through the tasks step by step.
Let me know what would be most useful or where you’d like me to start.

Borda:
@Kadermiyanyedi, feel free to pick a single that has not started yet...
[...]

PR #2162, opened 2026-03-06
refactor docstrings in `scr/supervision/detection`

omkar-334:
<details>

- [x] Self-reviewed the code
- [x] All tests pass locally

</details>

## Description

refactored docstrings in 3 files -

1. src/supervision/detection/core.py
2. src/supervision/detection/line_zone.py
3. src/supervision/detection/vlm.py

## Type of Change
- 📝 Documentation update

## Motivation and Context
Part of #2088

## Testing

Screenshot 2026-03-06 at 5 01 35 PM


Borda (dismissed):

Borda:
@omkar-334, this is great, as stated in #2088 before sanitizing the docstring, we need to ensure that the typing is correct, so I enabled the check for files you edited in this PR, please see pre-commit/linter/mypy to resolve all remaining issues 🦝

omkar-334:
Screenshot 2026-03-10 at 4 23 33 PM

omkar-334:
> @omkar-334, this is great, as stated in #2088 before sanitizing the docstring, we need to ensure that the typing is correct, so I enabled the check for files you edited in this PR, please see pre-commit/linter/mypy to resolve all remaining issues 🦝

@Borda all tests and pre-commit are passing. Thanks!

omkar-334:
@Borda , most of the pre-commit errors are because of `strict = true`, and after checking, i think they lie in 4 categories. 

  1. `np.ndarray` → `npt.NDArray[...]`
In strict mode, the generics must be parameterized. np.ndarray is a generic type, so using it without parameters raises `Missing type parameters`
  2. `cast()`
 Some libraries (e.g., OpenCV, NumPy operations) return Any. If our function has a concrete return type, mypy raises errors.
the fix for this is -: use `cast(TargetType, value)` to explicitly tell mypy the expected type (no runtime impact).
  3. Explicit variable annotations 
  When mypy cannot infer a variable’s type (common with NumPy calls), we need to add a explicit annotation.
  annotation: `x: npt.NDArray[np.bool_] = np.zeros(...)`
  4. Strict mode also requires parameters for built-in generics.
   Examples:
    - `dict` → `dict[str, Any]`
    -  `list` → `list[str]`
    -  `tuple` → `tuple[int, int]`

Borda (approved):

PR #2178, opened 2026-03-20
feat: Detections.from_inference supports compressed RLE masks

### LINKED ISSUE #1952: does `rle_to_mask` to mask support UTF-8 encoded string?
### Search before asking

- [x] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Question

Summary of RLE formats supported by COCO dataset by Gemini2.5-pro:

"""
### Structure of the RLE Output

The output of the `mask.encode()` function is a dictionary with the following structure:

*   `'size'`: A list of two integers `[height, width]` representing the dimensions of the mask.
*   `'counts'`: This field contains the run-length encoded data. It can be one of two types:
    *   **A list of integers**: This is an uncompressed RLE. The list represents the lengths of alternating runs of 0s and 1s. For example, `[2, 3, 1, 1]` for a mask `[0, 0, 1, 1, 1, 0, 1]` means two 0s, followed by three 1s, one 0, and one 1. If the first pixel is a 1, the RLE starts with a 0 count for the initial run of 0s.
    *   **A UTF-8 encoded string**: This is a compressed RLE, which is more space-efficient. The `pycocotools` library handles the compression and decompression of this string.

"""

In particular my dataset use UTF-8 encoded string, because it is outputed by `pycocotools`

Does the `sv.DetectionDataset.from_coco` function support loading a dataset with UTF-8 encoded RLE segmentation masks? 

From what I can tell it calls the `rle_to_mask` function, which is the function that handles the conversion does not support it 
https://github.com/roboflow/supervision/blob/deb1c9c4f4b0cd678416a67c8a13f2ef8ed6878f/supervision/dataset/utils.py#L142-L195

leeclemnet:
<details>

- [x] Self-reviewed the code
- [x] Added/updated tests
- [x] All tests pass locally

</details>

## Description

Add RLE mask support to `sv.Detections.from_inference `for models that return compressed COCO RLE (e.g. SAM3, semantic segmentation), and extend `rle_to_mask/mask_to_rle` with compressed string encoding/decoding — implemented as a pure Python COCO RLE codec with no new dependencies.

cc @PawelPeczek-Roboflow

## Type of Change

## Motivation and Context

Workflows currently supports COCO RLE compressed string serialization for segmentation masks, originally for SAM3 instance segmentation and now also for semantic segmentation models where masks may not be contiguous and therefore can't be represented as polygons. We would like `sv.Detections.from_inference()` to handle such data and automatically decode it into binary masks as it currently does for polygon mask data.

The existing `rle_to_mask` util supports this conversion but not from compressed strings (see https://github.com/roboflow/supervision/issues/1952).

Closes https://github.com/roboflow/supervision/issues/1952

## Changes Made

* Add RLE mask support to sv.Detections.from_inference for models like SAM3 and semantic segmentation that return compressed COCO RLE instead of polygons
* Extend rle_to_mask to accept compressed COCO RLE strings (no new dependencies — pure Python decoder)
* Add compressed flag to mask_to_rle for producing COCO RLE strings
* Move rle_to_mask and mask_to_rle from dataset/utils to detection/utils/converters with the other converters - necessary to avoid circular dependency, and dataset.utils already imports from detection.utils.converters - backwards compatibility maintained through re-export
* Bugfix: rle_to_mask return type was uint8, now bool as per function signature

## Testing

## Google Colab (optional)

Colab link:


Borda (dismissed):

src/supervision/detection/utils/converters.py
```diff
@@ -304,6 +308,236 @@ def mask_to_polygons(mask: npt.NDArray[np.bool_]) -> list[npt.NDArray[np.int32]]
...
+@overload
+def mask_to_rle(
+    mask: npt.NDArray[np.bool_], compressed: Literal[False] = ...
+) -> list[int]: ...
+
+
+@overload
+def mask_to_rle(
+    mask: npt.NDArray[np.bool_], compressed: Literal[True] = ...
+) -> str: ...
```
Borda: why is this needed?

src/supervision/detection/utils/converters.py
```diff
@@ -304,6 +308,236 @@ def mask_to_polygons(mask: npt.NDArray[np.bool_]) -> list[npt.NDArray[np.int32]]
...
+
+
+def mask_to_rle(
```
Borda, line 453: Can we reuse methods from CompactMask?

Borda:
@leeclemnet could you pls check that updated version?

src/supervision/dataset/utils.py
```diff
@@ -5,19 +5,43 @@
...
+    """Deprecated. Use `supervision.detection.utils.converters.mask_to_rle`."""
+    return void(mask, compressed=compressed)  # type: ignore[no-any-return]
+
+
+@deprecated(target=_rle_to_mask, deprecated_in="0.28.0", remove_in="0.30.0")  # type: ignore[untyped-decorator]
+def rle_to_mask(
+    rle: npt.NDArray[np.integer[Any]] | list[int] | str | bytes,
+    resolution_wh: tuple[int, int],
+) -> npt.NDArray[np.bool_]:
+    """Deprecated. Use `supervision.detection.utils.converters.rle_to_mask`."""
```
Borda, line 42:
```suggestion
    """Deprecated. Use `supervision.detection.utils.converters.rle_to_mask`."""
    return void(mask, compressed=compressed)  # type: ignore[no-any-return]
```

Borda (approved):

PR #2199, opened 2026-04-03
fix: slice numpy array values in `custom_data` per row in CSVSink

### LINKED ISSUE #1397: Save detection area with CSVSink
### Search before asking

### Description

I would like to save all the Detections data to csv. Currently the `area` is not saved, and my approach below isn't successful. This FR is to save the area (and potentially other detection attributes that are currently not saved)
```python
with sv.CSVSink(csv_path) as sink:
    for detection in detections:
        sink.append(detection, {"area": detection.area})
AttributeError: 'tuple' object has no attribute 'area'
```

### Use case

I will perform filtering in a separate application 

### Additional

_No response_

farukalamai:
<details>

- [x] Self-reviewed the code
- [ ] Updated documentation, follow [Google-style](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods)
- [x] Added/updated tests
- [x] All tests pass locally

</details>

## Description

Fixes a bug in `CSVSink` and `JSONSink` where passing a numpy array as a
`custom_data` value wrote the **entire array on every row** instead of the
per-detection scalar value.

## Type of Change

## Motivation and Context

When users pass computed per-detection values like `detections.area` via
`custom_data`, each row should receive its own scalar — not the whole array.

```python
# Before (broken): every row got the full array
with sv.CSVSink("out.csv") as sink:
    sink.append(detections, custom_data={"area": detections.area})
# area column: [400.0, 400.0] on every row ❌

# After (fixed): each row gets its own value
# area column: 400.0, 400.0 ✅
```

The root cause was `row.update(custom_data)` inside the per-detection loop,
which blindly wrote the whole value. The fix applies the same per-index
slicing logic that `detections.data` already uses correctly.

Closes #1397

## Changes Made

- `src/supervision/detection/tools/csv_sink.py` — slice numpy array values in `custom_data` per detection row
- `src/supervision/detection/tools/json_sink.py` — same fix
- `tests/detection/test_csv.py` — added test case for numpy array in `custom_data`

## Testing

## Google Colab (optional)

Colab link:

## Additional Notes

The fix is backward compatible — scalar values in `custom_data` (e.g.
`{"frame_number": 42}`) continue to work as before, written as-is on every
row.


Borda:
Thanks for the fix @farukalamai! I've pushed 3 follow-up commits to your branch:

1. **`fix: restore non-ndarray custom_data passthrough in CSVSink`** — the new `else` branch used `hasattr(value, "__getitem__")` to decide whether to slice custom_data values per detection. This accidentally indexed dicts by integer 0, raising `KeyError: 0` for the existing "Complex Data" test case (`custom_data={"metadata": {"sensor_id": 101}}` etc.). Non-ndarray types should be written as-is; only `np.ndarray` values need per-row indexing. Changed to `row[key] = value` in the else branch.

2. **`test: add JSONSink test for numpy array custom_data slicing`** — symmetric test for JSONSink to match the CSV coverage you added.

3. **`test: add mixed-type custom_data test (ndarray + scalar together)`** — exercises all three dispatch paths at once: ndarray sliced per row + scalar broadcast to every row.

All pre-commit hooks pass and 15/15 tests pass.

Borda (approved):

PR #2203, opened 2026-04-07
fix: is_empty() returns False for empty tracker arrays (issue #2195)

### LINKED ISSUE #2195: [Bug]: Inconsistent behavior of sv.Detections.is_empty() if tracker_id is not None
### Search before asking

### Bug

If `is_empty` is called on a `Detections` object, `False` is returned if `tracker_id` is an empty array instead of `None`.

### Environment

- Supervision: 0.27.0.post2
- Python: 3.10.20
- OS: Ubuntu 24.04

```python
import supervision as sv
import numpy as np

detections_wo_tracker_id = sv.Detections(
    xyxy=np.array([[0, 0, 10, 10], [0, 0, 20, 30]]), 
    class_id=np.array([1, 2]), 
    confidence=np.array([0.6, 0.7])
)
print(detections_wo_tracker_id[detections_wo_tracker_id.class_id == 0].is_empty())
# True

detections_w_tracker_id = sv.Detections(
    xyxy=np.array([[0, 0, 10, 10], [0, 0, 20, 30]]), 
    class_id=np.array([1, 2]), 
    confidence=np.array([0.6, 0.7]), 
    tracker_id=np.array([1, 2])
)
print(detections_w_tracker_id[detections_w_tracker_id.class_id == 0].is_empty())
# False
```

Zeesejo:
## Problem

`sv.Detections.is_empty()` returned `False` when `tracker_id` was set to an empty array `np.array([])` instead of `None`. This happened because the previous implementation compared `self == Detections.empty()`, and `Detections.empty()` sets `tracker_id=None` — so the equality check failed for any instance where `tracker_id=[]`.

**Minimal repro (before fix):**
```python
import numpy as np
import supervision as sv

detections = sv.Detections(
    xyxy=np.empty((0, 4), dtype=np.float32),
    tracker_id=np.array([])  # empty array, not None
)
print(detections.is_empty())  # ❌ returned False
```

## Fix

Replaced the equality-based check with a direct length check:

```python
def is_empty(self) -> bool:
    return len(self) == 0
```

This is robust to any optional field (`tracker_id`, `confidence`, `class_id`, etc.) being an empty array rather than `None`, since `__len__` is based solely on the number of bounding boxes (`len(self.xyxy)`).

Fixes #2195


Borda (changes requested):
This PR git too wild with removing a large portion of the docs, pls focus only on the described chnaged and add relevant tests

UNakade:
Like I said in #2195, this isn't as simple as it looks. If people are relying on the information stored in empty arrays (like the expected shape of its elements), this PR will create breaking changes. The failing tests are some of the examples that rely on `is_empty()` returning `False` even if the `sv.Detections` object has `len == 0`.

Borda:
> Like I said in #2195, this isn't as simple as it looks. If people are relying on the information stored in empty arrays (like the expected shape of its elements), this PR will create breaking changes. The failing tests are some of the examples that rely on `is_empty()` returning `False` even if the `sv.Detections` object has `len == 0`.

That is fair, but still doesn't explain why you decided to remove that much docs without replacement and didn't add tests that would support correctness of your implementation...

UNakade:
> That is fair, but still doesn't explain why you decided to remove that much docs without replacement and didn't add tests that would support correctness of your implementation... 

Dude, I didn't open this PR! I stopped working on this once I realised it could create breaking changes.

Zeesejo:
Hey @Borda and @UNakade — thanks for the feedback, really appreciate you digging into this.

You're both right. I oversimplified the fix — the `len == 0` approach doesn't account for cases where users might be relying on empty arrays still carrying shape/type information, and the failing tests make that pretty clear. I should've caught that before opening the PR.

On the docs removal — that was unintentional. I didn't mean to drop the `from_lmm`/`from_vlm` docstring examples, those got caught up in the diff by mistake. I'll restore those.

My plan to get this back on track:
- Revert the unrelated doc changes
- Add proper regression tests covering the original issue (#2195) — specifically `xyxy` empty with `tracker_id=np.array([])` and other optional fields as empty arrays
- Revisit the `is_empty()` logic to handle the breaking change concern more carefully

Will push an update shortly. Sorry for the noise!

Borda:
> > That is fair, but still doesn't explain why you decided to remove that much docs without replacement and didn't add tests that would support correctness of your implementation...
> 
> Dude, I didn't open this PR! I stopped working on this once I realised it could create breaking changes.

Sorry, I accidentally swapped you with the author since you reacted first...

Borda:

Borda:
resolved in #2209

Issue #2106, opened 2026-01-27
Refactor documentation examples to `doctests` for continuous validation


Borda:
### Search before asking

- [x] I have searched the issues and discussions and found no similar feature request.

### 🚀 The Problem
Currently, many documentation examples within the codebase are written as static Markdown code blocks. While these serve as visual guides, they are not executed or verified by the CI/CD pipeline.

Because these examples are not part of the test suite:
1.  **Code Rot:** As the API evolves/refactors, documentation examples frequently become outdated or incorrect without anyone noticing.
2.  **User Friction:** Users copying "official" examples may encounter immediate errors, leading to confusion and a lack of trust in the docs.
3.  **Missed Coverage:** These examples often represent the primary "happy path" for users, yet they are currently excluded from regression testing.

### 💡 Proposed Solution
We should refactor these valid usage examples into Python's standard `doctest` format. This allows them to be executed and verified automatically by the test runner (e.g., via `pytest --doctest-modules`).
  
### 🏆 Justification (Industry Standard)
Adopting `doctest` is a widely accepted best practice in the Python ecosystem (standard in libraries like **Pandas**, **NumPy**, and **Scikit-Learn**).
* **Verification:** It ensures the documentation *is* a test. If the docs say it works, the CI proves it.
* **Maintenance:** It catches breaking changes in the public API immediately during PR checks.
* **User Experience:** It guarantees that the first piece of code a user tries will work as expected.

#### Current State (Untested static block)
```
  '''python
  import cv2
  import supervision as sv
  from ultralytics import YOLO
  
  image = cv2.imread(...)
  model = YOLO('yolov8s.pt')
  result = model(image)[0]
  detections = sv.Detections.from_ultralytics(result)
  '''
```
  
#### Desired State (Tested `doctest`)
```
  >>> import cv2
  >>> import supervision as sv
  >>> from ultralytics import YOLO
  >>> image = cv2.imread("tests/data/image.jpg")
  >>> model = YOLO('yolov8s.pt')
  >>> result = model(image)[0]
  >>> detections = sv.Detections.from_ultralytics(result)
  >>> len(detections) > 0  # verifies the code actually ran and produced output
  True
```

### Are you willing to submit a PR?

- [x] Yes I'd like to help by submitting a PR!

PR #2207, opened 2026-04-07
refactor: convert LineZone docstring example to doctest

vincenzonetti:
<details>

- [x] Self-reviewed the code

</details>

## Description
i converted static docstring examples to doctest format for the LineZones file

## Type of Change
- 📝 Documentation update

## Motivation and Context
Part of #2106
                                                                                                                      
  ## Additional note
It's actually my first PR on the project. I have only changed one file, but if the change happens to be correct, I could improve this PR by doing the conversion on more files.


src/supervision/detection/line_zone.py
```diff
@@ -53,25 +53,18 @@ class LineZone:
...
-        import supervision as sv
-        from ultralytics import YOLO
-
-        model = YOLO("<SOURCE_MODEL_PATH>")
-        tracker = sv.ByteTrack()
-        frames_generator = sv.get_video_frames_generator("<SOURCE_VIDEO_PATH>")
-        start, end = sv.Point(x=0, y=1080), sv.Point(x=3840, y=1080)
-        line_zone = sv.LineZone(start=start, end=end)
-
-        for frame in frames_generator:
```
Borda, line 66: @vincenzonetti lets tkeep this pattern and in the docstring create trivial generators which created, for example, an empty file frames

vincenzonetti:
@Borda I've updated the `LineZone` example following your suggestions:
1. Wrapped the example in ` ```pycon ``` ` fences.
2. Used a trivial `frames_generator` to keep the loop pattern.
3. Added a second frame to demonstrate and validate an actual crossing (where `in_count` increments).

Regarding the other files (like `Detections` class in `core.py`), many examples depend on external libraries like `torch`, `ultralytics`, or `transformers` which are not installed in the default CI environment. Is using `# doctest: +SKIP` for these imports and model calls the preferred approach in this project, or should we aim for a different way to make them verifiable?

Borda:
> Regarding the other files (like `Detections` class in `core.py`), many examples depend on external libraries like `torch`, `ultralytics`, or `transformers` which are not installed in the default CI environment. Is using `# doctest: +SKIP` for these imports and model calls the preferred approach in this project, or should we aim for a different way to make them verifiable?

For these, we can create an integration group in pyproject.toml, and in the docstrings, we can have a conditional skip if a dependency is missing... and have one simple CI job which would install all the integration group 🦝 

See section "Skip Unconditionally" in the https://pypi.org/project/pytest-doctestplus/ where this would be inside the condition

Borda (approved):

PR #2209, opened 2026-04-10
fix: is_empty() returns False for empty tracker arrays

### LINKED ISSUE #2195: [Bug]: Inconsistent behavior of sv.Detections.is_empty() if tracker_id is not None

abritton2002:
## Summary

Fixes #2195.

`is_empty()` incorrectly returned `False` on genuinely empty `Detections` objects when `tracker_id` was an empty numpy array rather than `None`.

**Root cause:** The previous implementation compared `self` against `Detections.empty()` via `__eq__`, which uses `np.array_equal`. When a `Detections` object with `tracker_id` set is filtered to zero results, `tracker_id` becomes `np.array([])` (not `None`). `np.array_equal(np.array([]), None)` returns `False`, so the equality check failed and `is_empty()` returned `False`.

**Fix:** Replace the comparison-based approach with a direct length check on `self.xyxy`, which is the canonical source of truth for detection count:

```python
# before
def is_empty(self) -> bool:
    empty_detections = Detections.empty()
    empty_detections.data = self.data
    empty_detections.metadata = self.metadata
    return bool(self == empty_detections)

# after
def is_empty(self) -> bool:
    return len(self.xyxy) == 0
```

## Test plan

- [x] Added 4 parametrized regression tests in `tests/detection/test_core.py::test_is_empty`
  - Canonical `Detections.empty()` → `True`
  - Non-empty detections without `tracker_id` → `False`
  - Filtered-to-empty detections **with** `tracker_id` (the regression case) → `True`
  - One detection remaining after filter → `False`
- [x] All tests pass: `uv run python -m pytest tests/detection/test_core.py::test_is_empty -v`


Borda:
### `is_empty()` Robustness Across Detection Types

| Detection type               | `xyxy` always populated?              | `is_empty()` correct?   |
| ---------------------------- | ------------------------------------- | ----------------------- |
| Bounding box                 | Yes — required field                  | ✓                       |
| OBB                          | Yes — axis-aligned wrapper always set | ✓                       |
| Segmentation (dense mask)    | Yes — computed via `mask_to_xyxy()`   | ✓                       |
| Segmentation (CompactMask)   | Yes — same, never materialized        | ✓ + O(1)                |
| VLM outputs (polygons, etc.) | Yes — connectors compute xyxy first   | ✓                       |
| Empty (`cls.empty()`)        | Yes — `(0, 4)` shape                  | ✓                       |
| `KeyPoints`                  | N/A — separate class                  | ⚠ same old bug, unfixed |

Borda (approved):

PR #2217, opened 2026-04-17
fix(annotators): TraceAnnotator handles stationary tracker ids in smooth mode

shaun0927:
<details>

- [x] Self-reviewed the code
- [x] Added/updated tests
- [x] All tests pass locally

</details>

## Description

Make `TraceAnnotator(smooth=True)` survive stationary tracker ids by removing consecutive duplicate anchor points before the spline fit and falling back to the raw polyline when fewer than four unique points remain.

## Type of Change

## Motivation and Context

`TraceAnnotator.annotate` feeds the raw trace coordinates to `scipy.interpolate.splprep`. When a tracker stays at the same anchor for several frames in a row — a very common case for parked vehicles, paused people, or any object that briefly holds position — the trace buffer contains consecutive duplicate points. `splprep` rejects that as a zero-length input curve and raises `ValueError: Invalid inputs.`, which kills the whole `annotate` call for every object in the scene:

```python
import numpy as np
import supervision as sv

frame = np.zeros((200, 200, 3), dtype=np.uint8)
det = sv.Detections(
    xyxy=np.array([[100, 100, 120, 120]], dtype=float),
    class_id=np.array([1]),
    tracker_id=np.array([42]),
)
annot = sv.TraceAnnotator(smooth=True, trace_length=10)
for _ in range(6):
    frame = annot.annotate(frame, det)   # raises ValueError("Invalid inputs.")
```

The minimal fix is to dedupe consecutive identical points before `splprep` (the non-smooth branch is unaffected) and to fall back to the raw polyline if fewer than four unique points remain, which is the minimum that `splprep` can fit with its default cubic spline. A moving tracker still triggers the smoothed render as before.

## Changes Made

- `src/supervision/annotators/core.py` — `TraceAnnotator.annotate` filters out consecutive duplicate anchor points before calling `splprep` and renders the raw polyline when not enough unique points remain.
- `tests/annotators/test_core.py` — new `TestTraceAnnotatorSmoothStationary` class with two regression tests: (1) stationary tracker annotates without raising over multiple frames, (2) a moving tracker still produces a smoothed trace distinct from the input image.

## Testing

## Additional Notes

Moving traces remain smoothed because the dedupe step only removes *consecutive* identical points; any positional drift is preserved. When more than three unique points exist the control flow is identical to today.

### DIFF
--- a/src/supervision/annotators/core.py
+++ b/src/supervision/annotators/core.py
@@ -2054,11 +2054,15 @@ class TraceAnnotator(BaseAnnotator):
             xy = self.trace.get(tracker_id=tracker_id)
             spline_points: npt.NDArray[np.int32] = xy.astype(np.int32)
 
-            if len(xy) > 3 and self.smooth:
-                x, y = xy[:, 0], xy[:, 1]
-                tck, _u = splprep([x, y], s=20)
-                x_new, y_new = splev(np.linspace(0, 1, 100), tck)
-                spline_points = np.stack([x_new, y_new], axis=1).astype(np.int32)
+            if self.smooth:
+                unique_xy = xy[
+                    np.concatenate(([True], np.any(np.diff(xy, axis=0) != 0, axis=1)))
+                ]
+                if len(unique_xy) > 3:
+                    x, y = unique_xy[:, 0], unique_xy[:, 1]
+                    tck, _u = splprep([x, y], s=20)
+                    x_new, y_new = splev(np.linspace(0, 1, 100), tck)
+                    spline_points = np.stack([x_new, y_new], axis=1).astype(np.int32)
 
             if len(xy) > 1:
                 cv2.polylines(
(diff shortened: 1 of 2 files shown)


src/supervision/annotators/core.py
```diff
@@ -2054,11 +2057,22 @@ def annotate(
...
+            if self.smooth:
+                unique_xy = xy[
+                    np.concatenate(([True], np.any(np.diff(xy, axis=0) != 0, axis=1)))
+                ]
+                if len(unique_xy) > 3:
+                    try:
+                        x, y = unique_xy[:, 0], unique_xy[:, 1]
+                        tck, _u = splprep([x, y], s=20)
+                        x_new, y_new = splev(np.linspace(0, 1, 100), tck)
+                        spline_points = np.stack([x_new, y_new], axis=1).astype(
```
Borda:
```suggestion
                        xy_new = splev(np.linspace(0, 1, 100), tck)
                        spline_points = np.stack(xy_new, axis=1).astype(
```

Borda (approved):

PR #2239, opened 2026-04-30
fix: return empty int ndarray instead of None for class_id on empty VLM parse

YousefZahran1:
## What

On empty-parse paths in two VLM parsers, `class_id` was returned as `None` instead of a zero-length `ndarray` of dtype `int`, inconsistent with every other exit point in these functions and with the convention established by `from_google_gemini_2_5`, `from_qwen_2_5_vl`, and `from_qwen_3_vl`.

### Affected code paths

| Function | Condition | Before | After |
|---|---|---|---|
| `from_paligemma` | regex finds no matches | `None` | `np.empty((0,), dtype=int)` |
| `from_google_gemini_2_0` | `JSONDecodeError` | `None` | `np.empty((0,), dtype=int)` |
| `from_google_gemini_2_0` | empty bounding-box list | `None` | `np.empty((0,), dtype=int)` |

## Why it matters

Callers that do `class_id.shape`, `len(class_id)`, or `class_id[mask]` immediately after parsing will get an `AttributeError` / `TypeError` whenever zero detections are returned — a very common path in production (frame with nothing to detect). Downstream `Detections.from_vlm(...)` pipelines hit this at runtime while unit tests may pass because they check happy paths.

## Changes

- `src/supervision/detection/vlm.py` — 3 one-line fixes (no logic change)
- `tests/detection/test_vlm.py` — updated 10 test expectations to match the new contract

All 88 tests in `tests/detection/test_vlm.py` pass.

Fixes #2219


Borda:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/roboflow/supervision?pullRequest=2239) before we can accept your contribution.

@YousefZahran1 could you pls sign ^^ 🦝

YousefZahran1:
"CLA signed! Ready for review when you have a moment.

src/supervision/detection/vlm.py
Borda:
```suggestion
) -> tuple[npt.NDArray[Any], npt.NDArray[Any], npt.NDArray[Any]]:
```

PR #2281, opened 2026-05-29
feat(detection): support windowed GeoTIFF reads in InferenceSlicer

### LINKED ISSUE #2027: Support for TIF files in InferenceSlicer
### Search before asking

- [x] I have searched the Supervision [issues](https://github.com/roboflow/supervision/issues) and found no similar feature requests.

### Description

Add TIF support to InferenceSlicer to enable processing of multi-GB drone and aerial geospatial imagery.

### Use case

High-resolution or large scale drone and aerial geospatial imagery is stored in GeoTIFF files which can exceed the memory of the host machine (50GB+). Using PIL to load these images into memory is not possible! A [windowed read](https://rasterio.readthedocs.io/en/stable/topics/windowed-rw.html) pattern is generally used to process the file in chunks, which is near exactly what the InferenceSlicer does now.

This would add `rasterio` as an optional dependency.

The `callback()` function will still receive a numpy array, but the dimensions will reflect the number of bands (color channels) in the TIF. It is up to the callback to select or convert the bands to suit the model inference.

A check of the TIF image's coordinate reference system is also required to ensure the CRS is a "projected" image (in XY coordinates suitable for 2D display rather than "world" coords like lat-lon).

### Additional

Reference implementation for my use-case attached.

```python
import supervision
import rasterio

import numpy
import warnings
from concurrent.futures import ThreadPoolExecutor, as_completed
from tqdm import tqdm

class RasterioWindowedInferenceSlicer(supervision.InferenceSlicer):
    def __call__(self, image: rasterio.DatasetReader, progress_desc="InferenceSlicer") -> supervision.Detections:
        """
        Perform tiled inference on the full image and return merged detections.

        Args:
            image (rasterio.DatasetReader): A GeoTIFF dataset for windowed read.
            progress_desc (str): Description for progress bar.

        Returns:
            Detections: Merged detections across all slices.
        """
        detections_list: list[supervision.Detections] = []
        resolution_wh = image.shape

        if not image.crs.is_projected:
            raise ValueError(f"A projected coordinate system is required for pixel inference on a 2D map. Use gdalwarp to reproject. crs={image.crs.to_string()}")

        offsets = self._generate_offset(
            resolution_wh=resolution_wh,
            slice_wh=self.slice_wh,
            overlap_wh=self.overlap_wh,
        )

        with ThreadPoolExecutor(max_workers=self.thread_workers) as executor:
            futures = [
                executor.submit(self._run_callback, image, offset) for offset in offsets
            ]
            with tqdm(total=offsets.shape[0], desc=progress_desc) as pb:
                for future in as_completed(futures):
                    detections_list.append(future.result())
                    pb.update(1)

        merged = supervision.Detections.merge(detections_list=detections_list)
        if self.overlap_filter == supervision.OverlapFilter.NONE:
            return merged
        if self.overlap_filter == supervision.OverlapFilter.NON_MAX_SUPPRESSION:
            return merged.with_nms(
                threshold=self.iou_threshold,
                overlap_metric=self.overlap_metric,
            )
        if self.overlap_filter == supervision.OverlapFilter.NON_MAX_MERGE:
            return merged.with_nmm(
                threshold=self.iou_threshold,
                overlap_metric=self.overlap_metric,
            )

        warnings.warn(
            f"Invalid overlap filter strategy: {self.overlap_filter}",
            category=supervision.SupervisionWarnings,
        )
        return merged

    def _run_callback(self, image: rasterio.DatasetReader, offset: numpy.ndarray) -> supervision.Detections:
        """
        Run detection callback on a sliced portion of the image and adjust coordinates.

        Args:
            image (ImageType): The full image.
            offset (numpy.ndarray): Coordinates `(x_min, y_min, x_max, y_max)` defining
                the slice region.

        Returns:
            Detections: Detections adjusted to the full image coordinate system.
        """
        # image_slice = crop_image(image=image, xyxy=offset)
        xywh = supervision.xyxy_to_xywh(numpy.expand_dims(offset, axis=0)).squeeze()
        window = rasterio.windows.Window(*xywh)
        image_slice = image.read(window=window)
        detections = self.callback(image_slice)
        resolution_wh = image.shape

        detections = supervision.detection.tools.inference_slicer.move_detections(
            detections=detections,
            offset=offset[:2],
            resolution_wh=resolution_wh,
        )
        return detections
```

madhavcodez:
## Description

Closes #2027.

`InferenceSlicer` can now run tiled inference directly on an open **rasterio-style dataset**, reading each tile via a *windowed read* instead of loading the whole image into memory. This makes it possible to slice multi-GB aerial/drone **GeoTIFFs** that don't fit in RAM — the exact pattern `InferenceSlicer` already implements, just sourced from disk windows rather than an in-memory array.

The original reporter outlined this with a `rasterio.windows`-based reference implementation; this PR integrates the capability into `InferenceSlicer` itself (the issue has been open and unclaimed for ~5 months).

```python
import rasterio
import supervision as sv

def callback(tile):  # tile is (H, W, C); select/convert bands as needed
    return model.predict(tile)

slicer = sv.InferenceSlicer(callback, slice_wh=640, overlap_wh=100)

with rasterio.open("large_orthomosaic.tif") as dataset:
    detections = slicer(dataset)
```

## Design

- **rasterio stays fully optional — the core library imports zero rasterio symbols.** Datasets are detected by duck typing (`_is_windowed_raster`: a callable `.read` plus `.crs`/`.width`/`.height`), and tiles are read with rasterio's tuple-window form `read(window=((row_start, row_stop), (col_start, col_stop)))`. Installable via `pip install "supervision[geotiff]"` (new optional extra).
- **Back-compat:** the existing in-memory NumPy/PIL path is unchanged — same `crop_image`, `resolution_wh`, `compact_masks`, out-of-bounds warning, and `move_detections` semantics. All existing `InferenceSlicer` tests pass unmodified.
- **CRS validation:** a geographic (non-projected) CRS raises `ValueError`, since pixel-space tiling on lat/lon is meaningless; `crs is None` (a plain TIFF) is allowed.
- **Tiles preserve the source dtype** (e.g. `uint16`) and all bands; selecting/converting is the callback's responsibility (documented in the `__call__` docstring).

## Tests

`tests/detection/tools/test_inference_slicer_geotiff.py` — runs **without** rasterio installed via a lightweight fake dataset, and includes a real `rasterio.MemoryFile` integration test guarded by `importorskip`:

- windowed path produces the same merged detections as the in-memory array path (no overlap **and** with `overlap_wh > 0`);
- each tile handed to the callback is byte-identical to the `crop_image` tile;
- `crs=None` works, geographic CRS raises, projected CRS does not;
- source dtype (`uint16`) is preserved through to the callback.

## Notes for reviewers

- **Scope:** this accepts an already-open dataset rather than a file path, deliberately, so that no `rasterio` import leaks into core. Happy to add a `str | Path` convenience overload (opening/closing the dataset internally) in a follow-up if you'd prefer that ergonomics.
- The accepted input type is expressed via a small structural `WindowedRasterDataset` `Protocol` so static type checkers accept a rasterio dataset without supervision depending on rasterio's types.
- Changelog updated under **UnReleased**.


SkalskiP:
@Borda, I like the idea. Adding GeoTIFF support would make Supervision, and potentially RF-DETR, much more useful for the geospatial community.

Let’s make sure the new dependencies stay optional and that everything works both with and without them.

I’m happy to do a deep code review.

madhavcodez:
Thanks @SkalskiP. Confirming the dependency stays fully optional:

- `rasterio` is declared only under `optional-dependencies.geotiff` (`pip install supervision[geotiff]`), never in the core deps.
- Nothing on the import path pulls in rasterio. The slicer detects a dataset structurally through the `WindowedRasterDataset` Protocol plus a duck-type check (`_is_windowed_raster`), so the existing array / PIL path is untouched and runs fine with rasterio absent. The windowed-read branch only executes when you hand it an already-open dataset.

I also addressed the two Copilot notes on strict-mypy narrowing by casting to `WindowedRasterDataset` before the raster-only attribute access at both call sites; `mypy --strict` on the file is clean. Rebased onto the latest develop. Glad to take you up on the deep review whenever you have time.

Borda:
@madhavcodez, could you pls share a link to the sample image I can try? 🦝

Borda:
tested with:
```py
import ssl
from pathlib import Path
from urllib.request import HTTPSHandler, build_opener, install_opener, urlretrieve

import certifi
import numpy as np
import rasterio

import supervision as sv

URL = "https://raw.githubusercontent.com/rasterio/rasterio/main/tests/data/RGB.byte.tif"
PATH = Path("RGB.byte.tif")


if not PATH.exists():
    context = ssl.create_default_context(cafile=certifi.where())
    install_opener(build_opener(HTTPSHandler(context=context)))
    urlretrieve(URL, PATH)  # noqa: S310


def callback(tile: np.ndarray) -> sv.Detections:
    h, w = tile.shape[:2]
    return sv.Detections(
        xyxy=np.array([[w * 0.25, h * 0.25, w * 0.75, h * 0.75]], dtype=float),
        confidence=np.array([0.9]),
        class_id=np.array([0]),
    )


slicer = sv.InferenceSlicer(
    callback=callback,
    slice_wh=(256, 256),
    overlap_wh=(64, 64),
    overlap_filter=sv.OverlapFilter.NONE,
)

with rasterio.open(PATH) as dataset:
    print(dataset.width, dataset.height, dataset.count, dataset.dtypes, dataset.crs)
    detections = slicer(dataset)

print(len(detections))
print(detections.xyxy[:5])
```

PR #2286, opened 2026-06-02
refactor: enrich `KeyPoints` with visibility mask, split confidence fields, and redesign  uncertainty annotator

SkalskiP:
## Summary

- **Split confidence into two fields**: `keypoint_confidence` (per-keypoint, shape `(N, M)`) and `detection_confidence` (per-detection, shape `(N,)`). The old `confidence` attribute is preserved as a deprecated property that forwards to `keypoint_confidence`.
- **Add `visible` mask**: A boolean array `(N, M)` that lets callers mark keypoints as occluded/invisible without dropping data. `VertexAnnotator` and `EdgeAnnotator` now skip keypoints (and edges) where `visible` is `False`.
- **Redesign uncertainty annotator**: Replace `VertexEllipseAnnotator` with `VertexUncertaintyAnnotator`, which draws concentric filled ellipses at configurable sigma levels with opacity blending and a red-yellow-green color gradient.
- **Clean up validators**: Remove deprecated shims (`validate_keypoint_confidence`, `validate_keypoints_fields`), rename `validate_confidence` to `validate_detection_confidence`, and add `validate_visible`. Improve error messages with dimension-specific diagnostics.


src/supervision/key_points/core.py
```diff
@@ -963,6 +830,44 @@ def __getitem__(
...
+
+        Args:
+            index: The index, indices, or key to access a subset of the KeyPoints
+                or an item from the data.
+
+        Returns:
+            A subset of the KeyPoints object or an item from the data field.
+
+        Examples:
+            ```python
```
Borda, line 886: could this be pycon doctest?

docs/keypoint/annotators.md
```diff
@@ -77,7 +77,7 @@ comments: true
 
     </div>
 
-=== "VertexEllipseAnnotator"
+=== "VertexUncertaintyAnnotator"
```
Borda: Let's keep the Ellipse as agreed

```suggestion
=== "VertexEllipseAnnotator"
```

src/supervision/key_points/annotators.py
```diff
@@ -192,129 +258,167 @@ def annotate(self, scene: ImageType, key_points: KeyPoints) -> ImageType:
...
-        if sigma <= 0:
-            raise ValueError("sigma must be positive")
-        if thickness <= 0:
-            raise ValueError("thickness must be positive")
+        if len(sigma_levels) == 0:
+            raise ValueError("sigma_levels must contain at least one value")
+        if any(s <= 0 for s in sigma_levels):
+            raise ValueError("All sigma_levels must be positive")
+        if not 0 < opacity <= 1:
+            raise ValueError("opacity must be between 0 (exclusive) and 1 (inclusive)")
```
Borda: lets have these leisted in docstrig too

src/supervision/key_points/annotators.py
```diff
@@ -192,129 +258,167 @@ def annotate(self, scene: ImageType, key_points: KeyPoints) -> ImageType:
...
+        colors = []
+        for i in range(n):
+            t = i / (n - 1)
+            if t < 0.5:
+                r = 255
+                g = int(255 * (t / 0.5))
+            else:
+                r = int(255 * ((1.0 - t) / 0.5))
+                g = 255
+            colors.append(Color.from_rgb_tuple((r, g, 0)))
```
Borda: I guess fine, but overall, Pepe suggests using at least 3-character variable names; single-character variables are challenging to rename...

src/supervision/key_points/annotators.py
```diff
@@ -192,129 +258,167 @@ def annotate(self, scene: ImageType, key_points: KeyPoints) -> ImageType:
...
-            >>> key_points = sv.KeyPoints(
-            ...     xy=np.array([[[50, 50], [60, 60]]], dtype=np.float32),
-            ...     data={"covariance": np.array([[[[25, 0], [0, 9]], [[9, 0], [0, 4]]]], dtype=np.float32)}
-            ... )
-            >>> annotator = sv.VertexEllipseAnnotator(color=sv.Color.GREEN)
-            >>> annotated_frame = annotator.annotate(image.copy(), key_points)
-            >>> annotated_frame.shape
-            (100, 100, 3)
-
+            ```python
```
Borda: This code seems easily executable, so why convert it from docstest to plain code that hasn't been tested?

src/supervision/key_points/annotators.py
```diff
@@ -192,129 +258,167 @@ def annotate(self, scene: ImageType, key_points: KeyPoints) -> ImageType:
...
-                        continue
-                    if confidence < self.confidence_threshold:
-                        continue
-                ellipse = self._covariance_to_ellipse(
-                    covariance=covariances[detection_index, point_index]
-                )
-                if ellipse is None:
+                if (
+                    key_points.visible is not None
+                    and not key_points.visible[detection_index, point_index]
```
Borda: Would it raise an exception if any index is out of range?

src/supervision/key_points/annotators.py
```diff
@@ -411,16 +473,20 @@ def __init__(
...
-            color: The color to use for each keypoint label. If a list is provided,
-                the colors will be used in order for each keypoint.
-            text_color: The color to use for the labels. If a list is provided, the
-                colors will be used in order for each keypoint.
+            color: The color to use for each keypoint label.
+                If a list is provided, the colors will be used
+                in order for each keypoint.
+            text_color: The color to use for the labels.
+                If a list is provided, the colors will be used
+                in order for each keypoint.
```
Borda: in general, I would use full line length then have too many short lines

src/supervision/key_points/core.py
```diff
@@ -131,6 +44,23 @@ class KeyPoints:
     class simplifies data manipulation and filtering, providing a uniform API for
     integration with Supervision [keypoints annotators](/latest/keypoint/annotators).
 
+    === "RF-DETR"
+
+        [RF-DETR](https://github.com/roboflow/rf-detr) keypoint models return
+        `sv.KeyPoints` directly from `model.predict()` — no additional
+        conversion is needed.
+
+        ```python
```
Borda, line 84: Here is correct to be plain code since rfdetr is an external dependency

src/supervision/key_points/core.py
```diff
@@ -273,17 +202,38 @@ class simplifies data manipulation and filtering, providing a uniform API for
 
     xy: npt.NDArray[np.float32]
     class_id: npt.NDArray[np.int_] | None = None
-    confidence: npt.NDArray[np.float32] | None = None
+    keypoint_confidence: npt.NDArray[np.float32] | None = None
+    detection_confidence: npt.NDArray[np.float32] | None = None
+    visible: npt.NDArray[np.bool_] | None = None
```
Borda, line 238: Later, it would be cleaner to move these typing to the class declaration, not init

src/supervision/key_points/core.py
```diff
@@ -941,20 +794,34 @@ def _get_by_2d_bool_mask(self, mask: npt.NDArray[np.bool_]) -> KeyPoints:
...
+                ]
+            if visible_selected is not None and self.visible is not None:
+                visible_selected[row] = self.visible[row, row_indices]
         return KeyPoints(
             xy=xy_selected,
-            confidence=conf_selected,
+            keypoint_confidence=keypoint_confidence_selected,
+            detection_confidence=self.detection_confidence.copy()
+            if self.detection_confidence is not None
+            else None,
```
Borda: For these, I would rather do an assignment upfront and in this final stage just pass the variable, then making multiple if/else

src/supervision/key_points/core.py
```diff
@@ -999,33 +904,57 @@ def __getitem__(
 
         xy_selected = self.xy[i, j]
 
-        conf_selected = self.confidence[i, j] if self.confidence is not None else None
+        keypoint_confidence_selected = (
+            self.keypoint_confidence[i, j]
+            if self.keypoint_confidence is not None
+            else None
+        )
```
Borda:
```suggestion
        keypoint_confidence_selected = None
        if self.keypoint_confidence is not None:
            keypoint_confidence_selected = self.keypoint_confidence[i, j]
```
just shorter notation

src/supervision/key_points/core.py
```diff
@@ -999,33 +904,57 @@ def __getitem__(
...
         data_selected = get_data_item(self.data, cast(Any, i))
 
         if xy_selected.ndim == 1:
             xy_selected = xy_selected.reshape(1, 1, 2)
-            if conf_selected is not None:
-                conf_selected = conf_selected.reshape(1, 1)
+            if keypoint_confidence_selected is not None:
+                keypoint_confidence_selected = keypoint_confidence_selected.reshape(
+                    1, 1
+                )
```
Borda, line 971: not for this PR, but overall expanding line length to 120 chars would make these lines nicer...

src/supervision/key_points/core.py
```diff
@@ -1153,13 +1082,15 @@ def as_detections(
...
-                confidence = None
-            else:
-                confidence = self.confidence[i]
+            if self.detection_confidence is not None:
+                confidence = np.array([self.detection_confidence[i]], dtype=np.float32)
+            elif self.keypoint_confidence is not None:
+                confidence = self.keypoint_confidence[i]
                 if selected_keypoint_indices:
                     confidence = confidence[selected_keypoint_indices]
                 confidence = np.array([confidence.mean()], dtype=np.float32)
```
Borda, line 1134: These three times of the same variable for various staff/types make it challenging to follow. Can we have more specific names?

Borda: ok, it was as diff only due to rename, so lets leave it for other code cleaning

src/supervision/validators/__init__.py
```diff
@@ -89,26 +89,29 @@ def validate_confidence(confidence: Any, n: int) -> None:
...
                 f"got shape {actual_shape}"
             )
-
-
-@deprecated(  # type: ignore[untyped-decorator]
-    target=validate_key_point_confidence,
-    deprecated_in="0.27.0",
-    remove_in="0.31.0",
-)
-def validate_keypoint_confidence(confidence: Any, n: int, m: int) -> None:
```
Borda, line 155: It was deprecated before, so now we revert and will keep it?

Borda: or just deprecate the argument?

src/supervision/validators/__init__.py
```diff
@@ -166,34 +173,52 @@ def validate_detections_fields(
     n = len(xyxy)
     validate_mask(mask, n)
     validate_class_id(class_id, n)
-    validate_confidence(confidence, n)
+    validate_detection_confidence(confidence, n)
     validate_tracker_id(tracker_id, n)
     validate_data(data, n)
 
 
+def validate_visible(visible: Any, n: int, m: int) -> None:
```
Borda:
```suggestion
def validate_visibility(visible: Any, n: int, m: int) -> None:
```

Borda: `visibile` as Any is a bit hard to undestand what the expected input is, especially when you do validation...

```suggestion
def validate_visible(visible: np.NDarray, n: int, m: int) -> None:
```

SkalskiP: It's a deliberate pattern. These are validation functions that accept untrusted input and check whether it matches the expected type/shape. Using Any is intentional because the input could be anything the user passes (wrong type, wrong shape, etc.)

src/supervision/key_points/annotators.py
SkalskiP: Ultimately, I just used `color: Color | Sequence[Color] = (Color.GREEN, Color.YELLOW, Color.RED),` nad this function was just not needed.

src/supervision/key_points/annotators.py
SkalskiP: I updated the docstring to:
```
opacity: Opacity of the overlay mask. Must be between ``0`` and ``1``.
```

src/supervision/key_points/core.py
SkalskiP: it's doctest now

src/supervision/key_points/annotators.py
SkalskiP: it's doctest now

src/supervision/key_points/core.py
SkalskiP: done

src/supervision/key_points/core.py
SkalskiP: done

src/supervision/key_points/annotators.py
SkalskiP: Yes, it would raise an `IndexError` if either `detection_index` or `point_index` is out of range for the `visible` array's shape `(n, m)`.

However, in practice this won't happen because:

- `detection_index` comes from iterating over `key_points.xy` (which has shape `(n, m, 2))`, so it's always in range `[0, n)`.
- `point_index` comes from enumerating the points within `xy[detection_index]`, so it's always in range `[0, m)`.
- the visible array is validated at construction time (in `__post_init__`) to have shape `(n, m)` — the same `n` and `m` as `xy`.

docs/keypoint/annotators.md
```diff
@@ -90,12 +90,12 @@ comments: true
     covariance = np.zeros((len(key_points), key_points.xy.shape[1], 2, 2), dtype=np.float32)
     key_points.data["covariance"] = covariance
 
-    ellipse_annotator = sv.VertexEllipseAnnotator(
+    uncertainty_annotator = sv.VertexUncertaintyAnnotator(
         color=sv.Color.GREEN,
         thickness=2,
         sigma=2.0,
     )
```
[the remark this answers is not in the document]
SkalskiP: this is outdated

src/supervision/key_points/annotators.py
SkalskiP: done

src/supervision/key_points/annotators.py
```diff
@@ -194,135 +260,153 @@ def annotate(self, scene: ImageType, key_points: KeyPoints) -> ImageType:
...
         self,
-        color: Color = Color.ROBOFLOW,
-        thickness: int = 2,
-        sigma: float = 2.0,
-        covariance_data_key: str = "covariance",
-        confidence_threshold: float = 0.0,
-        max_axis_length: float | None = None,
-        line_style: Literal["solid", "dashed"] = "solid",
-        dash_length: int = 16,
+        sigma: float | Sequence[float] = (1.0, 2.0, 3.0),
```
Borda, line 270: I would probably go with just one as default but that is also fine

src/supervision/key_points/annotators.py
```diff
@@ -194,135 +260,153 @@ def annotate(self, scene: ImageType, key_points: KeyPoints) -> ImageType:
...
+        color: Color | Sequence[Color] = (Color.GREEN, Color.YELLOW, Color.RED),
```
Borda, line 271: could be good to have also just palet as input and colours will be sapled based on nb sigmas

src/supervision/key_points/annotators.py
```diff
@@ -194,135 +260,153 @@ def annotate(self, scene: ImageType, key_points: KeyPoints) -> ImageType:
...
-                dominant uncertainty overlays.
-            dash_length: Arc length in degrees for each dashed segment. Only used
-                when ``line_style="dashed"``.
+            sigma: Sigma multipliers for each ring, drawn from outermost to
+                innermost.  Accepts a single float or a sequence of floats.
+                Defaults to ``(1.0, 2.0, 3.0)``.
+            color: The color for each sigma level.  Accepts a single
+                ``Color`` or a sequence of colors (one per sigma level).
+                Defaults to ``(Color.GREEN, Color.YELLOW, Color.RED)``.
+            opacity: Opacity of the overlay mask. Must be between ``0`` and
```
Borda: To match the colors, can we also have optical for each sigma?

src/supervision/key_points/annotators.py
```diff
@@ -411,15 +453,15 @@ def __init__(
...
         """
         Args:
-            color: The color to use for each keypoint label. If a list is provided,
-                the colors will be used in order for each keypoint.
-            text_color: The color to use for the labels. If a list is provided, the
-                colors will be used in order for each keypoint.
+            color: The color to use for each keypoint label. If a list is
+                provided, the colors will be used in order for each keypoint.
+            text_color: The color to use for the labels. If a list is
+                provided, the colors will be used in order for each keypoint.
```
Borda, line 723: seems like your IDE does not follow max docsting length

src/supervision/validators/__init__.py
```diff
@@ -226,6 +233,29 @@ def validate_xy(xy: Any, n: int, m: int) -> None:
...
+    """Validate per-keypoint visibility mask.
+
+    Expects a 2D bool ``np.ndarray`` with shape ``(n, m)``.
+    """
+    if visible is None:
+        return
+    actual_shape = str(getattr(visible, "shape", None))
+    if not isinstance(visible, np.ndarray) or visible.ndim != 2:
+        raise ValueError(
+            f"visible must be a 2D np.ndarray with shape (n, m), but "
```
Borda:
```suggestion
            "visible must be a 2D np.ndarray with shape (n, m), but "
```
seems there sometime overuse of f"

Borda (dismissed):

PR #2295, opened 2026-06-07
Adding .Claude folder

nrshrivatsan:
# What

Claude code with skills for users to get started

### DIFF
--- a/.claude/skills/cookbooks.md
+++ b/.claude/skills/cookbooks.md
@@ -1,0 +1,60 @@
+# Cookbooks Skill: Progressive Discovery of Computer Vision Recipes
+
+You are a "Cookbook Curator" responsible for guiding users through the vast collection of Supervision recipes and notebooks. Your goal is to progressively disclose information to avoid overwhelming the user while providing direct access to resources.
+
+## Conversational Style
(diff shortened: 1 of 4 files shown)


Borda (changes requested):
@nrshrivatsan, it seems like an interesting contribution. Could you share your motivation or why AGENTS?MD and extended contribution guide is not sufficient?

nrshrivatsan:
Idea is simple - Supervision could be of immense value to non coders such as
- Medical researchers
- Materials researchers
- Biologists 

However the barrier to entry could be lowered if they could use Claude code to jump start ( instead of cold start)

Hence the idea. 

Hope this makes sense.

Borda:
This is great to hear just to undestand they would most likely use supervison as a tool not develop features for supervison, right?

nrshrivatsan:
Yes- tool based use is the target access pattern.

nrshrivatsan:
Any updates on this PR?

Borda:
> Any updates on this PR?

Still figuring out the benefit compare to the CALUDE. md

Borda:
BTW, how was this generated?

PR #2309, opened 2026-06-11
refactor: split VertexEllipseAnnotator into Area, Outline, and Halo variants

SkalskiP:
## Summary

- Extract shared covariance logic into `_BaseVertexEllipseAnnotator` (validation, eigendecomposition, ellipse param iteration)
- Rename `VertexEllipseAnnotator` to `VertexEllipseAreaAnnotator` (filled semi-transparent ellipses); keep `VertexEllipseAnnotator` as backward-compatible alias
- Add `VertexEllipseOutlineAnnotator` — stroke-only concentric rings with configurable `thickness`
- Add `VertexEllipseHaloAnnotator` — radial gradient fill with quadratic falloff (full opacity at center, fading to zero at edge)

### DIFF
--- a/docs/keypoint/annotators.md
+++ b/docs/keypoint/annotators.md
@@ -77,7 +77,7 @@ comments: true
 
     </div>
 
-=== "VertexEllipseAnnotator"
+=== "VertexEllipseAreaAnnotator"
 
     ```python
     import supervision as sv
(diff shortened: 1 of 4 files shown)


docs/keypoint/annotators.md
```diff
@@ -77,19 +77,56 @@ comments: true
 
     </div>
 
-=== "VertexEllipseAnnotator"
+=== "VertexEllipseAreaAnnotator"
```
Borda, line 80: do we need extra class not just fill to be
- None for just outline
- True to follow the same color
- Color for custom color

Borda (commented):
I like the idea of filling the area, just not really convinced if we need a full new class for something which could be just one argument

SkalskiP:
Well in detections we had a convention of having separate Annotators and allow user to compose them. So let's do it constiantly.

Borda:
> Well in detections we had a convention of having separate Annotators and allow user to compose them. So let's do it constiantly.

I am fine with convention; however, you shall still be able to combine multiple annotators and one with fill and another with outline enabled by arg.

SkalskiP:
I'm merging.

PR #2312, opened 2026-06-12
fix(obb): NMM now computes geometric union via min-area rotated rect

Borda:
Previously with_nmm for OBB detections kept the winner's OBB geometry unchanged (only confidence was merged), making it inconsistent with AABB NMM which expands to the union envelope. Now computes cv2.minAreaRect over all N×4 corners from the merge group — the MARC degenerates to the axis-aligned union for zero-rotation OBBs, preserving full consistency with AABB NMM.

- Replace winner-OBB xyxy patch with MARC of all merged corners
- Update ORIENTED_BOX_COORDINATES in data to reflect merged geometry
- Rename test to reflect new expected behaviour (union, not winner AABB)
- Add consistency test asserting axis-aligned OBB NMM == AABB NMM xyxy


src/supervision/detection/core.py
```diff
@@ -2562,6 +2562,17 @@ def with_nmm(
             A new Detections object containing the subset of detections
                 after non-maximum merging.
 
+        Note:
+            For detections carrying oriented bounding box data
+            (``data[ORIENTED_BOX_COORDINATES]``), each merge group's output OBB
+            is the minimum-area rotated rectangle (``cv2.minAreaRect``) enclosing
```
kounelisagis: I think that the zero-rotation consistency doesn't hold for diagonal merge groups - `minAreaRect` picks the angle by area, and the winning tilted rect can stick out past every input:

```python
quads = np.stack([
    box_corners(0, 0, 20, 20),    # zero-rotation OBBs in a diagonal staircase,
    box_corners(12, 12, 32, 32),  # middle box has top score
    box_corners(24, 24, 44, 44),
])
detections.with_nmm(threshold=0.05).xyxy
# develop:     [[12, 12, 32, 32]]
# this branch: [[-10, -10, 54, 54]]  <- negative coords, outside every input
# AABB NMM:    [[0, 0, 44, 44]]
```

The new three-detection test doesn't catch it because its boxes are laid out horizontally, where the min-area rect happens to be axis-aligned.

Maybe build the enclosing rect at the winner's angle (project all corners onto the winner's axes and take min/max) instead of letting `minAreaRect` pick the angle. For zero-rotation inputs that yields exactly the axis-aligned union, and for the rotated test case in this PR it produces an identical rect - so the existing tests pass unchanged.

Borda: > Maybe build the enclosing rect at the winner's angle (project all corners onto the winner's axes and take min/max) instead of letting minAreaRect pick the angle. For zero-rotation inputs that yields exactly the axis-aligned union, and for the rotated test case in this PR it produces an identical rect - so the existing tests pass unchanged.

Sounds good to me and really appreciate you are paying attention and helping us...

kounelisagis (approved):
LGTM!

SkalskiP:
box_nmm_demo
box_nms_demo
obb_nmm_demo
obb_nms_demo

tests/detection/test_core.py
```diff
@@ -1056,31 +1060,715 @@ def test_falls_back_without_obb_data(self, method: str) -> None:
...
+            ),
+            pytest.param(
+                [
+                    _make_obb_detections(
+                        [
+                            np.array(
+                                [[0, 0], [10, 0], [10, 5], [0, 5]],
+                                dtype=np.float32,
+                            )
+                        ],
```
Borda, line 1234: feel would be easier to read if the function accepts lists and does the conversion to np array inside

PR #2314, opened 2026-06-12
docs(cookbooks): add Oriented Bounding Boxes cookbook

kounelisagis:
## Summary

Adds a cookbook on oriented bounding boxes (OBB). There is no OBB content in the cookbooks yet, and as of 0.29.0 the inference-side OBB pipeline can be shown end to end in supervision.

It runs a DOTA-pretrained `yolo11n-obb` on Ultralytics' standard OBB sample, a top-down photo of a marina of moored boats, and is built around one idea: how an oriented box differs from an axis-aligned one. A first oriented detection, a side-by-side where the axis-aligned boxes bloat and overlap their neighbours while the oriented boxes stay tight to each hull, the payoff in `with_nms` and `Detections.area`, and a YOLO OBB dataset export with `as_yolo(is_obb=True)`. The boxes stay large and easy to read throughout.

Install pin is `supervision>=0.29.0` now that the release is out.


Borda:
I like the realistic case, but I think having a case where the boxes are larger would do better in demonstrating the usage

kounelisagis:
Reworked the cookbook around a "AABB vs OBB" story, which I think makes it click. The annotations stay large and easy to read throughout. I believe it's useful now that 0.29.0 is out. Pretty happy with how it turned out.

Borda (dismissed):

Borda (approved):

PR #2320, opened 2026-06-14
feat(utils): add `TkImageWindow` to unblock switch to `opencv-python-headless`

Borda:
- Add sv.TkImageWindow: tkinter+pillow desktop window replacing cv2.imshow/waitKey/destroyAllWindows under headless OpenCV; supports BGR/grayscale/BGRA frames, key polling, mouse callbacks, context manager
- Switch pyproject.toml from opencv-python to opencv-python-headless (breaking: cv2.imshow/waitKey/namedWindow no longer provided transitively; restore with pip install opencv-python)
- Migrate all 17 runnable examples from cv2.imshow/waitKey/destroyAllWindows to sv.TkImageWindow
- Add docstring with webcam ownership pattern to VideoInfo.from_video_path and get_video_frames_generator
- Add breaking-change CHANGELOG entry and two FAQ entries covering headless and webcam usage
- Add 18 tests for TkImageWindow covering show(), wait_key(), close(), context manager, mouse callbacks, _bgr_to_pil


SkalskiP:
@Borda  I took a quick look at the code. I agree, and I like this direction. Let me know when it'll be ready.

Borda:
This is expected as it refers to the contribution of this PR
```
* [404] <https://supervision.roboflow.com/latest/utils/image_window/> | Rejected status code (this depends on your "accept" configuration): Not Found
```

examples/time_in_zone/inference_stream_example.py
```diff
@@ -1,121 +0,0 @@
-import cv2
```
SkalskiP, line 1: why have we removed `inference_stream_example.py`, `rfdetr_stream_example.py` and `ultralytics_stream_example.py` examples?

SkalskiP:
General question: what happens when someone installs `supervision` together with `inference` or `ultralytics` in the same Python environment? Which OpenCV package gets installed, one or both? If both are installed, which one is used at runtime? Should we have any recommendations for users who run into this setup?

examples/time_in_zone/scripts/draw_zones.py
```diff
@@ -1,18 +1,16 @@
 import json
```
SkalskiP, line 1: One thing flagged by the code review agent. Did you check whether mouse callbacks in `TkImageWindow` still receive image-pixel coordinates after the window gets resized?

The concern is that `_fit_image` scales and letterboxes the frame, but `_on_mouse` forwards raw `event.x` / `event.y` from the label widget. So in `draw_zones.py`, clicks might land in the wrong place after resizing and save incorrect polygons.

src/supervision/utils/image_window.py
```diff
@@ -0,0 +1,282 @@
...
+from typing import Any
+
+import numpy as np
+import numpy.typing as npt
+from PIL import Image
+
+MouseCallback = Callable[[int, int, str], None]
+
+
+class TkImageWindow:
```
SkalskiP: Let's just call it `sv.ImageWindow`.

src/supervision/utils/image_window.py
```diff
@@ -0,0 +1,282 @@
...
+        self._root: Any = None
+        self._label: Any = None
+        self._photo: Any = None
+        self._key_event: Any = None
+        self._key_queue: list[str] = []
+        self._pil_image: Image.Image | None = None
+        self._win_w: int = 0
+        self._win_h: int = 0
+
+    # ------------------------------------------------------------------
```
SkalskiP: Do we need comments like this? I'm not a big fan of those.

src/supervision/utils/image_window.py
```diff
@@ -0,0 +1,282 @@
...
+        scale = min(width / iw, height / ih)
+        new_w, new_h = max(1, round(iw * scale)), max(1, round(ih * scale))
+    else:
+        new_w, new_h = width, height
+    if (new_w, new_h) == (iw, ih):
+        return image
+    return image.resize((new_w, new_h), Image.Resampling.LANCZOS)
+
+
+def _bgr_to_pil(image: npt.NDArray[np.uint8]) -> Image.Image:
```
SkalskiP, line 316: We have `sv.cv2_to_pillow` already. Mabe we could use it here? If `sv.cv2_to_pillow` does not support some image format like grayscale or BGRA maybe we could just extend it?

docs/changelog.md
```diff
@@ -54,6 +54,28 @@ date_modified: 2026-07-06
...
+  pip uninstall -y opencv-python-headless
+  pip install opencv-python
+  ```
+  Both wheel families share the `cv2` namespace; keep only one installed at a time.
+
+  **Co-installation warning:** `opencv-python` and `opencv-python-headless` share the `cv2` namespace and install conflicting files to the same path. pip does not detect this — both wheels install silently, but the resulting `cv2` behavior is non-deterministic depending on install order. If you also install packages that bring in `opencv-python` (e.g. `ultralytics`, `inference-sdk`), you may end up with both wheels. The safe fix is to explicitly remove the headless wheel afterward:
+  ```bash
+  pip uninstall -y opencv-python-headless
+  pip install opencv-python
+  ```
```
SkalskiP: Feels like we are saying the same thing twice here. Maybe we could consolidate to make the text less verbose but equally informative.

docs/changelog.md
```diff
@@ -54,6 +54,28 @@ date_modified: 2026-07-06
...
+  - `wait_key()` returns a tkinter keysym `str` (e.g. `"q"`, `"Escape"`) or `None`, not an `int` — update `key == ord("q")` to `key == "q"`.
+  - Mouse callback signature is `(x: int, y: int, event_type: str)` where `event_type` is `"down"`, `"up"`, or `"move"` — incompatible with cv2's `(event, x, y, flags, param)`.
+  - Only left-button events are captured; scroll, right-button, and modifier flags have no equivalent.
+  - Requires `python3-tk` (not pip-installable): `sudo apt-get install python3-tk` on Debian/Ubuntu, `brew install python-tk` on macOS with Homebrew/pyenv.
+
+- Changed: `supervision` now depends on `opencv-python-headless` instead of `opencv-python`. The headless wheel provides the same `cv2` API except for desktop GUI functions (`cv2.imshow`, `cv2.waitKey`, `cv2.namedWindow`, and mouse/keyboard callbacks), which are not available in headless builds. All non-GUI `cv2` behaviour — drawing, text metrics, contour hierarchy, video I/O, affine transforms, image I/O — is unchanged.
```
SkalskiP: Feels like this shouldn’t be treated as a regular bullet point. It should be framed more like a warning, something that immediately grabs attention.

docs/faq.md
```diff
@@ -53,6 +53,38 @@ Install `supervision[metrics]`, then use `supervision.metrics.mean_average_preci
...
+
+while True:
+    ret, frame = cap.read()
+    if not ret:
+        break
+    # run your detector, then annotate:
+    # annotated = annotator.annotate(frame, detections)
+
+cap.release()
+```
```
SkalskiP, line 75: If this isn’t supported by our video utils yet, I think we should add support for it.

examples/time_in_zone/inference_stream_example.py
Borda: good catch, taht was not intended, will revert it back

examples/time_in_zone/scripts/draw_zones.py
Borda: I checked separately, but this is the case. I will double-check again

src/supervision/utils/image_window.py
Borda:
```suggestion
class ImageWindow:
```

src/supervision/utils/image_window.py
Borda: Just visual separation, but fine to prune them

pyproject.toml
```diff
@@ -50,7 +50,7 @@ dependencies = [
   "defusedxml>=0.7.1",
   "matplotlib>=3.6",
   "numpy>=1.21.2",
-  "opencv-python>=4.5.5.64",
+  "opencv-python-headless>=4.5.5.64,<5",
```
Borda:
```suggestion
  "opencv-python>=4.5.5.64,<5",
```

Borda:
@SkalskiP so holding on the switch to `opencv-python-headless` yet and adding just the GUI

PR #2330, opened 2026-06-16
perf: vectorize `mask_to_xyxy` and `KeyPoints.as_detections` (bit-identical)

RubenHaisma:
## Description

Two self-contained hot paths still used Python `for`-loops doing one object at a time. This replaces each with a single batched NumPy operation. Both changes produce **bit-identical output** (the operations only *select* existing values via `any`/`argmax`/`min`/`max`, no new arithmetic), so behaviour is unchanged — they're purely faster. No API or data-model changes.

### 1. `mask_to_xyxy` (`detection/utils/converters.py`)

The old implementation scanned every pixel of every mask with a per-mask `np.where` to recover coordinate arrays, then took min/max. It now reduces the whole `(N, H, W)` stack to per-row / per-column occupancy with `masks.any(axis=...)` and reads the tight bounds off those 1D profiles with `argmax` (first `True`) and a reversed `argmax` (last `True`) — one pass, no Python loop.

`mask_to_xyxy` runs once per frame on every instance-segmentation pipeline (`from_ultralytics`, `from_sam3`, and the Florence-2 / transformer connectors), so the per-pixel scan was a real cost.

**Benchmark** (mean of 20 runs):

| masks | before | after | speedup |
|------|--------|-------|---------|
| 10 × 640×640 | 9.12 ms | 0.15 ms | ~59x |
| 20 × 640×640 | 18.63 ms | 0.30 ms | ~61x |
| 50 × 320×320 | 11.46 ms | 0.33 ms | ~35x |
| 5 × 1080×1920 | 23.58 ms | 0.24 ms | ~97x |

### 2. `KeyPoints.as_detections` (`key_points/core.py`)

The old implementation built one single-row `Detections` per skeleton and combined them with an N-way `Detections.merge` (which re-concatenates every field). It now computes all bounding boxes in one vectorized masked min/max (`[0, 0]` "missing" keypoints excluded exactly as before) and constructs a single `Detections` directly, dropping the per-object objects and the merge.

The keypoint-confidence mean is intentionally still reduced per skeleton rather than via `mean(axis=1)`, because NumPy's axis reduction reorders the float32 summation and would shift results by an ULP versus the original.

**Benchmark** (17-keypoint COCO pose, keypoint-confidence path; the detection-confidence path is faster still):

| skeletons | before | after | speedup |
|------|--------|-------|---------|
| 17 | 0.24 ms | 0.066 ms | ~3.6x |
| 50 | 0.64 ms | 0.15 ms | ~4.3x |
| 100 | 1.25 ms | 0.26 ms | ~4.8x |

## Correctness

Both functions were checked for **bit-identical** output against the previous implementation with `np.array_equal`:

- `mask_to_xyxy`: 2000 random fixtures + edge cases (no masks, zero-size, single pixel, full mask, empty masks within a batch). 0 mismatches, dtype preserved.
- `KeyPoints.as_detections`: 800 random fixtures spanning detection-confidence / keypoint-confidence / no-confidence, with and without `selected_keypoint_indices`, and missing `[0, 0]` keypoints. `xyxy`, `confidence`, `class_id`, `data` and length all match. 0 mismatches.

## Tests

- Added direct unit tests for `mask_to_xyxy` (it previously had only indirect coverage), including a bit-identity check against the scalar loop.
- Added `KeyPoints.as_detections` tests for the missing-keypoint, `detection_confidence`, and `selected_keypoint_indices` paths.
- Full suite passes locally; `ruff` and `mypy` clean.

### DIFF
--- a/src/supervision/detection/core.py
+++ b/src/supervision/detection/core.py
@@ -2632,7 +2632,9 @@ class Detections:
         return Detections.merge(result)
 
 
-def _merge_obb_corners(corners_list: list[np.ndarray]) -> npt.NDArray[np.floating]:
+def _merge_obb_corners(
+    corners_list: list[npt.NDArray[np.floating]],
+) -> npt.NDArray[np.floating]:
     """Merge multiple OBB corner arrays using winner-angle projection.
 
     The first entry in *corners_list* is the winner. Its orientation angle
--- a/src/supervision/detection/utils/converters.py
+++ b/src/supervision/detection/utils/converters.py
@@ -204,7 +204,7 @@ def mask_to_xyxy(masks: npt.NDArray[np.bool_]) -> npt.NDArray[np.int_]:
     Converts a 3D `np.array` of 2D bool masks into a 2D `np.array` of bounding boxes.
 
     Args:
-        masks: A 3D `np.array` of shape `(N, W, H)` containing 2D bool masks.
+        masks: A 3D `np.array` of shape `(N, H, W)` containing 2D bool masks.
 
     Returns:
         A 2D `np.array` of shape `(N, 4)` containing the bounding boxes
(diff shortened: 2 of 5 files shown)


Borda:
@RubenHaisma pls avoid force push so we can process review

RubenHaisma:
Apologies, my mistake. I'll push follow-up commits instead of amending from here so the review history stays intact. Thanks for taking a look.

Borda (approved):

PR #2331, opened 2026-06-16
fix(metrics): count false positives on background images and absent classes

RubenHaisma:
## Description

`Precision` and `F1Score` silently drop predictions in two situations, which **inflates the reported score**:

1. **Background images** — an image that has predictions but **no ground truth** is skipped entirely (`if len(targets) > 0:`), so its predictions are never counted as false positives.
2. **Hallucinated classes** — predictions whose class never appears in *any* ground truth are dropped, because the confusion matrix is built only from ground-truth classes (`np.unique(true_class_ids)`). Even `MICRO` averaging (a global TP/FP count) is affected.

### This is a bug, not intended behavior

The library's own test documents the intended semantics — `tests/metrics/test_precision.py::test_empty_targets`:

> All predictions are false positives → TP = 0, FP = 1 → precision = 0 / 1 = **0.0**

That test passes today only *by accident*: with a single target-less image, `stats` ends up empty and `_compute` returns the all-zeros default. Add a second image with one true positive and the dropped false positives surface:

```python
from supervision.metrics import Precision
from supervision.metrics.core import AveragingMethod

pred_a = sv.Detections(np.array([[0,0,10,10]],float), class_id=np.array([0]), confidence=np.array([0.9]))
gt_a   = sv.Detections(np.array([[0,0,10,10]],float), class_id=np.array([0]))
bg     = sv.Detections(np.array([[20,0,25,5],[40,0,45,5],[60,0,65,5]],float),
                       class_id=np.array([0,0,0]), confidence=np.array([0.9,0.9,0.9]))

r = Precision(averaging_method=AveragingMethod.MICRO).update([pred_a, bg], [gt_a, sv.Detections.empty()]).compute()
print(r.precision_at_50)   # before: 1.0   after: 0.25  (1 TP, 3 FP)
```

## Fix

- Process images that have predictions even when they have no targets — every such prediction is a false positive.
- Build the confusion-matrix class set from the **union** of ground-truth and predicted classes. The existing `num_true == 0` branch in `_compute_confusion_matrix` already scores those as pure false positives; it was simply never reached.
- Guard `WEIGHTED` averaging against zero total support (returning `0`, as the empty case did before). This also fixes a latent `ZeroDivisionError` in the by-size breakdown when a size bucket contains predictions but no targets of that size.

Minimal change — the existing matching block is untouched; the diff just adds the false-positive branch and the class union.

## Notes on averaging semantics

- **MICRO** and **MACRO** now correctly reflect false positives from background images and absent classes.
- **WEIGHTED** still weights each class by its ground-truth support, so a false-positive-only class (support 0) legitimately does not move the weighted score — consistent with `sklearn`'s weighted precision. Use `MICRO`/`MACRO` to see those false positives.
- **`Recall` is intentionally left unchanged**: false positives do not enter recall, and target-less images / absent classes contribute no recall, so its identical-looking guard is already correct.

## Tests

Adds regression tests to both `Precision` and `F1Score` for (a) false positives on a background image and (b) false positives of a class with no ground truth. Both fail on `develop` and pass with this change. Full test suite, `ruff`, and `mypy` pass locally.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -11,6 +11,8 @@ date_modified: 2026-06-15
 
 - Fixed [#2322](https://github.com/roboflow/supervision/pull/2322): COCO export now preserves all polygon parts for multi-component masks. Previously, only the first polygon was written when a non-crowd mask had disjoint segments; all parts are now included.
 
+- Fixed [#2331](https://github.com/roboflow/supervision/pull/2331): `sv.Precision` and `sv.F1Score` now count predictions on background images (empty target set) as false positives, and count predictions of classes absent from ground truth as false positives under `MICRO` and `MACRO` averaging. Previously both edge cases were silently ignored, inflating scores. `WEIGHTED` averaging is unchanged — absent classes retain weight 0, consistent with scikit-learn. Users relying on previous scores should re-evaluate after upgrading; no API change is required.
+
 ### 0.29.0 <small>Jun 15, 2026</small>
 
 - Added [#2277](https://github.com/roboflow/supervision/pull/2277), [#2286](https://github.com/roboflow/supervision/pull/2286): [`sv.VertexEllipseAreaAnnotator`](https://supervision.roboflow.com/0.29.0/keypoint/annotators/#supervision.key_points.annotators.VertexEllipseAreaAnnotator), [`sv.VertexEllipseOutlineAnnotator`](https://supervision.roboflow.com/0.29.0/keypoint/annotators/#supervision.key_points.annotators.VertexEllipseOutlineAnnotator), and [`sv.VertexEllipseHaloAnnotator`](https://supervision.roboflow.com/0.29.0/keypoint/annotators/#supervision.key_points.annotators.VertexEllipseHaloAnnotator) for visualizing keypoint uncertainty as covariance ellipses. Requires models that output keypoint uncertainty (e.g. RF-DETR keypoint models).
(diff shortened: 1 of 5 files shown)


RubenHaisma:
A note on impact and scope for reviewers, since this changes metric values.

**Why it matters.** Any evaluation that includes background images (frames with predictions but no ground truth), or a model that predicts a class absent from the labels, currently has those false positives silently dropped — so `Precision` and `F1Score` report higher than the true value. The more background / hard-negative images in the eval set, the larger the inflation.

**It matches the intended behavior already encoded in the suite.** `tests/metrics/test_precision.py::test_empty_targets` asserts precision `0.0` for all-false-positive predictions. That test passes today only because a single target-less image leaves `stats` empty and returns the zero default; adding a second image with one true positive exposes the gap (precision `1.0` instead of `0.25`).

**Scope, so there are no surprises:**
- `MICRO` and `MACRO` now reflect these false positives.
- `WEIGHTED` (the default) still weights each class by its ground-truth support, so a false-positive-only class legitimately contributes zero weight — consistent with scikit-learn. Weighted results on normal datasets are unchanged.
- `Recall` is intentionally left untouched (false positives don't enter recall).
- The matching block is unchanged; the diff only adds the false-positive branch and the ground-truth ∪ predicted class set, plus a zero-support guard on `WEIGHTED` averaging (which also removes a latent `ZeroDivisionError` in the by-size breakdown when a size bucket has predictions but no targets).

Happy to split the zero-support guard into its own commit, or adjust anything to make review easier.

Borda (dismissed):

PR #2334, opened 2026-06-16
fix(json_sink): serialize NumPy scalars in `custom_data`

RubenHaisma:
## Description

`JSONSink` buffers rows and writes them all on close. A NumPy **scalar** passed via `custom_data` — for example a `np.int64` frame index — slips through the row builder unchanged (it is not an `np.ndarray`, so the existing `str(...)` guard does not apply), and then `json.dump` raises on close, **discarding every buffered row**.

### Repro

```python
import numpy as np
import supervision as sv

d = sv.Detections(xyxy=np.array([[0, 0, 10, 10]], float),
                  class_id=np.array([0]), confidence=np.array([0.9]))
with sv.JSONSink("out.json") as sink:
    sink.append(d, custom_data={"frame": np.int64(7)})
# TypeError: Object of type int64 is not JSON serializable  -> out.json is empty
```

(`np.float64` happens to work because it subclasses `float`, but `np.int64` / `np.float32` do not — so the failure depends on the dtype the user logs.)

## Fix

Pass a `default` serializer to `json.dump` that converts NumPy scalars with `.item()` and arrays with `.tolist()`. This only affects values that previously failed to serialize; existing string-formatted fields are unchanged. The numpy scalar now serializes as a proper JSON number (`"frame": 7`).

## Tests

Adds a regression test writing a `np.int64` and `np.float32` via `custom_data` and asserting they round-trip as JSON numbers. `ruff` and `mypy` pass locally.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -7,6 +7,11 @@ date_modified: 2026-06-15
 
 ### UnReleased
 
+- Fixed [#2334](https://github.com/roboflow/supervision/pull/2334): `sv.JSONSink` now
+  serializes NumPy scalars (e.g. `np.int64` frame indices) in `custom_data` as JSON
+  numbers instead of raising `TypeError` at close time. File handle is now guaranteed
+  to close even when serialization fails.
+
 - Fixed [#2335](https://github.com/roboflow/supervision/pull/2335): `sv.KeyPoints(confidence=...)` now works again. The `0.29.0` refactor accidentally dropped the deprecated `confidence` constructor kwarg; it is now accepted and mapped to `keypoint_confidence` with a deprecation warning.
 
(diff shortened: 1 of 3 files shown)


suzunn:
I like the `default` serializer approach here because it keeps the row builder focused on shaping detections and only handles JSON-specific coercion at the write boundary.

One small follow-up I would add is coverage for the other two branches in `_json_default`: a `custom_data` value containing a NumPy array, and an intentionally unsupported object that should still raise `TypeError`. That would make the Codecov miss useful instead of just satisfying coverage, and it protects the important part of the contract: NumPy values are normalized, but arbitrary custom objects are not silently stringified.

Borda (approved):

PR #2335, opened 2026-06-17
fix: accept deprecated `confidence` kwarg in `KeyPoints()` constructor

SkalskiP:
### Problem

Since 0.29.0, passing `confidence=` to the `KeyPoints()` constructor raises `TypeError` because the old field was replaced by a `@property` that forwards to `keypoint_confidence`. The property handles read/write on existing objects gracefully, but the dataclass-generated `__init__` rejects the unknown keyword argument entirely.

### Solution

Switch to `@dataclass(init=False)` and define an explicit `__init__` that:

- Accepts a keyword-only `confidence` parameter alongside the new `keypoint_confidence`
- Emits a deprecation warning when `confidence` is used
- Raises `ValueError` if both `confidence` and `keypoint_confidence` are passed
- Forwards `confidence` to `keypoint_confidence` transparently

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -7,6 +7,8 @@ date_modified: 2026-06-15
 
 ### UnReleased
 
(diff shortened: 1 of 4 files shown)


SkalskiP:
btw, here is how they solved this problem in inference package: https://github.com/roboflow/inference/pull/2467

Borda (approved):

PR #2339, opened 2026-06-17
perf(annotators): share mask painting and give HaloAnnotator the CompactMask path

kounelisagis:
## What

`MaskAnnotator` paints `CompactMask` detections into their bounding-box crop, but `HaloAnnotator` never got that path: it built every mask full frame, painted with full-frame boolean indexing, and allocated its foreground mask as a 2M-element Python list per frame. This pulls the shared compact/dense painting into one `_paint_masks_by_area` helper used by both annotators, so the fast path lives in one place.

It also fixes a latent `HaloAnnotator` bug: when the masks paint nothing, `gray / gray.max()` divided by zero and blacked out the whole frame. It now returns the scene unchanged.

## Impact

1080p frame, 30 masks: `HaloAnnotator` on `CompactMask` goes from ~219 ms to ~54 ms (about 4x), dense masks from ~227 ms to ~184 ms. `MaskAnnotator` is unchanged, output bit-identical.

## Tests

Output is verified pixel equal to before for both annotators on dense and compact masks. Added a parametrized compact-vs-dense equivalence test and a regression test for the empty-mask case; existing annotator tests pass.

### DIFF
--- a/src/supervision/annotators/core.py
+++ b/src/supervision/annotators/core.py
@@ -767,7 +767,11 @@ class HaloAnnotator(BaseAnnotator):
         colored_mask = cv2.blur(colored_mask, (self.kernel_size, self.kernel_size))
         colored_mask[fmask] = [0, 0, 0]
         gray = cv2.cvtColor(colored_mask, cv2.COLOR_BGR2GRAY)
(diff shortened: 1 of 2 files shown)


Borda:
image

PR #2346, opened 2026-06-20
perf(annotators): skip corner circles when drawing square label backgrounds

kounelisagis:
## What

Both rounded-rectangle helpers (`LabelAnnotator.draw_rounded_rectangle` and the public `draw_rounded_rectangle` in `draw/utils`) always drew two rectangles plus four corner circles, even when `border_radius` is 0. Zero is the default for `LabelAnnotator` and `VertexLabelAnnotator`, so the common case paid for six cv2 calls per label per frame where one fill rectangle does the same thing (the four circles are zero-radius no-ops, and one of the two rectangles already covers the whole box).

This adds a square-corner fast path to both helpers: when `border_radius <= 0`, draw a single fill rectangle and return. The `radius > 0` path is untouched.

## Impact

I profiled a typical 1080p frame with 100 labels and `_draw_labels` was the hottest part of the annotation, with 40k zero-radius `circle` calls. With the fast path:

- `LabelAnnotator` at 1080p, 100 labels: ~2.1 ms to ~1.3 ms per frame (about 1.6x), best of 7 runs
- the rounded-rectangle call itself at radius 0: ~2.8x faster in isolation
- `VertexLabelAnnotator` benefits too, since it uses the shared `draw/utils` helper

Output is pixel identical, verified against develop for `LabelAnnotator`, `VertexLabelAnnotator`, and `draw_rounded_rectangle` at both radius 0 and radius > 0.

I kept the two helpers separate rather than merging them: they take different inputs (xyxy tuple plus BGR vs `Rect` plus `Color`), and routing one through the other would add per-label object allocation in exactly the hot path this speeds up.

## Tests

Added tests pinning the square output to a plain rectangle for both helpers, plus a rounded case that checks corners are actually cut. The public `draw_rounded_rectangle` had no tests before this.

### DIFF
--- a/src/supervision/annotators/core.py
+++ b/src/supervision/annotators/core.py
@@ -1479,6 +1479,33 @@ class LabelAnnotator(_BaseLabelAnnotator):
         color: tuple[int, int, int],
         border_radius: int,
     ) -> npt.NDArray[np.uint8]:
+        """Draw a filled rectangle with optional rounded corners on an image.
+
+        Args:
+            scene: BGR image array to draw on; modified in-place and returned.
+            xyxy: Bounding box as (x1, y1, x2, y2) pixel coordinates.
+            color: Fill color as a BGR tuple (e.g. ``(0, 0, 255)`` for red).
+            border_radius: Corner rounding radius in pixels. Values <= 0
+                (including values clamped to 0 by a degenerate box) draw a
+                plain filled rectangle with square corners.
+
+        Returns:
+            The annotated ``scene`` array.
+
+        Example:
+            ```python
+            import numpy as np
+            import supervision as sv
+
+            scene = np.zeros((200, 200, 3), dtype=np.uint8)
+            scene = sv.LabelAnnotator.draw_rounded_rectangle(
+                scene=scene,
+                xyxy=(10, 10, 100, 50),
+                color=(0, 255, 0),
+                border_radius=0,
+            )
+            ```
+        """
         x1, y1, x2, y2 = xyxy
         width = x2 - x1
         height = y2 - y1
--- a/src/supervision/annotators/core.py
+++ b/src/supervision/annotators/core.py
@@ -1420,7 +1420,7 @@ class LabelAnnotator(_BaseLabelAnnotator):
 
             box_xyxy = label_property[:4].astype(int)
 
-            self.draw_rounded_rectangle(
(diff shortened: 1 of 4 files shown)


src/supervision/annotators/core.py
```diff
@@ -1467,7 +1467,7 @@ def _draw_labels(
                 current_y += text_h + self.text_padding  # Move to next line position
 
     @staticmethod
-    def draw_rounded_rectangle(
+    def _draw_rounded_rectangle(
```
Borda: not sure about changing this from public to private...

kounelisagis: 🤦‍♂️

Borda:
@kounelisagis, could you pls add visualization before and after? 🦝

kounelisagis:
> @kounelisagis, could you pls add visualization before and after? 🦝

Sure! Here is before (develop) vs after (this PR), for both the square default and a rounded radius, with an amplified diff on the right:

label_before_after


Top row is `border_radius=0` (the default), bottom row is `border_radius=25`. The diff panel is `absdiff x20`, so any change would light up; it stays fully black in both cases. Output is pixel identical (max abs diff = 0) for square and rounded alike. The square path now draws one filled rectangle instead of two rectangles plus four zero-radius corner circles, and the rounded path is untouched.

<details>
<summary>How it was generated</summary>

```python

scene = np.zeros((360, 640, 3), np.uint8)
scene[:] = np.linspace(40, 120, 640)[None, :, None].astype(np.uint8)
det = sv.Detections(
    xyxy=np.array([[40, 40, 200, 150], [260, 60, 460, 180],
                   [120, 210, 360, 330], [420, 210, 600, 320]], float),
    confidence=np.array([0.91, 0.86, 0.78, 0.69]),
    class_id=np.array([0, 1, 2, 1]),
    data={"class_name": np.array(["person", "car", "dog", "car"])},
)
box = sv.BoxAnnotator(thickness=2)
for radius in (0, 25):
    lab = sv.LabelAnnotator(border_radius=radius, text_scale=0.7, text_padding=8)
    img = lab.annotate(box.annotate(scene.copy(), det), det)
```

Run on `develop` and on this branch, then compare with `cv2.absdiff` (max abs diff is 0 in both cases).
</details>

Borda (dismissed):

PR #2353, opened 2026-06-24
fix(detection): do not crash `from_inference` on partial `tracker_id`

kounelisagis:
## Bug

`Detections.from_inference` raises when a Roboflow/Inference result has some predictions with a `tracker_id` and some without:

```
ValueError: tracker_id must be a 1D np.ndarray with shape (3,), but got shape (2,)
```

In `process_roboflow_result`, every branch appends `xyxy`, `confidence`, `class_id` and `class` for each prediction, but `tracker_id` was appended only when the key was present. A mixed batch then leaves `tracker_ids` shorter than the boxes, and `Detections` validation rejects it.

To be clear about scope: ByteTrack (both supervision's own and the inference block) attaches a `tracker_id` to every detection, so a standard tracking pipeline won't hit this. It comes up when predictions are assembled or merged by hand, or come from a source that only tracks some of them. So this is more of a robustness fix: the connector degrades gracefully instead of raising a confusing shape error on that input.

## Fix

Collect `tracker_id` for every prediction (`None` when absent) and build the array only when all detections carry one, otherwise keep it `None`. Fully tracked results still get the array, fully untracked results still get `None`, so existing behavior is unchanged; only the previously crashing mixed case changes (it now drops `tracker_id` rather than misaligning).

## Tests

Added a `process_roboflow_result` case for a mixed batch and an end to end `from_inference` test that asserts it no longer raises and keeps all detections.

### DIFF
--- a/src/supervision/detection/utils/internal.py
+++ b/src/supervision/detection/utils/internal.py
@@ -143,15 +143,13 @@ def process_roboflow_result(
             class_name.append(prediction["class"])
             confidence.append(prediction["confidence"])
             masks.append(mask)
-            if "tracker_id" in prediction:
-                tracker_ids.append(prediction["tracker_id"])
+            tracker_ids.append(prediction.get("tracker_id"))
         elif "points" not in prediction:
             xyxy.append([x_min, y_min, x_max, y_max])
             class_id.append(prediction["class_id"])
             class_name.append(prediction["class"])
             confidence.append(prediction["confidence"])
-            if "tracker_id" in prediction:
-                tracker_ids.append(prediction["tracker_id"])
+            tracker_ids.append(prediction.get("tracker_id"))
         elif len(prediction["points"]) >= 3:
             polygon = np.array(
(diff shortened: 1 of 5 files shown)


src/supervision/detection/utils/internal.py
```diff
@@ -143,15 +143,13 @@ def process_roboflow_result(
             class_name.append(prediction["class"])
             confidence.append(prediction["confidence"])
             masks.append(mask)
-            if "tracker_id" in prediction:
-                tracker_ids.append(prediction["tracker_id"])
+            tracker_ids.append(prediction.get("tracker_id"))
```
Borda, line 149: I think if tracker_id is not in prediction, then this adds None to tracker_ids

kounelisagis: Appending `None` keeps `tracker_ids` aligned one-per-prediction, and the array is only built when none are missing, so a mixed batch becomes `tracker_id=None` instead of a misaligned array.

Borda (approved):

PR #2372, opened 2026-06-29
feat: add image URL loader

Erol444:
## Summary

Adds `load_image_from_url()` to `supervision.utils.image` and exports it from the top-level `supervision` package. The helper validates and normalizes HTTP(S) URLs, downloads the image with `requests`, decodes it through OpenCV, and raises clear errors for invalid URLs, request failures, or undecodable image bytes.

Also adds mkdocs API documentation and focused unit tests for successful decoding, invalid image bytes, request failures, and non-HTTP URL rejection.

## Validation

- `uv run --frozen pytest tests/utils/test_image.py tests/utils/test_video.py`
- `SKIP=mypy uv run --frozen pre-commit run --files src/supervision/utils/internal.py src/supervision/utils/image.py src/supervision/utils/video.py src/supervision/__init__.py tests/utils/test_image.py tests/utils/test_video.py docs/utils/image.md docs/utils/video.md`
- `git diff --check`

Note: the local mypy hook was skipped because it fails before checking this code due to a NumPy stub syntax issue under the repo's configured Python 3.9 target.


src/supervision/utils/image.py
```diff
@@ -26,13 +24,64 @@
...
+        image = cv2.imdecode(
+            np.frombuffer(response.content, dtype=np.uint8),
+            cv_imread_flags,
+        )
+    finally:
+        response.close()
+    if image is None:
+        raise ValueError("Data pointed by URL could not be decoded into image.")
+
+    return image
```
Borda: How about also saving the image locally as an artifact?
```py
# 1. Generate a safe temporary file path
with tempfile.NamedTemporaryFile(delete=False, suffix=".jpg") as temp_file:
    temp_path = temp_file.name

# 2. Download using the existing utility
download_assets(url=url, save_path=temp_path)

# 3. Read it as a standard OpenCV array
image = cv2.imread(temp_path)

File: src/supervision/utils/image.py [hunk 2]
```diff
@@ -26,13 +24,64 @@
...
+        requests.RequestException: If the request fails or returns an error status.
+
+    Examples:
+        ```pycon
+        >>> import supervision as sv
+        >>> image = sv.load_image_from_url(
+        ...     "https://media.roboflow.com/notebooks/examples/dog-9.jpeg"
+        ... )  # doctest: +SKIP
+        >>> image.shape  # doctest: +SKIP
+        (576, 768, 3)
```
Borda: pls do not write doctest when you skip each line... 🦝

src/supervision/utils/image.py
Erol444: why? to cache it for next req?

src/supervision/utils/image.py
```diff
@@ -27,12 +32,124 @@
...
+        np.frombuffer(value, dtype=np.uint8),
+        cv_imread_flags,
+    )
+    if image is None:
+        raise ValueError("Data pointed by URL could not be decoded into image.")
+
+    return cast(npt.NDArray[np.uint8], image)
+
+
+def _write_image_url_cache(cache_path: Path, value: bytes) -> None:
```
Borda: Does downloading assets already do it?

src/supervision/utils/internal.py
```diff
@@ -46,6 +47,43 @@ def warn_deprecated(message: str) -> None:
     warnings.warn(message, category=SupervisionWarnings, stacklevel=2)
 
 
+def prepare_url(value: str) -> str:
```
SkalskiP: `utils/internal.py` is focused on deprecation and introspection helpers. A URL normalizer used by a single caller feels out of place there.

SkalskiP:
`supervision` already has an `assets` package for downloading files from our curated pool of images and videos. Since this new utility also downloads images, I think we should keep its behavior close to the existing assets flow.

In particular, it should use the same hashing strategy, cache files in the same location, and follow similar timeout, retry, and loading behavior.

At the moment, these two paths do not share any code, and their behavior is quite different.

I’d extract one small primitive into `utils/file.py`:

```python
def _download_to_file(url: str, target: Path, *, timeout: float = 30.0, stream: bool = False) -> None:
    """Download `url` to `target` atomically using a temp file and `os.replace`."""
```

Then both `load_image_from_url` and `download_assets` would call it.

SkalskiP (dismissed):

src/supervision/utils/image.py
Erol444: Done — the loader now saves the download locally: bytes go through the shared `_download_to_file` primitive into a cache file (md5-of-URL name), and `cv2` reads from that file.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/supervision/utils/image.py
Erol444: Changed — the example is now a plain `python` fenced block with no doctest prompts, so nothing is skipped.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/supervision/utils/image.py
Erol444: It does now — both `download_assets` and `load_image_from_url` call the shared `_download_to_file` primitive in `utils/file.py` (atomic temp file + `os.replace`).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/supervision/utils/internal.py
```diff
@@ -1,10 +1,13 @@
 import functools
 import inspect
 import os
+import urllib.parse
 import warnings
 from collections.abc import Callable
 from typing import Any, Generic, TypeVar
 
+import requests
+
```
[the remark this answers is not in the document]
Erol444: Resolved by relocating `prepare_url` to `supervision/utils/file.py` (which already imports `requests`); `utils/internal.py` no longer touches `requests`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/supervision/utils/internal.py
Erol444: Moved `prepare_url` to `utils/file.py`, next to the new shared `_download_to_file` download primitive — `utils/internal.py` is back to deprecation/introspection helpers only.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/supervision/utils/image.py
```diff
@@ -27,12 +32,124 @@
...
+    Examples:
+        ```python
+        import supervision as sv
+
+        image = sv.load_image_from_url(
+            "https://media.roboflow.com/quickstart/dog.jpeg"
+        )
+        image.shape
+
+        ```
```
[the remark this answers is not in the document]
Erol444: Keeping the plain `python` block: @Borda explicitly asked not to write doctests where every line is skipped, so `pycon` + `# doctest: +SKIP` is out here.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/supervision/utils/file.py
```diff
@@ -1,11 +1,100 @@
...
 from pathlib import Path
+from shutil import copyfileobj
 from typing import Any
 
 import numpy as np
+import requests
 import yaml
 
 
+def prepare_url(value: str) -> str:
```
SkalskiP: This function should be private since we do not intend for users to call it directly. Please use the `_*` prefix.

I also think the name should be less generic and more descriptive. How about `_normalize_http_url`? We should also use `url` instead of `value` for the argument name.

src/supervision/utils/image.py
```diff
@@ -27,12 +31,116 @@
...
 from supervision.utils.iterables import create_batches, fill
 
 RelativePosition = Literal["top", "bottom"]
 
 MAX_COLUMNS_FOR_SINGLE_ROW_GRID = 3
 
+DEFAULT_IMAGE_URL_CACHE_DIR = Path(tempfile.gettempdir()) / "supervision" / "image-url"
+
+
+def _get_image_url_cache_path(value: str, cache_dir: str | Path | None) -> Path:
```
SkalskiP, line 48: `assets` writes to `CWD`, while the URL loader writes to `{tmpdir}/supervision/image-url/`. Could we at least share a common `supervision` cache location?

src/supervision/utils/file.py
SkalskiP: Most failures return the generic `Provided URL is invalid`, while an unsupported scheme gets `Only HTTP(S) URLs are supported`. Should we make these errors more descriptive?

src/supervision/utils/file.py
```diff
@@ -1,11 +1,100 @@
...
+
+    if parsed_url.scheme not in {"http", "https"}:
+        raise ValueError("Only HTTP(S) URLs are supported")
+    if parsed_url.hostname is None:
+        raise ValueError("Provided URL is invalid")
+
+    return prepared_url
+
+
+def _download_to_file(
```
SkalskiP, line 54: thanks for the refactor! `_download_to_file` is exactly what I had in mind

tests/utils/test_file.py
```diff
@@ -1,10 +1,89 @@
...
+        # given
+        url = "https://media.roboflow.com/quickstart/dog.jpeg"
+
+        # when
+        result = prepare_url(value=url)
+
+        # then
+        assert result == url
+
+    @pytest.mark.parametrize(
```
SkalskiP, line 29: can we add explicit scheme-rejection cases? `ftp://`, `javascript:` etc.

src/supervision/utils/image.py
```diff
@@ -27,12 +31,116 @@
...
+        np.frombuffer(value, dtype=np.uint8),
+        cv_imread_flags,
+    )
+    if image is None:
+        raise ValueError("Data pointed by URL could not be decoded into image.")
+
+    return cast(npt.NDArray[np.uint8], image)
+
+
+def load_image_from_url(
```
SkalskiP, line 80: `docs/changelog.md` has no entry for `sv.load_image_from_url`.

src/supervision/utils/image.py
Erol444: Added a shared `SUPERVISION_CACHE_DIR = {tmpdir}/supervision` root in `utils/file.py`; the URL loader cache is now `SUPERVISION_CACHE_DIR / "image-url"`. Left `download_assets` defaulting to CWD since its docstring/return contract promises that for backward compatibility — happy to move it under the shared root too if you prefer.

src/supervision/utils/image.py
```diff
@@ -27,12 +31,126 @@
...
+        except ValueError:
+            cache_path.unlink(missing_ok=True)
+
+    _download_to_file(prepared_url, cache_path, timeout=timeout)
+    try:
+        return _decode_image_from_bytes(
+            value=cache_path.read_bytes(),
+            cv_imread_flags=cv_imread_flags,
+        )
+    except ValueError:
```
Borda, line 150: Shall this be rather `finalize`?

Borda (approved):

src/supervision/utils/image.py
Erol444: finally would run on success as well. In case of error it deletes file and raises error

Borda (approved):

Issue #2318, opened 2026-06-14
[Feature]: Geometry-aware dispatch to stop the recurring "silent AABB" bug class


kounelisagis:
### Search before asking

- [x] I have searched the issues and discussions and found no similar feature request.

### Feature Description

**What:** A small geometry-aware dispatch layer so detection ops use the geometry each detection actually carries (mask / OBB / box) instead of silently falling back to the axis-aligned `xyxy`.

**Why:** Several recent fixes are the same bug in different functions: code grabs `detections.xyxy` and forgets the real geometry lives in `data["xyxyxyxy"]` (OBB) or `mask`. Because `xyxy` is always present, the wrong answer is the silent default:
- `Detections.area` returned the AABB area, not the rotated body (#2306)
- `with_nms` / `with_nmm` dropped crossed OBBs via AABB IoU (#2303)
- `as_yolo` dropped OBB rotation on export (#2289)
- `roboflow/inference`: consensus IoU, and stitch / crop corner shifts (https://github.com/roboflow/inference/pull/2420, https://github.com/roboflow/inference/pull/2427, https://github.com/roboflow/inference/pull/2430)

Same root cause each time: `Detections` holds several geometries at once and `xyxy` is just a derived envelope, so every consumer re-implements "which geometry is the real one?" and forgetting it fails silently. The next geometry-aware function will hit it again.

**How:** Centralize the dispatch into a few batched helpers and route call sites through them instead of reaching for `xyxy`. No data-model change, keeps the vectorized design, backward compatible. Migration can be incremental, then deprecate the ad-hoc paths.

Proposed helpers (batched, no per-detection loops):
- `detection_iou(a, b)`: dispatches mask / OBB / box IoU
- `detection_area(det)`: mask sum / shoelace / box area
- `move_detections(det, offset)`: shifts `xyxy`, `mask`, OBB corners and keypoints together (what #2427 / #2430 each forgot)
- a `test_op_respects_geometry` contract test that fails when an op can't tell an OBB from its envelope

Rollout can be incremental and each step is independent / backward compatible: (1) consolidate the duplicated `move_detections` (pure dedup, no behavior change), (2) the read-only `detection_iou` / `detection_area` dispatch, (3) the contract test.

[...]

PR #2374, opened 2026-06-29
refactor: centralize geometry-aware dispatch for detection_area and detection_iou

mvanhorn:
<details>

- [x] Self-reviewed the code
- [x] Added/updated tests

</details>

## Description

Centralizes the "which geometry is the real one?" dispatch for detection area and IoU into a single internal module so the mask -> OBB -> box precedence lives in one place instead of being re-implemented (and silently forgotten) at each call site.

## Type of Change

- 🔨 Refactoring (no functional changes)

## Motivation and Context

A recurring "silent AABB" bug class keeps reappearing: code reaches for `detections.xyxy` and forgets the real geometry lives in `data["xyxyxyxy"]` (OBB) or `mask`. Because `xyxy` is always present, the axis-aligned answer is the silent default. Recent fixes (#2306 area returning AABB area, #2303 `with_nms`/`with_nmm` dropping crossed OBBs via AABB IoU, #2289 `as_yolo` dropping OBB rotation) share this root cause: every consumer re-derives geometry precedence and forgetting it fails silently.

@kounelisagis endorsed (2026-06-17) keeping the policy internal ("exact when richer geometry is present") behind a single geometry-resolution helper rather than a public per-op mode argument, and suggested a metamorphic test.

This PR is a deliberate first slice of the multi-part issue: the read-only dispatch helpers plus the contract test land first. `move_detections` consolidation and the `with_nms`/`with_nmm`/`as_yolo` call-site migration are left as independent follow-ups per the issue's incremental rollout.

Refs #2318

## Changes Made

- Add internal `src/supervision/detection/utils/geometry.py` with two batched, read-only helpers: `detection_area` (mask sum / `CompactMask.area` -> shoelace OBB area -> `box_area`) and `detection_iou` (`mask_iou_batch` -> `oriented_box_iou_batch` -> `box_iou_batch`, by the richest geometry both operands carry).
- Route `Detections.area` in `core.py` through `detection_area` so the property is a thin call into the shared helper, with no behavior change and the vectorized design preserved.
- Add `tests/detection/utils/test_geometry.py`: area-parity across mask/OBB/AABB plus the metamorphic contract (identical envelopes with different OBB corners must differ; rotating an OBB off-axis leaves `detection_area` invariant while the envelope area changes).

## Testing

Added `tests/detection/utils/test_geometry.py` covering area parity (mask/OBB/AABB), the metamorphic rotation-invariance contract, empty `Detections`, `CompactMask` routing, and the AABB IoU fallback. The helpers are read-only and preserve `Detections.area` behavior exactly.

Note: I was unable to run the full suite in my local sandbox (the test stack pulls `cv2`/`matplotlib`, which aren't installed in my environment); the new tests are written to run under the repo's CI. Happy to adjust if CI surfaces anything.

## Google Colab (optional)

N/A

## Screenshots/Videos (optional)

N/A (internal refactor; no user-visible output change)

## Additional Notes

The new helpers are intentionally module-private (no public API surface, no per-op mode argument) to match the maintainer's stated preference. Follow-ups can migrate `with_nms`/`with_nmm`/`as_yolo` and consolidate `move_detections` onto the same dispatch.


Borda:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/roboflow/supervision?pullRequest=2374) before we can accept your contribution.

@mvanhorn mind check this out ^^ 🦝

Borda (dismissed):

Borda (approved):

mvanhorn:
Thanks for shepherding the refactor, @Borda - geometry-aware dispatch living in one place beats four copies.

PR #2385, opened 2026-07-02
fix: replace np.cross with explicit 2-D determinant (NumPy 2.0 deprecation)

uttam12331:
## What

Replace two `np.cross()` calls on 2-D vectors with the explicit scalar z-component determinant formula to eliminate the NumPy 2.0 `DeprecationWarning`.

## Why

NumPy 2.0 deprecated `np.cross()` for arrays with size-2 in their last dimension — support will be removed in a future release. With `supervision` allowing `numpy>=1.21.2`, users on the current NumPy line hit this warning on every call that goes through `LineZone.trigger` or `get_polygon_center`.

Affected sites (both on `develop`):
- `src/supervision/geometry/utils.py:44` — `get_polygon_center` signed-area computation
- `src/supervision/detection/utils/internal.py:673` — `cross_product` helper called three times per `LineZone.trigger` call

## Changes

Both `np.cross(a, b)` calls are replaced with the equivalent 2-D determinant:
```python
# before
np.cross(a, b)          # DeprecationWarning on NumPy 2.0+

# after — equivalent, no warning
a[0] * b[1] - a[1] * b[0]   # scalar form
a[:, 0] * b[:, 1] - a[:, 1] * b[:, 0]  # vectorized form
```

## Testing

- Manual verification: zero `DeprecationWarning`s emitted from both call sites on NumPy 2.3.1
- 95 existing tests pass: `tests/detection/test_line_counter.py`, `tests/detection/test_polygonzone.py`, `tests/geometry/`
- Numerical correctness verified against the expected determinant values

### DIFF
--- a/src/supervision/detection/utils/internal.py
+++ b/src/supervision/detection/utils/internal.py
@@ -669,6 +669,10 @@ def cross_product(
         ]
     )
     vector_start = np.array([vector.start.x, vector.start.y])
+    # Explicit 2-D cross product (scalar z-component) to avoid
+    # np.cross DeprecationWarning on NumPy >=2.0 (issue #2384).
+    diff = anchors - vector_start
     return cast(
-        npt.NDArray[np.number], np.cross(vector_at_zero, anchors - vector_start)
+        npt.NDArray[np.number],
+        vector_at_zero[0] * diff[..., 1] - vector_at_zero[1] * diff[..., 0],
     )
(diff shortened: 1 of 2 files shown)


Borda:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/roboflow/supervision?pullRequest=2385) before we can accept your contribution.

@uttam12331 pls need to sign CLA ^^ 🦝

PR #2387, opened 2026-07-02
feat: add supervision AI coding skills for Claude Code and Cursor

### LINKED ISSUE #2244: [Feature]: Add Skill Integration
### Search before asking

- [x] I have searched the issues and discussions and found no similar feature request.

### Feature Description

**What**

Hi team!

Since I've been using daily `supervision` and `inference` libraries, I've created a set of skills for both **Supervision** and **Inference**. You can find them [here](https://github.com/David-rn/cv_skills/tree/main). They still need a bit of refinement but so far so good. 

**Why**

I believe there could be real value in having official skills maintained directly within each library's repository. This would:

- Keep the skills in sync with the latest API changes and releases (since each skill could have the version number of the library)
- Give users a reliable, first-party resource for AI-assisted development

**How**

One idea would be to add a CLI command (e.g. `supervision skills install` or a simple entry point) that copies the relevant skill directly into their current project (`.claude/skills/`). This would make it seamless to integrate AI assistance into any workflow built on top of supervision. That way, the library version installed in the user's environment will match the skill installed, so any change would be tracked.

I'd be happy to contribute to this effort or discuss the approach further. Looking forward to hearing your thoughts!

### Example Usage

_No response_

greeshmab21:
## Summary

Adds an official AI coding skill for `supervision`, so agents like Claude Code and Cursor can use the library correctly instead of guessing at API names. Closes #2244.

The skill is organized as a top-level `SKILL.md` plus focused reference files under `skills/references/`, each targeting a spot where agents commonly hallucinate methods/params or pick the wrong tool.

## What's included

- **`skills/SKILL.md`** — entry point. The standard detect + annotate pattern (model → `sv.Detections` → annotate → save), plus a decision guide for `sv.process_video` vs. `InferencePipeline`: file-in/file-out with a simple callback vs. live/streaming sources with threaded inference. Links out to each reference file.

- **`skills/references/detection.md`** — building `sv.Detections` from ultralytics/inference/SAM/transformers, key attributes (`xyxy`, `confidence`, `class_id`, `tracker_id`, `mask`), and filtering via boolean-mask indexing (`detections[detections.class_id == 0]`) instead of the nonexistent `.filter()`.

- **`skills/references/annotators.md`** — the full annotator class list with correct parameter names, the chaining/compose pattern for layering multiple annotators, and the `BoundingBoxAnnotator` → `BoxAnnotator` naming mistake.

- **`skills/references/tracking.md`** — `sv.ByteTrack` setup, correct constructor param (`track_activation_threshold`, not `confidence_threshold`), correct update method (`update_with_detections`, not `update`), and filtering by `tracker_id`.

- **`skills/references/video.md`** — `sv.process_video`'s `show_progress` param (the `progress=True` typo silently no-ops), `VideoInfo` attributes including `fps` as a float (not int, to avoid timestamp drift), and `VideoSink` for manual frame-by-frame writing.

- **`skills/references/utils.md`** — `PolygonZone.trigger()` for zone membership, `LineZone` counting with `in_count`/`out_count`, and `sv.Color`/`sv.Color.from_hex`/`sv.ColorPalette.DEFAULT` for styling.

## Why

Agents frequently invent plausible-but-wrong supervision API surface (`.filter()`, `BoundingBoxAnnotator`, `tracker.update()`, `progress=True`) or reach for the wrong video-processing tool. This skill packages the correct patterns and the specific mistakes to avoid so agents get it right on the first try.

## Testing

Docs-only change (markdown skill files, no source code touched). No tests affected.


skills/references/tracking.md
Borda, line 1: this was depercated

Borda (changes requested):
I think that shall be covered in llm.txt, and any agent shall be able to learn it from there

greeshmab21:
Thanks for the feedback @Borda! I understand the suggestion about llm.txt — would you be open to a hybrid approach where we keep the structured skill files for Claude Code/Cursor specific integration patterns, while llm.txt covers the general API documentation? Happy to restructure if you have a preferred format in mind.

SkalskiP:
I agree with @Borda. I think the right first step is an automated pipeline that generates and publishes llms.txt, preferably as part of our existing documentation build.

This gives agents a single source derived from our documentation and keeps the content synchronized with API changes.

We also need to be more deliberate. These files will shape how coding agents use and describe Supervision. The initial structure, conventions, and core content should therefore come from the core maintainer team.

For these reasons, I don't think we should merge this PR in its current form. Thanks for exploring the idea and preparing a concrete proposal.

PR #2403, opened 2026-07-04
fix(key_points): guard empty selectors, ndarray indices and degenerate boxes

### LINKED ISSUE #2401: [Bug]:  KeyPoints indexing and as_detections crash or drop detections on valid input
### Search before asking

### Bug

  Three related defects in `KeyPoints` input handling, all on current `develop`.
  Grouping them since they share the same root cause pattern (missing
  empty/type guards) and touch the same class.

 **1. `KeyPoints[[]]` crashes with `IndexError`.**

  ```python
  # src/supervision/key_points/core.py:968
  if isinstance(i, list) and all(isinstance(x, bool) for x in i):
      i = np.array(i)
  ```

  `all(...)` on an empty list is vacuously `True`, so an empty list is treated
  as a boolean mask and converted to `np.array([])` with dtype `float64`, which
  then fails as an index. An empty list is a completely valid selector and it's
  what any "filter matched nothing" produces and `Detections[[]]` already
  handles it correctly, returning an empty `Detections`. `KeyPoints` is the odd
  one out.

  **2. `as_detections(selected_keypoint_indices=...)` crashes on a numpy 
  array.**

  ```python
  # src/supervision/key_points/core.py:1293 and :1316
  if selected_keypoint_indices:
  ```

  The parameter is typed `Iterable[int] | None`, and a numpy array of indices is
  the most natural thing to pass. Truthiness on a multi-element array raises 
  `ValueError: The truth value of an array with more than one element is 
  ambiguous`.

  **3. `as_detections()` silently drops skeletons with collinear or single
  keypoints.**

  ```python
  # src/supervision/key_points/core.py:1325
  detections = detections.select(cast(Any, detections.area) > 0)
  ```

  This line is meant to drop skeletons with no valid keypoints (all `[0, 0]`),
  but `area > 0` also removes any skeleton whose fitted box is degenerate for a
  legitimate reason: a single visible keypoint, or keypoints that happen to lie
  on a horizontal/vertical line. The docstring promises "the bounding box that
  fits all key points" and says nothing about dropping detections in this case.

### Environment

  supervision: develop (bd0f44fc)
  python:      3.12
  os:          macOS

 ```python
  import numpy as np
  import supervision as sv

  # 1. Empty list indexing
  kp = sv.KeyPoints(xy=np.array([[[1, 2], [3, 4]], [[5, 6], [7, 8]]],
  dtype=np.float32))
  kp[[]]   # IndexError: arrays used as indices must be of integer (or boolean)
  type

  # 2. numpy array for selected_keypoint_indices
  kp2 = sv.KeyPoints(xy=np.array([[[10, 10], [20, 20], [30, 15]]],
  dtype=np.float32))
  kp2.as_detections(selected_keypoint_indices=np.array([0, 1]))
  # ValueError: The truth value of an array with more than one element is
  ambiguous

  # 3. Collinear keypoints dropped
  kp3 = sv.KeyPoints(xy=np.array([[[10, 20], [30, 20]]], dtype=np.float32))
  print(len(kp3), len(kp3.as_detections()))   # 1 0 -- the skeleton vanished

  # Same for a single keypoint
  kp4 = sv.KeyPoints(xy=np.array([[[15, 25]]], dtype=np.float32))
  print(len(kp4), len(kp4.as_detections()))   # 1 0
  ```

JESUSROYETH:
Hi 👋

@abhijithneilabraham reported three small defects in `KeyPoints` input handling on `develop` (#2401), and the diagnosis there was already correct. This PR implements the fix for the three of them. They share the same root pattern — a missing empty/type guard — so it made sense to fix them together.

I could reproduce all three on `develop` (Python 3.12) with the snippet from the issue:

```python
import numpy as np
import supervision as sv

kp = sv.KeyPoints(xy=np.array([[[1, 2], [3, 4]], [[5, 6], [7, 8]]], dtype=np.float32))
kp[[]]   # IndexError: arrays used as indices must be of integer (or boolean) type

kp2 = sv.KeyPoints(xy=np.array([[[10, 10], [20, 20], [30, 15]]], dtype=np.float32))
kp2.as_detections(selected_keypoint_indices=np.array([0, 1]))   # ValueError: ambiguous truth value

kp3 = sv.KeyPoints(xy=np.array([[[10, 20], [30, 20]]], dtype=np.float32))
print(len(kp3), len(kp3.as_detections()))   # 1 0 -- the skeleton is gone
```

**Changes** (all in `src/supervision/key_points/core.py`):

1. `__getitem__`: `all(isinstance(x, bool) for x in i)` is vacuously `True` on an empty list, so `kp[[]]` was treated as a boolean mask and turned into a `float64` array, which then fails as an index. Added a non-empty guard so an empty list stays a plain selector and returns an empty `KeyPoints`, the same way `Detections[[]]` already does.

2. `as_detections`: `if selected_keypoint_indices:` raises `ValueError: ambiguous truth value` when you pass a numpy array, which is the most natural type for indices. Changed both checks to `is not None`.

3. `as_detections`: the final filter used `detections.area > 0`, in order to drop skeletons with no valid keypoints, but that also drops legitimate skeletons whose box is degenerate for a real reason — a single visible keypoint, or keypoints on a straight line. `has_valid` already computes the right intent a few lines above, so I filter by that instead. Skeletons with all `[0, 0]` keypoints are still dropped.

**Validation:**

- The three cases above now return an empty `KeyPoints`, a valid `Detections`, and keep the degenerate skeleton.
- The all-`[0, 0]` case is still dropped, and boolean-mask / int / list indexing keep working.
- `pytest tests/ -k key` passes (169 tests, no regressions).

Fixes #2401. Credit to @abhijithneilabraham for the diagnosis and the minimal repros ..

### DIFF
--- a/src/supervision/key_points/core.py
+++ b/src/supervision/key_points/core.py
@@ -965,9 +965,9 @@ class KeyPoints:
         if isinstance(i, int):
             i = [i]
 
-        if isinstance(i, list) and all(isinstance(x, bool) for x in i):
+        if isinstance(i, list) and i and all(isinstance(x, bool) for x in i):
             i = np.array(i)
-        if isinstance(j, list) and all(isinstance(x, bool) for x in j):
+        if isinstance(j, list) and j and all(isinstance(x, bool) for x in j):
             j = np.array(j)
 
         if isinstance(i, np.ndarray) and i.dtype == bool:
@@ -1290,7 +1290,7 @@ class KeyPoints:
             return Detections.empty()
 
         xy = self.xy
-        if selected_keypoint_indices:
+        if selected_keypoint_indices is not None:
             indices = np.asarray(list(selected_keypoint_indices), dtype=np.intp)
             xy = xy[:, indices, :]
 
@@ -1313,7 +1313,7 @@ class KeyPoints:
             confidence = self.detection_confidence.astype(np.float32)
         elif self.keypoint_confidence is not None:
             keypoint_confidence = self.keypoint_confidence
-            if selected_keypoint_indices:
+            if selected_keypoint_indices is not None:
                 keypoint_confidence = keypoint_confidence[:, indices]
             confidence = keypoint_confidence.mean(axis=1).astype(np.float32)
         else:
@@ -1322,6 +1322,6 @@ class KeyPoints:
         detections = Detections(xyxy=xyxy, confidence=confidence)
         detections.class_id = self.class_id
         detections.data = self.data
-        detections = detections.select(cast(Any, detections.area) > 0)
+        detections = detections.select(has_valid)
 
         return detections


abhijithneilabraham:
This issue is already covered in this PR: https://github.com/roboflow/supervision/pull/2402, if there's additional changes, would you like to make it there?

JESUSROYETH:
Hi @abhijithneilabraham 👋 you are right, #2402 already covers this, and it is more complete than mine — you also guard the empty-indices case (`size == 0`), which I only handled as `is not None`, and you added the tests. So there is nothing useful I can add on top. I will close this one in favour of yours. Thanks for pointing it out ..

PR #2436, opened 2026-07-16
fix(metrics): avoid RuntimeWarning division by zero in F1Score computation

### LINKED ISSUE #2434: [Bug]: Fix RuntimeWarning division by zero in F1Score computation
### Search before asking

### Bug

While running the test suite with default warning flags (pytest -W default), I noticed a RuntimeWarning: invalid value encountered in divide coming from src/supervision/metrics/f1_score.py:509.
Root Cause:
f1_score = np.where(denominator == 0, 0, 2 * true_positives / denominator)
Because np.where evaluates both branches eagerly before masking, division by zero still occurs during intermediate evaluation when denominator contains zeros.

Proposed Fix:
Replace eager division with np.divide using explicit out= and where= masking so division by zero is never calculated in memory. I am submitting a PR to fix this now.

### Environment

- Supervision: 0.30.0.dev0 (latest develop branch)
- Python: 3.11.9
- OS: Windows 11

```
# 1. You can reproduce this instantly on the latest develop branch by running the existing test suite with warnings enabled:
# pytest tests/metrics/test_f1_score.py -k test_medium_bucket_scores_target_matched_small_prediction -W default

# 2. Minimal standalone code reproducing the eager evaluation bug at line 509:
import numpy as np

true_positives = np.array([0.0, 1.0])
denominator = np.array([0.0, 2.0])

# This throws: RuntimeWarning: invalid value encountered in divide
# Because np.where evaluates the division across the entire array BEFORE applying the mask:
f1_score = np.where(denominator == 0, 0, 2 * true_positives / denominator)
```

vikassaini77:
## Description

This PR fixes a `RuntimeWarning: invalid value encountered in divide` that occurs in the `F1Score` computation when there are no true positives, false positives, or false negatives (i.e. the denominator evaluates to zero). 

Instead of using `np.where` which eagerly evaluates the division across all elements (triggering the warning), this PR updates the calculation to use `np.divide` with `out` and `where` parameters. This ensures the division is safely bypassed when the denominator is 0.

## Related Issue

Fixes #2434

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

## How Has This Been Tested?

- Verified locally that the `RuntimeWarning` is suppressed when the denominator is 0.


Borda:
@vikassaini77 thank you for your contribution, and even if it is solid, let me move forward with #2437, which is his first-time contribution 🦝

PR #2463, opened 2026-07-31
fix(dataset): keep `class_id` integral for VOC background images

LuShadowX:
<details>

- [x] Self-reviewed the code
- [x] Added/updated tests
- [x] All tests pass locally

</details>

## Description

`DetectionDataset.from_pascal_voc` raises on any dataset containing a background image — an image whose annotation file has no `object` elements:

```
ValueError: Detection annotation for image '.../background.png' contains
non-integer class_id values with dtype float64.
```

In `detections_from_xml_obj`, `class_id` is built as `np.array([...])` over the per-object index list. When that list is empty NumPy infers `float64`, and `DetectionDataset.__init__`'s integer-dtype validation then rejects the `Detections`. Passing `dtype=int` fixes it; `xyxy` a few lines above already handles the empty case explicitly.

The export side is unaffected — `save_pascal_voc_annotations` already writes these files correctly (covered by `test_zero_detection_image_writes_xml_without_object_elements`), so a dataset saved by supervision cannot be read back by supervision.

## Type of Change

## Motivation and Context

Background images are standard negative samples in detection training. The other four formats round-trip them fine — YOLO, COCO, CreateML and LabelMe all load the same dataset without error — so this is specific to the VOC loader.

Reproduction:

```python
from supervision.dataset.core import DetectionDataset

cv2.imwrite("images/background.png", np.zeros((20, 20, 3), np.uint8))
open("annotations/background.xml", "w").write("<annotation></annotation>")

DetectionDataset.from_pascal_voc("images", "annotations")   # ValueError
```

## Tests

Two tests in `TestLoadPascalVocBackgroundImages`, both failing before the change and passing after: one asserts the loaded `class_id` has an integer dtype, one loads a dataset mixing an annotated and a background image and checks the per-image detection counts.

Full suite on `develop`: 3600 passed, 1 skipped. With this change: 3602 passed, 1 skipped.


Borda:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/roboflow/supervision?pullRequest=2463) before we can accept your contribution.

@LuShadowX mind have a look and sign ^^ 🦝

Borda (approved):

LuShadowX:
happy to help! Thanks for quick review and merge..

PR #2468, opened 2026-08-03
fix(metrics): track prediction-only classes in Recall

### LINKED ISSUE #2467: Recall reports a different class set than Precision and F1Score for the same data
`Precision`, `F1Score` and `Recall` return `matched_classes` and a `*_per_class` array that
look like parallel outputs of the same computation. On identical input they are not, and
nothing raises.

## Reproduction

Ground truth contains only class 0. The model predicts class 0 correctly and also predicts
class 1, which never appears in the ground truth at all.

```python
import numpy as np
import supervision as sv
from supervision.metrics import F1Score, Precision, Recall, MetricTarget, AveragingMethod

targets = sv.Detections(
    xyxy=np.array([[10., 10., 50., 50.]]),
    class_id=np.array([0]),
)
predictions = sv.Detections(
    xyxy=np.array([[10., 10., 50., 50.], [100., 100., 140., 140.]]),
    class_id=np.array([0, 1]),
    confidence=np.array([0.9, 0.9]),
)

for name, cls in (("precision", Precision), ("recall", Recall), ("f1", F1Score)):
    m = cls(metric_target=MetricTarget.BOXES, averaging_method=AveragingMethod.MACRO)
    m.update(predictions, targets)
    r = m.compute()
    print(name, r.matched_classes)
```

Observed on `develop` (supervision 0.29.1):

```
precision  [0 1]      precision_per_class shape (2, 10)
recall     [0]        recall_per_class    shape (1, 10)
f1         [0 1]
```

The practical consequence is that the obvious thing to do with these results is wrong:

```python
for cid, p, r in zip(pres.matched_classes, pres.precision_per_class[:, 0],
                     rec.recall_per_class[:, 0]):
    ...
```

`zip` truncates to one row while precision reported two classes, and the hallucinated class
disappears from the table with no error. If the class ordering had differed rather than the
length, the values would have been attributed to the wrong classes instead.

## Where it comes from

#2331, "count false positives on background images and absent classes", changed
`f1_score.py` and `precision.py` and added a regression test to each. It did not touch
`recall.py`. The line it replaced in `precision.py`:

```python
unique_classes, class_counts = np.unique(true_class_ids, return_counts=True)
```

is still present at `src/supervision/metrics/recall.py:331`.

`tests/metrics/test_recall.py` is also the only metric test file with no coverage for a
class that appears solely in predictions. `test_detection.py`, `test_f1_score.py`,
`test_mean_average_recall.py` and `test_precision.py` all have it.

The two docstrings have diverged accordingly. `PrecisionResult.matched_classes` says "the
class IDs present in either predictions or ground truth ... Classes that appear only in
predictions (no ground-truth instances) are included". `RecallResult.matched_classes` says
"the class IDs of all matched classes", with `recall_per_class` documented as
`(num_target_classes, num_iou_thresholds)`.

## The question, because I do not think this is obvious

Recall for a class with no ground-truth instances is undefined, not zero. Including such a
class would add a 0.0 into the macro mean and lower reported recall, which is defensible for
precision (those predictions are real false positives) and arguably wrong for recall.

So there are two reasonable fixes and they are not equivalent:

1. **Match `Precision`.** `unique_classes` becomes the union, and macro recall changes for
   anyone whose model predicts classes absent from the ground truth. Consistent, but it
   moves a published number.
2. **Keep recall's scoring and fix only the contract.** Leave the macro value alone and make
   the difference explicit, so the per-class arrays are never silently mismatched.

Happy to send the PR with tests for whichever you prefer. I would lean to (2) plus an
explicit note in `RecallResult`, but this is your metric to define and I did not want to
change a number you publish on a guess.

arthi-arumugam-git:
> **Correction, added after merge.** Two claims in the "Why include the class rather than leave it out" section below are wrong, and I have left them in place rather than quietly rewriting them. The `WEIGHTED` zero-support guard came from #2408, not #2331 (which never touched `recall.py`), and the state it protects **is** reachable here via a size bucket holding predictions but no targets, so the guard is load-bearing and must not be removed. #2331 also stated that `Recall` was intentionally left unchanged, so this PR overturned a deliberate decision rather than fixing drift. Full detail in [this comment](https://github.com/roboflow/supervision/pull/2468#issuecomment-5261907990). The merged diff is unaffected.

Closes #2467.

#2331 made `Precision` and `F1Score` include classes that appear only in predictions, and
added regression tests to both. `Recall` was not touched. The line #2331 replaced in
`precision.py` is still present at `recall.py:331`, so the three metrics disagree about
which classes exist for identical input:

```
precision.matched_classes -> [0 1]     precision_per_class (2, 10)
recall.matched_classes    -> [0]       recall_per_class    (1, 10)
f1.matched_classes        -> [0 1]
```

They read as parallel outputs of the same computation, so the obvious thing to do with them
is wrong: `zip` truncates to one row, and the class that was predicted but never present in
the ground truth disappears from the recall view with nothing raised.

## Why include the class rather than leave it out

I opened #2467 as a question rather than a PR, because recall for a class with no
ground-truth instances is undefined rather than obviously zero, and matching `Precision`
moves a published number. Two things settled it:

- **sklearn infers labels from the union of `y_true` and `y_pred`.** For a label present
  only in predictions it reports per-class recall `0.0` and includes it in the macro
  average. `test_precision.py` already cites sklearn consistency as the standard for the
  `WEIGHTED` behaviour, so the same standard applies here.
- **`recall.py` already carries #2331's `WEIGHTED` zero-support guard**, whose comment
  refers to "only false-positive classes". That state cannot arise in `recall.py`, because
  `unique_classes` came from the ground truth alone. The guard was propagated across; the
  union that gives it meaning was not.

## What changes, stated plainly

`MICRO` is unchanged: an absent class contributes no false negatives. `WEIGHTED` is
unchanged: its ground-truth support is zero, so its weight is zero. **`MACRO` does change**
for any evaluation containing predictions of a class with no ground-truth instances, because
that class now contributes `0.0` to the mean. The changelog entry says so explicitly rather
than burying it.

`RecallResult.matched_classes` and the `recall_per_class` shape are documented to match
`PrecisionResult`.

## Tests

Four new cases in `tests/metrics/test_recall.py`, all failing before this change:

- `MICRO`, `MACRO` and `WEIGHTED` against a class predicted but never in the targets,
  parametrized, pinning that only `MACRO` moves.
- `test_tracked_classes_match_precision_and_f1`, which asserts the three metrics agree on
  the class set and the per-class row count for the same data. That is the invariant that
  was actually broken, so it is worth a test of its own.

Whole metrics suite: **280 passed**, up from 276, with no existing test modified.
`ruff format` and `ruff check` clean.


arthi-arumugam-git:
Good catch, and it was a real hole rather than a nitpick. Fixed in 9e74574.

You are right that building the union inside `_compute_recall_for_classes` only covers samples that reach it. Samples with predictions and no targets were skipped earlier in `_compute`, so on one normal sample plus one background image predicting class 2:

```
precision.matched_classes -> [0 2]
recall.matched_classes    -> [0]
```

which is exactly the invariant the test in this PR asserts, so the test was weaker than the claim it was making. `Recall._compute` now handles `len(targets) == 0 and len(predictions) > 0` the way `Precision` does. No recall value moves, since a background image produces no false negatives; only the tracked class set does.

Both suggestions are in: the source fix, and the test extended with a second sample using `Detections.empty()` targets, asserting all three metrics agree on `matched_classes` and per-class row count. It fails without the source change. Metrics suite is at 281 passing, and the changelog now mentions background images explicitly.

Borda (approved):

arthi-arumugam-git:
Correcting two things I got wrong in the description above. The merged code is unaffected, but the reasoning is wrong in a way that could mislead someone later, and the practical upshot is the opposite of what I implied.

**1. The `WEIGHTED` zero-support guard did not come from #2331.**

I wrote that `recall.py` "already carries #2331's `WEIGHTED` zero-support guard". #2331 only touched `f1_score.py` and `precision.py`. It never modified `recall.py`. The guard was added three weeks later, by #2408, in `072f784`. It is absent at `beb0470` and `4624dde` and present from `072f784` onward.

**2. The state it protects is reachable here.**

I wrote that it "cannot arise in `recall.py`, because `unique_classes` came from the ground truth alone". The guard's own comment names the case, in the clause I did not quote:

```python
if class_counts.sum() == 0:
    # No ground-truth support (e.g. only false-positive classes, or a
    # size bucket with predictions but no targets): weighting is
    # undefined, so report 0 as the empty case did before.
```

A size bucket holding predictions but no targets reaches `_compute_recall_for_classes` with `class_counts.sum() == 0` on `develop` today, independently of this PR. Without the guard, `np.average(..., weights=class_counts)` raises `ZeroDivisionError`. So it is load-bearing rather than propagated boilerplate, and it should not be removed on the strength of what I wrote.

**3. One more thing worth recording.**

#2331 stated explicitly that "`Recall` is intentionally left unchanged", with its reasoning. I framed the difference between the three metrics as drift. It was a deliberate decision, and this PR overturned it. I still think overturning it is right, for the sklearn-consistency reason and because the class sets genuinely could not be compared row for row, but it should have been argued as a reversal rather than as fixing an oversight.

Nothing here affects the diff that merged: the union, the tests, and the changelog entry all stand. Apologies for the noise on a closed PR. I would rather the record be right.

PR #2470, opened 2026-08-03
test(metrics): verify live sklearn parity

Daniiiil1:
## Summary

- compare `sv.Precision`, `sv.Recall`, and `sv.F1Score` with live scikit-learn results during every test run
- cover balanced labels, target-only labels, prediction-only labels, non-contiguous IDs, background images, target-only batches, prediction-only batches, and fully empty input
- verify per-class ordering and values plus `MICRO`, `MACRO`, and `WEIGHTED` aggregates across all ten IoU thresholds
- add scikit-learn to the development dependency group so local and CI environments use the same independent reference implementation

## Why

The metrics tests cite sklearn behavior, but no test currently enforces parity. That leaves class-union and zero-support behavior vulnerable to silent drift, especially for prediction-only classes and background images.

The test data maps unmatched targets and predictions to an excluded background sentinel in the sklearn event stream. This preserves the foreground TP/FP/FN counts used by Supervision while keeping object-detection inputs realistic.

Following review feedback, the expected scores are computed directly with `precision_score`, `recall_score`, and `f1_score` instead of being stored as frozen constants. New scenarios and averaging modes therefore need only test inputs, not manually regenerated outputs.

## Relationship to #2468

#2468 has merged. This branch is based on the current `develop` branch; the Recall implementation and its regression tests come from `develop`.

## Validation

- `uv run pytest tests/metrics/test_sklearn_parity.py tests/metrics/test_recall.py -q` — 132 passed
- `uv run pytest -q --disable-warnings` — 3,561 passed, 20 skipped
- `uv run pre-commit run --all-files` — all hooks passed, including Ruff, MyPy, formatters, pyproject validation, and codespell

### DIFF
--- a/tests/metrics/test_sklearn_parity.py
+++ b/tests/metrics/test_sklearn_parity.py
@@ -1,0 +1,409 @@
+"""Golden parity tests for classification-style detection outcomes.
+
+The constants in this module were generated with scikit-learn 1.7.1. Unmatched
+targets and predictions are represented by ``BACKGROUND_CLASS_ID`` in the sklearn
+event stream, while ``labels`` contains only foreground classes. This produces the
+same per-class TP, FP, and FN counts as Supervision's detection metrics.
+
+To regenerate the constants locally without adding sklearn to the test environment::
+
+    from sklearn.metrics import f1_score, precision_score, recall_score
+
+    functions = {
+        "precision": precision_score,
+        "recall": recall_score,
+        "f1": f1_score,
+    }
+    for case in PARITY_CASES:
+        for metric_name, function in functions.items():
+            print(case.name, metric_name)
+            print(
+                function(
+                    case.sklearn_y_true,
+                    case.sklearn_y_pred,
+                    labels=case.sklearn_labels,
+                    average=None,
+                    zero_division=0,
+                )
+            )
+            for average in ("micro", "macro", "weighted"):
+                print(
+                    average,
+                    function(
+                        case.sklearn_y_true,
+                        case.sklearn_y_pred,
+                        labels=case.sklearn_labels,
+                        average=average,
+                        zero_division=0,
+                    ),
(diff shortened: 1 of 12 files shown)


tests/metrics/test_sklearn_parity.py
```diff
@@ -0,0 +1,409 @@
+"""Golden parity tests for classification-style detection outcomes.
+
+The constants in this module were generated with scikit-learn 1.7.1. Unmatched
+targets and predictions are represented by ``BACKGROUND_CLASS_ID`` in the sklearn
+event stream, while ``labels`` contains only foreground classes. This produces the
+same per-class TP, FP, and FN counts as Supervision's detection metrics.
```
Borda: I would prefer it if the sklearn outcome were computed as part of the test, then claim frozen outputs, which make it not very flexible for adding extra metrics or input cases

Daniiiil1: Thanks — updated. The parity tests now compute the reference values directly with scikit-learn’s `precision_score`, `recall_score`, and `f1_score` for every case and averaging mode. I removed the frozen outputs and added scikit-learn as a development dependency. Validation: 132 focused tests and the complete suite of 3,561 tests pass; all pre-commit checks pass as well.

Daniiiil1:
The CI failures came from scikit-learn 1.9 rejecting completely empty y_true/y_pred inputs, while 1.7.2 still accepted them. I separated that edge case from the live sklearn parity matrix and now test the Supervision empty-input contract directly. The non-empty cases still compute their reference values with sklearn at runtime.

Validation:
- Python 3.13 + scikit-learn 1.9.0: 96 focused tests passed
- full suite: 3,561 passed, 20 skipped
- pre-commit: all hooks passed

Borda (approved):

PR #2478, opened 2026-08-08
harden: disable external XML entity processing in pascal_voc.py

anupamme:
## Summary
Harden input handling in `src/supervision/dataset/formats/pascal_voc.py` (flagged by semgrep).

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | gitlab.bandit.B313.B314.B315.B316.B318.B319.B320.B405.B406.B407.B408.B409.B410 |
| **Severity** | HIGH |
| **Scanner** | semgrep |
| **Rule** | `gitlab.bandit.B313.B314.B315.B316.B318.B319.B320.B405.B406.B407.B408.B409.B410` |
| **File** | `src/supervision/dataset/formats/pascal_voc.py:4` |
| **Assessment** | Defensive hardening |

**Description**: Found use of the native Python XML libraries, which is vulnerable to XML external entity (XXE)
attacks. The Python documentation recommends the 'defusedxml' library instead. Use 'defusedxml'.
See https://github.com/tiran/defusedxml for more information.

## Threat Model Context

This is a Python library - vulnerabilities affect applications that import this code.

## Changes
- `src/supervision/dataset/formats/pascal_voc.py`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path; it only tightens handling of untrusted input and leaves valid inputs unaffected.

---
*This patch removes an exploit primitive — a code pattern that, while not independently exploitable today, could be chained with other weaknesses by automated exploit-development tooling. Proactive removal of such primitives raises the bar against increasingly capable automated attack tools.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*


Borda (approved):

Borda:
Seems this is not vital 😕

anupamme:
Thanks for taking a look. I agree that this is more defence-in-depth than a directly exploitable vulnerability in Supervision.

The reason I submitted it is that the Pascal VOC loader parses XML dataset files that can originate from third parties. The stdlib XML parser exposes an unnecessary XXE/entity-expansion primitive. defusedxml is already a runtime dependency, so the mitigation is a very small, low-risk change with no new dependency cost.

I don’t want to overstate the severity, though. If the project doesn’t consider malicious dataset XML within scope, I’m happy to treat this as hardening rather than a security fix.

Borda:
@anupamme I would be fine with it, but your PR didn't pass CI checks so if you resolve the problem and make it vital for supervison, I am happy to move forward with it 🦝

PR #2484, opened 2026-08-10
docs: convert fenced examples to doctests in annotators/core.py

ayo0la:
<details>

- [x] Self-reviewed the code
- [x] All tests pass locally

</details>

## Description

Converts 6 of the 7 remaining fenced ` ```python ` examples in `annotators/core.py` to `pycon` doctest format so they run under `pytest --doctest-modules`, following the precedent of #2474 and #2475 (same conversion for `dataset/formats/coco.py` and `dataset/formats/createml.py`).

`IconAnnotator.annotate`'s example is intentionally left as a fenced block — it loads a real PNG via `icon_path`, which per [CONTRIBUTING.md's doctest guidance](https://github.com/roboflow/supervision/blob/develop/.github/CONTRIBUTING.md) is not appropriate for a doctest (no filesystem access outside `supervision/assets/`).

## Type of Change

- 📝 Documentation update

## Motivation and Context

Fenced examples render in the docs but are never executed, so they silently rot when the API changes. A `pycon` doctest is checked on every test run. Several of the converted examples previously relied on `ultralytics.YOLO` model inference or a real video/image path, which is neither deterministic nor testable in CI — those are rewritten to build synthetic `Detections` and a blank `np.zeros` scene instead, matching the pattern already used by `BoxAnnotator.annotate`'s existing doctest in the same file.

## Changes Made

- `OrientedBoxAnnotator.annotate`: fenced example → doctest, synthetic `Detections` with `ORIENTED_BOX_COORDINATES` data instead of a real YOLO-OBB model
- `LabelAnnotator.draw_rounded_rectangle`: fenced example → doctest, asserts the filled pixel color
- `TraceAnnotator.annotate`: fenced example → doctest, two synthetic detections over the same `tracker_id` in place of a real video/model loop
- `HeatMapAnnotator.annotate`: fenced example → doctest, asserts `heat_mask` accumulated
- `PixelateAnnotator.annotate`: fenced example → doctest, synthetic scene/detections
- `ComparisonAnnotator.annotate`: fenced example → doctest, two synthetic detection sets
- Normalised `Examples:` → `Example:` for Google-style consistency where each docstring now has exactly one example

## Testing

`uv run pytest --doctest-modules src/supervision/annotators/core.py` — 25 passed (up from 19 baseline; +6 doctests, no regressions), verified against both the pure-NumPy `_cv2` fallback and real `opencv-python-headless`.
`uv run pytest tests/annotators/` — 247 passed.
`pre-commit run --files src/supervision/annotators/core.py` — all hooks pass, including `check doctest fences`.

### DIFF
--- a/src/supervision/annotators/core.py
+++ b/src/supervision/annotators/core.py
@@ -328,7 +328,7 @@ class OrientedBoxAnnotator(BaseAnnotator):
             The annotated image, matching the type of `scene` (`numpy.ndarray`
                 or `PIL.Image.Image`)
 
-        Example:
+        Examples:
             ```python
             from supervision import _cv2 as cv2
             import supervision as sv
@@ -345,6 +345,25 @@ class OrientedBoxAnnotator(BaseAnnotator):
                 scene=image.copy(),
                 detections=detections
             )
+            ```
+
+            ```pycon
+            >>> import numpy as np
+            >>> import supervision as sv
+            >>> from supervision.config import ORIENTED_BOX_COORDINATES
+            >>> image = np.zeros((100, 100, 3), dtype=np.uint8)
+            >>> obb = np.array([[[20, 20], [80, 20], [80, 80], [20, 80]]])
+            >>> detections = sv.Detections(
+            ...     xyxy=np.array([[20, 20, 80, 80]]),
+            ...     class_id=np.array([0]),
+            ...     data={ORIENTED_BOX_COORDINATES: obb},
+            ... )
+            >>> oriented_box_annotator = sv.OrientedBoxAnnotator()
+            >>> annotated_frame = oriented_box_annotator.annotate(
+            ...     scene=image.copy(),
+            ...     detections=detections
+            ... )
+
             ```
         """
         if not isinstance(scene, np.ndarray):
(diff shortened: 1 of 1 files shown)


src/supervision/annotators/core.py
```diff
@@ -329,22 +329,23 @@ def annotate(
...
+            >>> annotated_frame = oriented_box_annotator.annotate(
+            ...     scene=image.copy(),
+            ...     detections=detections
+            ... )
 
-            oriented_box_annotator = sv.OrientedBoxAnnotator()
-            annotated_frame = oriented_box_annotator.annotate(
-                scene=image.copy(),
-                detections=detections
-            )
```
Borda, line 347: I think in these cases we shall keep both; the new doctest is correct on simple testing, but the original code causes the integration  story so they are not easily inter-replaceable

src/supervision/annotators/core.py
```diff
@@ -3338,19 +3344,25 @@ def annotate(
             The annotated image.
 
         Example:
-            ```python
-            import supervision as sv
-
-            image = ...
-            detections_1 = sv.Detections(...)
-            detections_2 = sv.Detections(...)
+            ```pycon
```
Borda, line 3406: here it is fine :)

Borda (commented):
for the two cases where we showcase `model = YOLO('yolov8x.pt')` we shall hepp the integration demo

src/supervision/annotators/core.py
ayo0la: Agreed, done. Kept the integration example and added the doctest beneath it under `Examples:` for OrientedBox, Trace and HeatMap. Rebased onto develop, so Trace and HeatMap now carry the RF-DETR version of the integration example from #2494. Comparison, Pixelate and draw_rounded_rectangle stay doctest-only.

Borda (approved):

PR #2499, opened 2026-08-22
test(detection): add the geometry contract test

adhavan18:
<details>

- [x] Self-reviewed the code
- [x] Added/updated tests
- [x] All tests pass locally

</details>

## Description

Adds the `test_op_respects_geometry` contract test that #2318 lists as its third rollout step, after the `detection_area` / `detection_iou` dispatch that landed in #2374 and #2382.

The contract is stated once: **a geometry-aware op must not give the same answer for an oriented box as for its axis-aligned envelope.** A new op joins by adding one entry to `GEOMETRY_AWARE_OPS`; an op that silently falls back to `xyxy` fails here rather than in a bug report months later.

## Type of Change

- 🧪 Test update

## Motivation and Context

#2318 describes the pattern rather than a single bug: `Detections` carries several geometries at once, `xyxy` is always present, so forgetting the real one fails silently and looks like a fresh bug each time.

- `Detections.area` returned the envelope area, not the rotated body (#2306)
- `with_nms` / `with_nmm` dropped crossed oriented boxes via envelope IoU (#2303)
- `as_yolo` dropped rotation on export (#2289)
- `DetectionsSmoother` averaged `xyxy` and carried the corners over unsmoothed from the oldest frame in the window (#2489)

Each was found by someone hitting it in practice. The contract test is the part of #2318 that makes the next one fail in CI instead.

Refs #2318.

## Changes Made

- Adds `tests/detection/test_geometry_contract.py`.
- `GEOMETRY_AWARE_OPS` is a registry of ops, each run twice: once against a detection carrying oriented corners, once against the same detection with the corners stripped. The contract is that the two disagree.
- Exact-value tests alongside the contract, so a failure says which geometry was used rather than only that a number moved.
- A guard on the fixture itself, so a broken fixture cannot quietly weaken the contract.

The fixture is a square rotated 45 degrees. Its envelope has exactly twice its area, which keeps the expectations exact:

```
area(oriented)  = 2 * d^2      area(envelope)  = 4 * d^2
```

For IoU, two diamonds offset along the diagonal meet only along a shared edge while their envelopes overlap over `d * d` with a union of `7 * d * d`:

```
iou(oriented)   = 0            iou(envelope)   = 1/7
```

## Testing

A test that never fails is worth nothing, so I checked it catches the bug class by reintroducing two of the historical occurrences.

Reverting `detection_area` to the envelope, i.e. #2306:

```text
FAILED test_op_respects_geometry[detection_area]
E   AssertionError: detection_area gave the same answer for an oriented box and its
    envelope ([400.0]), so it is reading xyxy rather than the geometry the detection
    actually carries
FAILED TestGeometryContractValues::test_area_uses_the_rotated_body
2 failed, 3 passed
```

Reverting `detection_iou` to box IoU, i.e. #2303:

```text
FAILED test_op_respects_geometry[detection_iou]
FAILED TestGeometryContractValues::test_iou_uses_the_rotated_body
2 failed, 3 passed
```

Both restored afterwards; the diff contains no source change.

```text
tests/detection/                     1624 passed, 1 skipped
tests/detection/test_geometry_contract.py   5 passed
```

`ruff check` and `ruff format --check` clean.

## Additional Notes

Two things I deliberately left out, both worth your call rather than mine:

- **Scope.** Only the two ops that currently go through `_geometry_dispatch` are registered. `as_coco` / `as_pascal_voc` and the annotators also read geometry, but adding them here would assert behaviour they do not have yet, which turns a contract test into a failing wishlist. They are one line each once the dispatch reaches them.
- **`move_detections`.** #2318 lists consolidating it as step 1, and I could not find a `move_detections` in the tree to consolidate, so I have not assumed what its shape should be.

I came to this from #2489, which was the `DetectionsSmoother` instance of the same bug: the corners were carried over unsmoothed while `xyxy` was averaged. That one was found by reading the code rather than by a test, which is the gap this closes.

### DIFF
--- a/tests/detection/test_geometry_contract.py
+++ b/tests/detection/test_geometry_contract.py
@@ -1,0 +1,159 @@
+"""Contract tests for geometry-aware detection ops (issue #2318).
+
+Several fixes have been the same bug in different functions: an op reaches for
+`xyxy` and forgets the real geometry lives in `data["xyxyxyxy"]`. Because `xyxy`
+is always present the wrong answer is the silent default, so each occurrence
+looked like a fresh bug rather than a recurring one:
+
+* `Detections.area` returned the envelope area, not the rotated body (#2306)
+* `with_nms` / `with_nmm` dropped crossed oriented boxes via envelope IoU (#2303)
+* `as_yolo` dropped oriented-box rotation on export (#2289)
+* `DetectionsSmoother` averaged `xyxy` and carried the corners over unsmoothed
+  from the oldest frame in the window (#2489)
+
+These tests state the shared contract once: **an op that claims to be
+geometry-aware must produce a different answer for an oriented box than for its
+axis-aligned envelope.** A new op is covered by adding one entry to
+`GEOMETRY_AWARE_OPS`, and an op that silently falls back to `xyxy` fails here
+rather than in a bug report months later.
+
+The fixture is a square rotated 45 degrees, whose envelope has exactly twice its
+area. That makes the expected values exact rather than approximate, so a failure
+says which geometry was used rather than only that a number moved.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+
+import numpy as np
+import pytest
+from numpy.testing import assert_allclose
+
+from supervision.config import ORIENTED_BOX_COORDINATES
+from supervision.detection._geometry_dispatch import detection_area, detection_iou
+from supervision.detection.core import Detections
+
+#: Half-diagonal of the rotated square used throughout.
+HALF_DIAGONAL = 10.0
+
+
+def diamond(
+    cx: float = 0.0, cy: float = 0.0, half_diagonal: float = HALF_DIAGONAL
+) -> np.ndarray:
+    """Corners of a square rotated 45 degrees about ``(cx, cy)``.
+
+    Its side is ``half_diagonal * sqrt(2)``, so its area is ``2 * half_diagonal ** 2``,
+    exactly half the area of its axis-aligned envelope.
+    """
+    return np.array(
+        [
+            [
+                [cx, cy - half_diagonal],
+                [cx + half_diagonal, cy],
+                [cx, cy + half_diagonal],
+                [cx - half_diagonal, cy],
+            ]
+        ],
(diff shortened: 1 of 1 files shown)


tests/detection/test_geometry_contract.py
```diff
@@ -0,0 +1,159 @@
...
+                [cx + half_diagonal, cy],
+                [cx, cy + half_diagonal],
+                [cx - half_diagonal, cy],
+            ]
+        ],
+        dtype=np.float32,
+    )
+
+
+def envelope_of(corners: np.ndarray) -> np.ndarray:
```
Borda: make this and onther helpers as private

tests/detection/test_geometry_contract.py
```diff
@@ -0,0 +1,159 @@
...
+#: contract is that the two disagree; a fallback to `xyxy` makes them equal.
+GEOMETRY_AWARE_OPS: list[
+    tuple[str, Callable[[Callable[..., Detections]], np.ndarray]]
+] = [
+    ("detection_area", lambda build: detection_area(build())),
+    (
+        "detection_iou",
+        lambda build: detection_iou(build(), build(HALF_DIAGONAL, HALF_DIAGONAL)),
+    ),
+]
```
Borda: This seems to be rather pytest fixture

adhavan18:
Thanks @Borda, both done in `960f999`.

- The four fixture builders (`diamond`, `envelope_of`, `oriented`, `envelope_only`) and the ops registry are now underscore-private, since they are module-internal and not meant to be imported.
- On the "rather pytest fixture" note: I kept `_GEOMETRY_AWARE_OPS` as a plain module-level list feeding `@pytest.mark.parametrize` rather than converting it to a fixture. A fixture is resolved per-test and cannot be passed to `parametrize`, which needs its values at collection time to generate one test case per op with a readable id (`test_op_respects_geometry[detection_area]`, `[detection_iou]`). Happy to switch to a fixture-based shape if you had a specific one in mind, but I did not want to lose the per-op parametrization.

5 tests pass, `ruff check` / `ruff format --check` clean.

Borda (approved):

PR #2514, opened 2026-08-25
fix(detection): prevent integer overflow in `box_area`

AdvaitS:
## Description

`Detections.box_area` computed `(y2-y1) * (x2-x1)` in the input dtype, so integer-coordinate boxes silently wrapped to a negative area. This also affects `Detections.area` for axis-aligned boxes (the AABB fallback branch).

```python
import numpy as np
import supervision as sv

d = sv.Detections(xyxy=np.array([[0, 0, 50000, 50000]], dtype=np.int32))
d.box_area  # -> array([-1794967296], dtype=int32)   # expected 2.5e9
```

This is the same class of bug fixed for `box_iou` in #2485, but `box_area` was missed.

## Changes

- `box_area` now promotes integer `xyxy` to `float64` before multiplying, so the product cannot overflow. Floating inputs are untouched and keep their dtype (preserving the documented AABB "input dtype" contract).
- Added `test_box_area_does_not_overflow_integer_dtypes` covering `int16`/`int32`/`uint16`/`uint32`.
- Added a changelog entry.

## Testing

- Full suite: `3636 passed, 20 skipped`.
- Ruff check + format clean.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -13,6 +13,7 @@ date_modified: 2026-08-11
 
 ### Fixed
 
+- `sv.Detections.box_area` (and therefore `sv.Detections.area` for axis-aligned boxes) now computes integer-coordinate box areas in `float64`, preventing integer overflow for large boxes (e.g. an `int32` `50000 x 50000` box previously wrapped to a negative area).
 - Docs deployment for `latest` no longer fails with `error: version 'latest' already exists` when `latest` exists as an alias of a released version; the publish workflow now always deletes `latest` before redeploying it ([#2512](https://github.com/roboflow/supervision/issues/2512)).
 
 ### 0.30.1 <small>Aug 24, 2026</small>
--- a/src/supervision/detection/core.py
+++ b/src/supervision/detection/core.py
@@ -2834,8 +2834,8 @@ class Detections:
         ``with_nms``, ``with_nmm``, and this property — always store OBB corners
         under ``config.ORIENTED_BOX_COORDINATES`` with that shape.
 
-        **Return dtype**: ``float64`` (OBB branch), input dtype (AABB fallback),
-        ``int64`` (mask branch).
+        **Return dtype**: ``float64`` (OBB branch and integer AABB fallback),
+        floating input dtype (floating AABB fallback), ``int64`` (mask branch).
 
         Returns:
             An array containing the area of each detection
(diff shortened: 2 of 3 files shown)


fallenmi (changes requested):
## Review Summary

**Recommendation:** 🟠 Request Changes — the overflow fix introduces a precision regression for valid `int64` and `uint64` boxes with large origins.

**PR Completeness:**
- ✅ Complete: clear reproduction, changelog entry, tests for 16/32-bit integer product overflow, and a green upstream matrix.
- ❌ Missing: 64-bit integer cases where the coordinates themselves must remain exact during subtraction.

**Quality Scores:**
- Code: 3/5 🟡 — the intended overflow fix works for the covered dtypes, but casting both corners before subtraction loses narrow widths at large origins.
- Testing: 3/5 🟡 — good 16/32-bit coverage, but `int64`/`uint64` are omitted.
- Documentation: 5/5 🟢 — the user-visible change is clearly documented.

**Risk Level:** 3/5 🟡 — wrong zero areas can propagate through the public `Detections.area` API.

**Critical Issue (Must Fix):**

`xyxy.astype(np.float64)` runs before the coordinate differences are calculated. Above `2**53`, adjacent 64-bit integer coordinates can round to the same float, so a non-degenerate one-unit-wide box becomes zero-width.

I reproduced this through the public API on exact base `0753191a778f2582b66f3fadb4cd254ca3d1fae1`, exact head `2cdb8b8745ae79933acad6c60551525e577c114d`, and current GitHub merge `d91f65747364e6314831d44d0fcb9f673e2de585`. For each of `int64` and `uint64`, boxes `[2**53, 0, 2**53 + 1, 1]` and adjacent coordinates near the dtype maximum return area `1` on base but `0.0` on head/current merge. `sv.box_iou(box, box)` still returns `1.0` for those exact same head inputs because its existing `_coordinate_difference` path subtracts integer values before converting to float.

The independent eight-case matrix demonstrates the tradeoff directly: base passes all four large-origin precision cases but fails four 16/32-bit overflow cases; head/current merge fix those four overflow cases but fail all four large-origin cases. The focused upstream `TestDetectionsArea` slice passes (18 base; 22 head/current merge), so the new regression is not covered by the submitted tests.

**Next Steps:**
1. Preserve exact integer coordinate differences before converting the widths/heights for overflow-safe multiplication, following the large-origin invariant already used by `box_iou`.
2. Add `int64` and `uint64` adjacent-coordinate tests above `2**53`, including coverage through both `box_area` and `area`.

Disclosure: I used OpenAI Codex to assist this review; I independently verified the exact refs, current merge, public-API behavior, focused tests, and live CI state described above.

Borda (approved):

PR #2521, opened 2026-08-31
feat(detection): expose `match_detections` public matching primitive

### LINKED ISSUE #2476: match_detections(): expose the unified greedy matcher as a public cross-Detections primitive
## Search before asking

- [x] I have searched the supervision issues and found no similar feature request.

## Description

#2378 pointed out that `ConfusionMatrix` and `Recall`/`F1Score`/`MeanAverageRecall` were matching predictions to targets with different algorithms, and #2380 fixed it the same day by unifying everything onto one greedy matcher in `supervision/metrics/utils/matching.py` (`_greedy_match` + `_match_detection_batch_with_target_indices`). All five metric modules now delegate to it.

That matcher is exactly the primitive users keep needing outside metrics — "given two `Detections`, which instances correspond?" — but it's private and shaped for metric internals (pre-computed IoU matrix, class-index arrays, per-IoU-threshold correctness matrix, COCO-style scored/unscored rounds). There is no public API that takes two `Detections` and returns matched pairs. `box_iou_batch` is public, so today everyone hand-rolls the assignment step on top of it — with the same one-to-many/ordering bugs #2378 was about — or sidesteps geometry entirely with VLM calls (e.g. the planogram-compliance guide pairs an RF-DETR detector with two Gemini VQA calls, where the "which detection corresponds to which reference slot" half of the problem is a matching problem: https://blog.roboflow.com/planogram-compliance-detection/).

## Proposed API

```python
matched_pairs, unmatched_a, unmatched_b = sv.match_detections(
    detections_a,           # sv.Detections
    detections_b,           # sv.Detections
    iou_threshold=0.5,
    class_agnostic=False,
)
```

Returns index arrays, not new `Detections`, so it composes with existing slicing:

- `matched_pairs`: `np.ndarray` of shape `(M, 2)` — column 0 indexes into `detections_a`, column 1 into `detections_b`
- `unmatched_a`: `(len(detections_a) - M,)` indices into `detections_a`
- `unmatched_b`: `(len(detections_b) - M,)` indices into `detections_b`

`detections_a[unmatched_a]`, `detections_b[matched_pairs[:, 1]]`, etc. all just work.

### Semantics

| Case | Behavior |
|---|---|
| Assignment | one-to-one, greedy, highest-IoU-first — identical to `_greedy_match` from #2380 |
| IoU ties | stable order (`np.argsort(-iou, kind="stable")`), matching the internals |
| `class_agnostic=False` (default) | a pair additionally requires equal `class_id` — mirrors the `correct_class` mask in `_match_detection_batch_with_target_indices` |
| `class_agnostic=True` | IoU-only matching |
| IoU below `iou_threshold` | never paired, regardless of class |
| `detections_a` or `detections_b` empty | `matched_pairs.shape == (0, 2)`; every index of the other side is unmatched |
| `confidence` | ignored — the primitive is pure geometry+class; callers who want metric-style score ordering sort upstream |
| Symmetry | greedy-on-IoU makes `match_detections(a, b)` equal `match_detections(b, a)` with columns swapped (up to ordering among exactly-tied IoU pairs) |

Single scalar `iou_threshold` only — the per-threshold sweep and the scored/unscored two-round COCO behavior stay in the metrics layer where they belong.

## Use cases

1. **Eval tooling beyond aggregate numbers.** The metrics give mAP/F1; users routinely want the *instances* — per-image TP/FP/FN lists to render with annotators, mine hard examples, or debug a confusion pair. Today that means reimplementing the matcher the library already contains.
2. **Before/after change detection.** My production use: a litter-mapping app that photo-verifies cleanups by detecting on a before and an after photo of the same scene and diffing — unmatched-in-before = removed, unmatched-in-after = new debris. The matching step currently lives outside supervision because there's nothing to call.
3. **Tracker bootstrap / cross-source association.** `ByteTrack` is sequential and stateful; a stateless pairwise matcher covers the two-frame case, detector-A-vs-detector-B comparison, and re-association after a tracker reset.
4. **Annotation QA.** Agreement between two annotators, or human labels vs model suggestions, is the same primitive with ground truth on both sides.

Demand honesty: I found only a couple of forum threads asking for this, so I'm pitching it as a missing primitive the library already contains privately — not as a heavily-requested feature.

## Implementation offer

This is a thin wrapper: `box_iou_batch` → candidate mask (`iou >= threshold`, optionally `class_id` equality) → reuse `_greedy_match` → collect indices. No behavior change to any metric; ~40 lines plus tests and docs. I'm happy to submit the PR — after 0.30.0 ships, so this adds no review load to the release. Also fine with a different spelling if maintainers prefer (`Detections.match(other)` method, or a home in `supervision.detection.utils` next to `box_iou_batch`).

alvaromoureupm:
<details>

- [x] Self-reviewed the code
- [x] Added/updated tests
- [x] All tests pass locally

</details>

## Description

Exposes the unified greedy highest-IoU-first matcher (introduced internally in #2380) as a public primitive function `sv.match_detections`, as requested in #2476.

```python
matched_pairs, unmatched_a, unmatched_b = sv.match_detections(
    detections_a,           # sv.Detections
    detections_b,           # sv.Detections
    iou_threshold=0.5,
    class_agnostic=False,
)
```

- Returns `matched_pairs` of shape `(M, 2)` where column 0 indexes into `detections_a` and column 1 into `detections_b`.
- Returns 1D index arrays `unmatched_a` and `unmatched_b` (computed with vectorized `np.setdiff1d`).
- Respects equal `class_id` matching by default unless `class_agnostic=True`.
- Reuses the internal `_greedy_match` helper so public matching behavior stays consistent with metrics (`MeanAveragePrecision`, `F1Score`, `ConfusionMatrix`).

## Type of Change

## Motivation and Context

Users currently re-implement IoU-based detection matching (ground-truth vs. prediction pairing, cross-model comparison, deduplication) even though a battle-tested matcher already exists internally. This PR makes it a first-class public primitive.

Closes #2476

## Changes Made

- Added `match_detections()` in `src/supervision/detection/utils/iou_and_nms.py` with Google-style docstring and executable `pycon` example.
- Exported `match_detections` in `supervision.__init__` / `__all__`.
- Added docs autogeneration entry in `docs/detection/utils/iou_and_nms.md`.
- Added changelog entry in `docs/changelog.md`.
- Added `TestMatchDetections` unit tests in `tests/detection/utils/test_iou_and_nms.py`.

## Testing

`uv run pytest tests/detection/utils/test_iou_and_nms.py tests/test_public_api.py` → 371 passed. `ruff check`, `ruff format --check`, and `mypy` on the touched files pass (only pre-existing warnings remain).

## Additional Notes

The `_greedy_match` import is placed inside the function body to avoid a circular import between `supervision.detection` and `supervision.metrics`.


Borda:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/roboflow/supervision?pullRequest=2521) before we can accept your contribution.

@alvaromoureupm ^^ 🙏

adawgwats:
@alvaromoureupm thanks for picking this up — I opened #2476, said I'd send a PR, and then didn't, so the initiative is appreciated. No hard feelings needed in either direction; this should stay your PR.

I reviewed the implementation and sent the changes to your branch rather than opening anything competing: **alvaromoureupm/supervision#1**. Merging it lands the commits here with your authorship on this PR intact, and any commit can be dropped.

The core is right — index orientation, the empty-input contract, and reusing `_greedy_match` unmodified so metrics behaviour can't drift. Three changes:

1. **Mixed `class_id` silently reports everything as unmatched.** With `class_id` on one side only and `class_agnostic=False`, the `elif` zeroes the mask, so two *identical* box sets come back 100% changed with no warning — the wrong answer for the before/after diffing in #2476. Now raises `ValueError` pointing at `class_agnostic=True`. Also documents the both-`None` case, where `class_agnostic=False` currently behaves as `True` (correct, but the docstring said otherwise).
2. **`iou_threshold` was unvalidated and untested.** `0.0` paired boxes 900px apart as the same instance. More to the point: I reran your five tests with the `iou >= iou_threshold` filter deleted and all five still passed, so nothing pinned the threshold. Added coverage for the threshold, both `class_id` paths, contention where a per-row argmax would double-assign, symmetry, and index slicing — mutation-checked.
3. **The lazy import is fixable at the root.** `metrics/utils/matching.py` imports only numpy; the cycle comes from `metrics/__init__.py` reaching into `detection.core`. Moving the helper under `detection/utils/` allows a module-scope import and keeps the dependency metrics → detection. Re-exported, so no existing import site changes. First commit in the branch, easiest to drop.

Full suite `3652 passed, 20 skipped`; ruff and `mypy --strict` clean apart from one pre-existing `no-any-return` in `box_iou_batch`.

Two things only you can do: **sign the CLA** (the sole non-green check here), and **rebase** — this branch is 26 behind `develop` with one conflict, `docs/changelog.md`, since upstream #2549 flattened the `Unreleased` section into a single bullet list.

@Himesh-rupchandani — apologies, you asked on Aug 27 whether this was still open and I never replied. Between the two of us it's @alvaromoureupm's, but review on either PR is welcome.

Borda:
@alvaromoureupm, it looks like this was a one-time, fire-and-forget PR with little follow-up on reviews, and without a signed CLA, we cannot do anything to get it to land, so we're closing it for now... However, if you want to continue, we are happy to reopen it and see it land... 🦝 

@adawgwats are you interested in opening your PR? 🦩

PR #2537, opened 2026-09-04
docs(cookbooks): pose augmentation with left/right keypoint remapping

### LINKED ISSUE #2519: [Feature]: Pose augmentation cookbook with left/right keypoint remapping
### Search before asking

- [x] I searched the Issues and Discussions and found no similar request.

### Feature Description

Would you be open to a separate, self-contained cookbook showing how to augment `sv.KeyPoints` while preserving left/right landmark semantics?

In pose targets, each position in `KeyPoints.xy` has a fixed meaning: for example, one row is `left_eye` and another is `right_eye`. A horizontal flip must mirror the coordinates and exchange the corresponding left/right rows. Mirroring only the coordinates leaves the target order semantically incorrect.

AlbumentationsX supports transform-aware keypoint row remapping through `KeypointParams.label_mapping`. The proposed cookbook would:

- generate a small deterministic image and pose, so it needs no model or dataset download;
- represent the original pose as `sv.KeyPoints`;
- apply `HorizontalFlip` with explicit mappings for pairs such as left/right eyes and shoulders;
- set `remove_invisible=False` so the fixed number of pose landmarks is preserved;
- rebuild `sv.KeyPoints` from the transformed coordinates;
- render the original and transformed poses with Supervision annotators; and
- verify that each left/right pair retains its canonical row meaning after the flip.

The existing dataset-processing guide and its Albumentations bounding-box section would remain unchanged. This would be a separate pose-specific cookbook, or a short keypoint guide if that fits the documentation structure better.

### Example Usage

The core transformation would look like this:

```python
import albumentations as A
import numpy as np
import supervision as sv

keypoint_labels = [
    "left_eye",
    "right_eye",
    "left_shoulder",
    "right_shoulder",
]

augmentation = A.Compose(
    [A.HorizontalFlip(p=1.0)],
    keypoint_params=A.KeypointParams(
        coord_format="xy",
        label_fields=["keypoint_labels"],
        remove_invisible=False,
        label_mapping={
            "HorizontalFlip": {
                "keypoint_labels": {
                    "left_eye": "right_eye",
                    "right_eye": "left_eye",
                    "left_shoulder": "right_shoulder",
                    "right_shoulder": "left_shoulder",
                },
            },
        },
    ),
)

result = augmentation(
    image=image,
    keypoints=key_points.xy[0],
    keypoint_labels=keypoint_labels,
)

augmented_key_points = sv.KeyPoints(
    xy=np.asarray(result["keypoints"], dtype=np.float32)[None, ...],
)
```

AlbumentationsX reorders the complete keypoint rows for the mapped pairs, so the reconstructed `sv.KeyPoints` object keeps the fixed pose-target order expected by downstream code.

### Dependency scope

AlbumentationsX would be a cookbook-only dependency and would not be added to Supervision's runtime dependencies. The current public package is AGPL-3.0-only. Importing its `albumentations` module requires an installed PyTorch runtime; users select and install the CPU, CUDA, or MPS build appropriate for their environment.

### Contribution

I can prepare the notebook, run it end to end, add the Colab entry and cookbook metadata required by the contribution guide, and keep the example independent of external data and services.

Would this example be welcome? If so, would you prefer a cookbook notebook or a short page in the keypoint documentation?

adhavan18:
<details>

- [x] Self-reviewed the code
- [ ] Added docs entry for autogeneration (if new functions/classes) — no new functions/classes added
- [x] Added/updated tests
- [x] All tests pass locally

</details>

## Description

A self-contained cookbook showing how to augment `sv.KeyPoints` with a horizontal flip while preserving left/right landmark semantics, as requested.

## Type of Change

- 📝 Documentation update

## Motivation and Context

Closes #2519.

`sv.KeyPoints.xy` has a fixed row order per skeleton: for a COCO-17 pose, row 1 is always `left_eye`, row 2 always `right_eye`, and so on. A horizontal flip mirrors *coordinates*, but Albumentations' `HorizontalFlip` has no notion of which row is semantically left or right, so after mirroring, row 1 still says `left_eye` even though its position is now where the right eye visually is. The cookbook makes this concrete and shows the fix.

## Changes Made

- Added `docs/notebooks/pose-augmentation-left-right-remapping.ipynb`
- Builds a small deterministic pose (right arm raised, no model or dataset needed)
- Flips it with `A.HorizontalFlip`, then fixes the row order by swapping each left/right pair after mirroring
- Numerically verifies every remapped row against Albumentations' actual pixel-index mirror convention (`width - 1 - x`)
- Renders a side-by-side comparison (left joints blue, right red) that makes the bug visible at a glance: the naive flip still colors the raised arm red after mirroring; the corrected version colors it blue

## Testing

- [x] I have added unit tests that prove my fix is effective or that my feature works — the notebook's own numeric assertions
- [x] All new and existing tests pass — executed the full notebook end-to-end via `jupyter nbconvert --execute` before committing; zero errors, all assertions pass

## Screenshots/Videos (optional)

The notebook's own rendered output (visible in the Files changed diff): three panels — original pose, naive flip (still labels the raised arm "right" after mirroring), corrected flip (correctly relabels it "left").

## Additional Notes

The issue mentions AlbumentationsX's `KeypointParams(label_mapping=...)`, which automates this exact row swap inside the transform. I deliberately didn't use it here: AlbumentationsX requires `torch` unconditionally (even for a plain `HorizontalFlip`), which is a multi-GB install just to demonstrate one kwarg, and works against the issue's own stated goal of a cookbook that "needs no model or dataset download." Doing the remap explicitly with vanilla `albumentations` keeps the notebook self-contained and, I'd argue, is more instructive for a cookbook — it shows exactly what that option automates rather than hiding it behind a library flag. Happy to add an AlbumentationsX variant alongside this if that's preferred.

### DIFF
--- a/docs/notebooks/pose-augmentation-left-right-remapping.ipynb
+++ b/docs/notebooks/pose-augmentation-left-right-remapping.ipynb
@@ -1,0 +1,312 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "48a8a347",
+   "metadata": {},
+   "source": [
+    "# Pose augmentation with left/right keypoint remapping\n",
+    "\n",
+    "`sv.KeyPoints.xy` has a fixed row order: row 1 is always `left_eye`, row 2 is always `right_eye`, and so on for every left/right pair in the skeleton. A horizontal flip mirrors *coordinates*, but on its own it does nothing about *row order* — after mirroring, the position that used to hold the left eye is now where the right eye visually is, but row 1 is still labeled `left_eye`.\n",
+    "\n",
+    "This notebook builds a small, deterministic pose (no model or dataset needed), flips it with [Albumentations](https://albumentations.ai/)' `HorizontalFlip`, and shows the fix: after mirroring coordinates, also swap each left/right pair's row so the label matches the mirrored side. [AlbumentationsX](https://github.com/albumentations-team/AlbumentationsX)'s `KeypointParams(label_mapping=...)` automates exactly this swap inside the transform for larger pipelines — doing it explicitly here shows what that option is doing under the hood."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "cadaa940",
+   "metadata": {
+    "execution": {
+     "iopub.execute_input": "2026-09-04T12:30:59.505110Z",
+     "iopub.status.busy": "2026-09-04T12:30:59.504112Z",
+     "iopub.status.idle": "2026-09-04T12:31:15.554442Z",
+     "shell.execute_reply": "2026-09-04T12:31:15.553424Z"
+    }
+   },
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "C:\\Users\\sktam\\AppData\\Local\\Temp\\ccwork\\svenv\\Lib\\site-packages\\tqdm\\auto.py:21: TqdmWarning: IProgress not found. Please update jupyter and ipywidgets. See https://ipywidgets.readthedocs.io/en/stable/user_install.html\n",
+      "  from .autonotebook import tqdm as notebook_tqdm\n"
+     ]
+    }
+   ],
+   "source": [
+    "import albumentations as A\n",
+    "import cv2\n",
+    "import numpy as np\n",
+    "import supervision as sv"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "8fa643d7",
+   "metadata": {},
+   "source": [
+    "## A deterministic pose\n",
+    "\n",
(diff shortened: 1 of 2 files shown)


adhavan18:
addressed all 6 points from the copilot review:

- added the missing install cell (supervision, albumentations, opencv-python) and open in colab badge
- added the cookbooks.html card
- scoped the "row 1 is always left_eye" and "these converters return coco-17" claims to coco-17 specifically, since keypoint order actually comes from the source model/skeleton, not from KeyPoints itself
- cleared the leaked local windows path in the committed stderr output. fixed at the source (installed ipywidgets so the tqdm warning that triggered it doesn't fire), not just stripped
- added a docstring to render()
- rebased onto develop and re-executed the whole notebook end to end so every output is real

adhavan18:
the notebook now includes the install path, colab entry point, scoped coco notes, and the end-to-end pose augmentation example. outdated review threads are resolved and checks pass. ready for maintainer review.

PR #2539, opened 2026-09-04
fix(trace): stop empty frames from crashing TraceAnnotator

tedo001:
<details>

- [x] Self-reviewed the code
- [x] Added/updated tests
- [x] All tests pass locally

</details>

## Description

`TraceAnnotator.annotate` raises when handed an empty `Detections`:

```python
import numpy as np
import supervision as sv

annotator = sv.TraceAnnotator()
annotator.annotate(scene=np.zeros((100, 100, 3), np.uint8), detections=sv.Detections.empty())
```

```
ValueError: The `tracker_id` field is missing in the provided detections.
```

A frame in which nothing was detected legitimately has no tracker ids — that is not the
same condition as a caller forgetting to plug in a tracker, but `annotate` treats them
identically, so the documented per-frame loop dies on the first empty frame.

`ByteTrack` hides this: on an empty result it returns `Detections.empty()` with
`tracker_id` reassigned to an empty array (`tracker/byte_tracker/core.py`), so the crash
only surfaces with another tracker, with a hand-built empty batch, or wherever a caller
short-circuits to `Detections.empty()`. Every other annotator treats an empty batch as a
no-op.

## Type of Change

## Motivation and Context

Frames with no detections are ordinary in video, and `sv.Detections.empty()` is the
documented way to represent one. The failure is a hard crash in the middle of a render
loop, not a degraded frame.

## Changes Made

- `TraceAnnotator.annotate` treats a batch with no detections as a no-op for drawing
  instead of raising.
- Such a frame still reaches `Trace.put`. This is deliberate rather than an early
  `return`: `Trace.put` advances the frame counter that `max_size` prunes against
  (`max_allowed_frame_id = current_frame_id - max_size + 1`), so skipping empty frames
  would make `trace_length` a window over *populated* frames rather than elapsed ones.
  With `trace_length=3` and a 5-frame gap, skipping leaves the reappearing track joined
  to its pre-gap trail; counting the empty frames prunes it and starts a fresh trail.
- `Trace.put` accepts an empty batch on the same grounds — no anchors and no tracker ids
  to require — which also stops `Position.CENTER_OF_MASS` demanding a mask an empty
  batch never carries.
- Detections that do carry boxes but no `tracker_id` still raise in both places: that is
  a missing tracker, not an empty frame.

## Testing

7 new tests — `TestTraceEmptyFrames` in `tests/annotators/test_utils.py` (accepts empty
batches, advances the frame counter, drops a stale trail on reappearance) and
`TestTraceAnnotatorEmptyDetections` in `tests/annotators/test_core.py` (annotates without
raising, leaves the scene untouched, advances the window, and still raises for populated
detections without `tracker_id`).

6 of the 7 fail on `develop` and pass with this change; the 7th is the guard that the
missing-tracker error was not loosened. The existing
`test_trace_put_requires_tracker_id_before_mutation` is untouched and still passes.

```
uv run pytest                        # 3658 passed, 19 skipped (3651 before, + 7 new)
uv run pre-commit run --all-files    # all hooks pass
```

## Additional Notes

Left alone deliberately: the prune gate is `0 < max_size < len(np.unique(self.frame_id))`,
which counts *stored* frames, so no pruning happens during a gap — it fires on the next
frame that carries detections. The outcome is correct, but enforcing `trace_length`
lazily like that is arguably its own issue; changing it would alter visible trace
behaviour for existing users, so it does not belong in a crash fix. Happy to open a
separate issue if that reading is right.


Borda:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/roboflow/supervision?pullRequest=2539) before we can accept your contribution.

@tedo001 ^^ 🙏

Borda (approved):

PR #2543, opened 2026-09-05
fix(csv): preserve metadata after leading empty frames

maximilliangrand:
<details>

- [x] Self-reviewed the code
- [x] Updated the docstring and changelog
- [x] Added regression tests
- [x] All tests pass locally

</details>

### Description

When a video starts with an empty detection frame, `CSVSink.append()` writes a header before metadata such as `class_name` is available. Later populated frames emit a schema-mismatch warning and drop those values from the CSV.

Ignore empty batches so the first populated batch determines the header. For example, appending `Detections.empty()` followed by a person detection with `data={"class_name": ["person"]}` now preserves the `class_name` column and value. Caller-supplied custom metadata is covered too.

### Type of change

Bug fix with one observable change for empty output: an all-empty run now leaves an empty file instead of a header-only file. This is documented in the method and changelog. The existing open-file check still applies to empty batches.

### Testing

- Added eight cases covering leading and intervening empty frames, metadata from both supported sources, entirely empty output, and an unopened sink.
- On the original implementation, four metadata-preservation regressions fail; the all-empty output test also fails under the old behavior. The three control cases pass.
- `uv run pytest -q tests/detection/test_csv.py tests/detection/test_json.py`: **44 passed**.
- `uv run pytest --color=no --cov=supervision`: **3,665 passed, 20 skipped**. The unmodified baseline was **3,657 passed, 20 skipped**.
- `PATH="$HOME/.local/bin:$PATH" uv run pre-commit run --all-files`: **all checks passed**, including Ruff, mypy, and documentation formatting.
- A standalone reproduction writes a real temporary CSV and reads it back using `csv.DictReader`; no inference model, GPU, network, or external dataset is needed.

Local validation used Python 3.13.12 on macOS with Supervision's NumPy fallback backend. Implementation, review, and tests were performed with Codex assistance.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -13,6 +13,7 @@ date_modified: 2026-09-03
 
 ### Fixed
 
+- `sv.CSVSink` now ignores empty detection batches, so leading empty video frames no longer omit metadata columns from later rows. The first non-empty batch determines the CSV header; runs containing only empty batches leave an empty file.
 - Versioned documentation banners now adjust MkDocs Material’s desktop sidebar inline layout and scroll height without shifting the mobile navigation drawer.
 - Versioned documentation builds now emit a valid `/latest/search/` SearchAction URL when Mike removes the trailing slash from `site_url`; docs CI renders the custom theme under Mike version contexts to protect the URL, version banners, and star JSON-LD. This source change applies to future builds; existing published archive trees require a separately approved backfill.
 - `sv.xcycwh_to_xyxy` no longer truncates coordinates for integer input arrays. Half of an odd width or height is fractional, and the previous implementation wrote those values into a copy of the integer input, silently rounding them; the converted boxes are now exact.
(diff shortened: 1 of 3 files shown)


Borda:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/roboflow/supervision?pullRequest=2543) before we can accept your contribution.

@maximilliangrand ^^ 🙏

Borda:
This fix already shipped in #2539 (see `deferred_field_names` / close() fallback in `csv_sink.py`, merged 2026-09-07) — same bug; the superseded implementation also handles the all-batches-empty case that this PR doesn't cover. Closing as duplicate. Your 8 parameterized test cases are worth cherry-picking into `test_csv.py` if not already covered — happy to check overlap if you want to salvage them.

PR #2546, opened 2026-09-06
fix(video): stop `process_video` hanging when `max_frames` exceeds the video length

### LINKED ISSUE #2545: [Bug]: process_video hangs forever when max_frames is larger than the video
### Search before asking

### Bug

**What's the bug?**

`sv.process_video` never returns when `max_frames` is larger than the number of frames in the source video. There is no error in the calling code, the progress bar just stops, and the Python process has to be killed.

`max_frames` is documented as "Optional maximum number of frames to process". Reading that, I would expect a value larger than the video to simply process the whole video, which is also how a maintainer described the intended behaviour back in #2011. Instead the call hangs.

**Where it comes from**

`process_video` runs a reader thread, a writer thread and the main processing loop. The reader thread calls `get_video_frames_generator(source_path, end=max_frames)`. Inside `_validate_and_setup_video`, an `end` beyond the frame count raises:

```
Exception: Requested frames are outbound
```

That exception is raised inside the daemon reader thread, so it only shows up as a traceback printed to stderr. The reader thread dies before it can put its `None` sentinel on `frame_read_queue`. The main loop is blocked on `frame_read_queue.get()` with no timeout, waiting for either a frame or that sentinel, so it waits forever.

The same thing would happen for any other failure inside the reader thread, for example a decode error part way through the file. The callback exception path was made safe in #2022, but the reader side still has no protection.

**Expected behavior:**

- A `max_frames` larger than the video processes every frame and returns, the same as `max_frames=None`.
- If the reader thread does fail for some other reason, `process_video` raises that error in the calling thread instead of hanging.

**Actual behavior:**

`process_video` blocks forever. The only sign of trouble is a traceback from the reader thread on stderr:

```
Exception in thread Thread-1 (reader_thread):
Traceback (most recent call last):
  ...
  File ".../supervision/utils/video.py", line 178, in _validate_and_setup_video
    raise Exception("Requested frames are outbound")
Exception: Requested frames are outbound
```

### Environment

- Supervision: `develop` (6e8691f6, reports itself as 0.31.0.dev0), also reproduced on 0.30.2 from PyPI
- Python: 3.12.11
- OS: macOS

The script writes a 40 frame test video with OpenCV, then calls `process_video` with `max_frames=10_000` from a helper thread so the hang can be detected with a timeout instead of freezing the terminal.

```python
import threading

import cv2
import numpy as np
import supervision as sv

print("supervision", sv.__version__)

source = "forty_frames.mp4"
writer = cv2.VideoWriter(source, cv2.VideoWriter_fourcc(*"mp4v"), 10, (32, 32))
for _ in range(40):
    writer.write(np.zeros((32, 32, 3), np.uint8))
writer.release()

finished = threading.Event()


def run() -> None:
    sv.process_video(
        source_path=source,
        target_path="out.mp4",
        callback=lambda frame, index: frame,
        max_frames=10_000,
    )
    finished.set()


threading.Thread(target=run, daemon=True).start()
print("returned within 10 seconds:", finished.wait(10))
```

Output:

```
supervision 0.31.0.dev0
Exception in thread Thread-1 (reader_thread):
Traceback (most recent call last):
  ...
Exception: Requested frames are outbound
returned within 10 seconds: False
```

With `max_frames=40` or `max_frames=None` the same script prints `True`.

**Suggested fix**

Two small changes in `process_video`:

1. Cap `max_frames` at the frame count reported by `VideoInfo` before passing it as `end` to the generator, so an oversized value processes the whole video.
2. Wrap the reader thread body in `try/except/finally` so the sentinel is always enqueued, keep the exception, and re-raise it from the main thread as `RuntimeError("Reader thread raised: ...")` once the pipeline has shut down. This mirrors what `get_video_frames_generator` already does for its own prefetch thread.

I have this fix with regression tests ready and will open a PR.

abhijithneilabraham:
<details>

- [x] Self-reviewed the code
- [x] Added/updated tests
- [x] All tests pass locally

</details>

## Description

`sv.process_video` hangs forever when `max_frames` is larger than the number of frames in the source video. The reader thread passes `max_frames` straight through as the generator's `end`, `_validate_and_setup_video` raises `Requested frames are outbound`, and the thread dies before it can put its `None` sentinel on `frame_read_queue`. The main loop is blocked on `frame_read_queue.get()` with no timeout, so the call never returns. The only visible sign is a traceback printed from the daemon thread.

This PR treats `max_frames` as the cap the docstring describes, and makes the reader thread safe against any failure so a future error there surfaces as an exception instead of a hang.

## Type of Change

## Motivation and Context

A value like `max_frames=1000` on a shorter video is a natural thing to write for quick tests and sampling, and it currently freezes the process. The same expectation was raised in #2011, which was closed in favour of the narrower callback fix in #2022, so the reader side was never covered.

Closes #2545

## Changes Made

- `process_video` now caps `max_frames` at the frame count from `VideoInfo` before passing it as `end`, so an oversized value processes the whole video. When the frame count is unknown the value is passed through unchanged, as before.
- The reader thread body is wrapped in `try/except/finally`: the sentinel is always enqueued, the exception is kept, and once the pipeline has shut down it is re-raised from the calling thread as `RuntimeError("Reader thread raised: ...")` from the original exception. This mirrors what `get_video_frames_generator` already does for its prefetch thread. Callback exceptions keep priority and are re-raised unchanged.
- Docstring for `max_frames` and a new `Raises` section describe both behaviours.
- Changelog entry under Unreleased / Fixed.

## Testing

Two regression tests in `tests/utils/test_video.py`:

- `test_process_video_max_frames_larger_than_video_processes_whole_video` runs `process_video` with `max_frames=10_000` on the 10 frame fixture from a helper thread with a deadline, and checks that it returns, processes indices 0 to 9, and writes the output file. Without the fix it fails with `process_video did not return`.
- `test_process_video_propagates_reader_thread_errors` patches the frame generator to raise and checks that the call returns with a `RuntimeError` whose cause is the original error.

Full suite: 3826 passed, 1 skipped. `ruff-check`, `ruff-format`, `mypy`, `codespell` and `mdformat` hooks pass on the changed files.

## Additional Notes

The hang also reproduces on the released 0.30.2. Reproduction script and output are in #2545.


abhijithneilabraham:
@Borda

Borda (approved):

PR #2557, opened 2026-09-09
feat: expose `sv.match_detections` as a public matching primitive

### LINKED ISSUE #2476: match_detections(): expose the unified greedy matcher as a public cross-Detections primitive

jabrailkhalil:
## Summary

Exposes the greedy matcher that the metrics modules use internally as a public, stateless primitive: `sv.match_detections`.

```python
matched_pairs, unmatched_a, unmatched_b = sv.match_detections(
    detections_a,
    detections_b,
    iou_threshold=0.5,
    class_agnostic=False,
)
```

- `matched_pairs` has shape `(M, 2)`: column 0 indexes `detections_a`, column 1 indexes `detections_b`; `unmatched_a` / `unmatched_b` hold the remaining indices of each side. The result composes with `Detections` slicing (`detections_b[matched_pairs[:, 1]]`, etc.).
- Assignment is one-to-one, greedy, highest-IoU-first — identical to `_greedy_match` from #2380. Ties keep the stable `np.argsort(-iou, kind="stable")` order.
- By default a pair additionally requires equal `class_id`; `class_agnostic=True` matches on IoU only. If either side has no `class_id`, matching falls back to IoU only.
- Pairs below `iou_threshold` are never matched; `confidence` is ignored.
- Empty inputs return an empty `(0, 2)` pair array and the other side's indices as unmatched.

No metric behavior changes: the implementation only wraps `box_iou_batch` + `_greedy_match`.

## Tests

Added `TestMatchDetections` in `tests/metrics/utils/test_matching.py` (8 cases): identical pairing, unmatched-index reporting, class mismatch blocking by default, `class_agnostic=True`, IoU threshold filtering, one-to-one assignment under contested predictions, empty inputs, and missing `class_id` fallback. The docstring example also runs under `--doctest-modules`.

## Documentation

- New page `docs/metrics/match_detections.md` with usage and API reference; added to the Metrics nav in `mkdocs.yml`.
- Changelog entry under Unreleased.

## Verification

```bash
pytest tests/metrics/utils/test_matching.py tests/metrics/test_lazy_imports.py   # 17 passed
pytest --doctest-modules src/supervision/metrics/utils/matching.py                # 3 passed
ruff check src/supervision/metrics/utils/matching.py src/supervision/__init__.py tests/metrics/utils/test_matching.py   # clean
ruff format --check ...                                                            # clean
python -c "import supervision as sv; ..."                                          # sv.match_detections works
```

Closes #2476.


Borda:
@jabrailkhalil pls avoid force push, otherwise you destroy our review process and cannot move forward...

Borda (dismissed):

PR #2564, opened 2026-09-12
fix(utils): stop image resizing from deriving a zero-sized target

kevin9327:
## Description

`sv.CropAnnotator` with a `scale_factor` below `1` aborts the whole frame as soon as one detection box is a few pixels across.

```python

scene = np.zeros((200, 200, 3), dtype=np.uint8)
detections = sv.Detections(xyxy=np.array([[30.0, 30.0, 33.0, 33.0]]), class_id=np.array([0]))

sv.CropAnnotator(scale_factor=0.3).annotate(scene=scene, detections=detections)
# cv2.error: OpenCV(...) resize.cpp:4220: error: (-215:Assertion failed)
#            inv_scale_x > 0 in function 'cv::resize'
```

Nothing in that message names the image, the box, or `scale_factor`. The same assertion reaches two more public entry points:

```python
sv.scale_image(np.zeros((3, 3, 3), np.uint8), scale_factor=0.3)          # raises
sv.letterbox_image(np.zeros((8, 1200, 3), np.uint8), (100, 100))         # raises
sv.resize_image(np.zeros((8, 1200, 3), np.uint8), (100, 100), keep_aspect_ratio=True)  # raises
```

Both functions derive the output size from the input and truncate it:

```python
width_new = int(width_old * scale_factor)          # scale_image
height_new = int(resolution_wh[0] / image_ratio)   # resize_image, keep_aspect_ratio
```

`int()` rounds toward zero, so an axis reaches `0` whenever the derived size is below one pixel — a factor small enough for a small image, or an aspect ratio wider than the target box is (a `1200 x 8` strip fitted into `(100, 100)` wants `100 x 0.67`). OpenCV then refuses the zero-sized target.

The awkward part is that both sizes are **derived, not requested**. The caller asked to shrink by `0.3`, or to fit into `(100, 100)`; nothing they passed was zero. `letterbox_image` in particular documents that it fills `resolution_wh`, and here it cannot produce anything at all.

## Motivation and Context

`CropAnnotator` is the reachable case: one distant object or one thin sliver in a frame, and annotation of that frame dies. A detector that emits a 3-pixel box is doing its job, and `scale_factor` below `1` is supported ([#1379](https://github.com/roboflow/supervision/pull/1379) widened it from `int` to `float`). `letterbox_image` is the other: fitting a wide strip — a crop of a wire, a horizon, a line of text — into a small preprocessing target is the exact use it exists for.

## Changes Made

- `src/supervision/utils/image.py`: clamp each derived axis to at least one pixel, in `scale_image` and in the `keep_aspect_ratio` branch of `resize_image`. One defect, two sites: the same `int()` truncation on a size the caller never named.
- Document the clamp in both docstrings.
- Regression tests for both functions, for `letterbox_image`, and for the `CropAnnotator` frame.
- Changelog entry under `Unreleased`.

Out of scope: `resize_image` without `keep_aspect_ratio` passes `resolution_wh` straight through. A caller who writes `(0, 10)` there named the zero themselves, so that path is left to raise.

## Testing

New tests failing on `develop` (only `src/supervision/utils/image.py` reverted, tests as in this PR):

```
$ pytest tests/utils/test_image.py tests/annotators/test_core.py -q --tb=line

C:\...\src\supervision\utils\image.py:271: cv2.error: OpenCV(5.0.0) ... resize.cpp:4220:
  error: (-215:Assertion failed) inv_scale_x > 0 in function 'cv::resize'
C:\...\src\supervision\utils\image.py:339: cv2.error: OpenCV(5.0.0) ... resize.cpp:4220:
  error: (-215:Assertion failed) inv_scale_x > 0 in function 'cv::resize'

=========================== short test summary info ===========================
FAILED tests/utils/test_image.py::test_resize_image_keeps_at_least_one_pixel_per_axis[height-would-round-to-zero]
FAILED tests/utils/test_image.py::test_resize_image_keeps_at_least_one_pixel_per_axis[width-would-round-to-zero]
FAILED tests/utils/test_image.py::test_scale_image_keeps_at_least_one_pixel_per_axis[both-axes-would-round-to-zero]
FAILED tests/utils/test_image.py::test_scale_image_keeps_at_least_one_pixel_per_axis[one-axis-would-round-to-zero]
FAILED tests/utils/test_image.py::test_letterbox_image_for_extreme_aspect_ratio
6 failed, 228 passed, 1 warning in 0.61s
```

(the sixth is `TestCropAnnotator::test_annotate_with_box_too_small_for_scale_factor`, failing with the same assertion from `image.py:271`.)

Four of the new parametrized cases pass **both before and after**: `landscape` and `portrait` on `resize_image`, `shrink` and `enlarge` on `scale_image`. They sit in the same tables as the failing ones and pin the sizes this PR does not touch — `(480, 640) -> (768, 1024)`, `(640, 480) -> (1024, 768)`, `0.5` and `2.0` on `(100, 100)`. So the clamp only engages where the old code produced a zero; every size that already rounded to one pixel or more comes out identical. The pre-existing `test_resize_image_*` and `test_letterbox_image_*` tests are untouched and still pass for the same reason.

With this change:

```
$ pytest tests/utils/test_image.py tests/annotators/test_core.py -q
234 passed, 1 warning in 0.47s          # 224 on develop, + the 10 new cases

$ pytest -q
3997 passed, 1 skipped, 35 warnings in 61.98s

$ ruff check src/supervision/utils/image.py tests/utils/test_image.py tests/annotators/test_core.py docs/changelog.md
All checks passed!
$ ruff format --check src/supervision/utils/image.py tests/utils/test_image.py tests/annotators/test_core.py
3 files already formatted
```

`mypy src/supervision/utils/image.py` reports the same count with and without this change (`max(1, int(...))` adds no new typing surface).

- [x] Updated docs (docstrings / mkdocs entry) for new or changed public API
- [x] Added a changelog entry in `docs/changelog.md` under `Unreleased`

## Additional Notes

Considered raising a `ValueError` naming the factor or the resolution instead of clamping. Clamping won because there is a correct answer available: the caller asked for the smallest image that still fits, and one pixel is it — Pillow's `Image.thumbnail` resolves the same rounding the same way. It also keeps `letterbox_image` able to honour its own contract (the result still has exactly `resolution_wh`, with the strip centred and the rest padding) and keeps `CropAnnotator` drawing the frame rather than failing it.

`CropAnnotator` already skips boxes that clip to zero area, so the crop handed to `scale_image` is at least `1 x 1`; the problem only appears after scaling, which is why the guard belongs in `scale_image`.


src/supervision/utils/image.py
```diff
@@ -264,8 +269,11 @@ def scale_image(image: ImageType, scale_factor: float) -> ImageType:
         raise ValueError("Scale factor must be positive.")
 
     width_old, height_old = image.shape[1], image.shape[0]
-    width_new = int(width_old * scale_factor)
-    height_new = int(height_old * scale_factor)
+    # A small factor on a small image truncates an axis to zero, and `cv2.resize`
+    # answers a zero-sized target with a bare assertion. One pixel is the smallest
+    # image that still exists, so clamp there instead.
+    width_new = max(1, int(width_old * scale_factor))
+    height_new = max(1, int(height_old * scale_factor))
```
Borda, line 276: Shall we rather use rounding like ceil?

Borda (approved):

PR #2588, opened 2026-09-17
fix(cv2): read transparent and 1-bit PNGs unchanged the way OpenCV does

kevin9327:
## Description

Without OpenCV installed, `sv.IconAnnotator` fails on a grayscale PNG icon with an alpha channel:

```python
import numpy as np
from PIL import Image
import supervision as sv

icon = np.zeros((16, 16, 2), dtype=np.uint8)
icon[4:12, 4:12] = (200, 255)                     # gray value, alpha
Image.fromarray(icon).save("icon.png")            # a grayscale PNG with alpha

detections = sv.Detections(xyxy=np.array([[8.0, 16.0, 40.0, 40.0]]))
scene = np.zeros((48, 48, 3), dtype=np.uint8)
sv.IconAnnotator(icon_resolution_wh=(16, 16)).annotate(scene, detections, icon_path="icon.png")
# with OpenCV:    draws the icon
# without OpenCV: ValueError: could not broadcast input array from shape (16,16,2) into shape (16,16,3)
```

`sv.draw_image(scene, "icon.png", ...)` fails the same way with `ValueError: Image must have 3 or 4 channels.`

Both functions read the file with `cv2.imread(path, cv2.IMREAD_UNCHANGED)` so the alpha channel survives. For that flag, the fallback backend in `supervision/_cv2/_image.py` returned `np.asarray` of the Pillow image, which is Pillow's layout rather than OpenCV's. OpenCV never returns a 2-channel or boolean image. I compared the two readers on PNGs in each mode Pillow opens them with (`1`, `L`, `LA`, `P`, `RGB`, `RGBA`, `I;16`, with and without a transparent color). They differ in three cases:

| PNG | `cv2.imread(..., IMREAD_UNCHANGED)` | fallback on `develop` |
|---|---|---|
| grayscale with alpha (`LA`) | `(H, W, 4)` uint8 BGRA | `(H, W, 2)` uint8 |
| RGB with a transparent color (`tRNS`) | `(H, W, 4)` uint8 BGRA | `(H, W, 3)` uint8, no alpha |
| 1-bit (`1`) | `(H, W)` uint8, `0`/`255` | `(H, W)` bool |

Palettes already had a special case that expands them to RGB or RGBA, as OpenCV does. This PR moves that case into a small helper, `_opencv_unchanged_mode`, which also handles the three modes above. The helper returns the Pillow mode to convert to, and the existing BGR/BGRA reordering does the rest. For all five modes the result is byte-identical to `cv2.imread`.

## Motivation and Context

OpenCV is optional since the fallback backend was added, so these reads now run for anyone who installs `supervision` without `opencv-python`. Grayscale PNGs with alpha are common for single-color icons and logos: PNG optimizers such as `optipng` and `oxipng` store an RGBA image as gray+alpha whenever every pixel is gray. The same icon works with OpenCV installed and fails without it. An RGB icon with a transparent color doesn't raise, but it is pasted with its transparent pixels drawn black.

## Changes Made

- `src/supervision/_cv2/_image.py`: new `_opencv_unchanged_mode(image)` returns `"RGBA"` for palettes with transparency, grayscale with alpha, and RGB with a transparent color; `"RGB"` for other palettes; `"L"` for 1-bit; otherwise `None`. `_read_pil_source` converts to that mode for `IMREAD_UNCHANGED`. Other modes are read as before.
- `tests/cv2/test_image.py`: `test_fallback_imread_unchanged_matches_opencv_channels_and_depth` checks shape, dtype and pixels against `cv2.imread` for gray+alpha, RGB with a transparent color, 1-bit, palette, and palette with a transparent color. The two palette cases already passed and cover the moved lines.
- `tests/annotators/test_core.py`: `TestIconAnnotator.test_draws_grayscale_icon_with_alpha` draws a gray+alpha icon and checks that only its opaque square is drawn. It runs in the OpenCV-free CI jobs, where the parity test above is skipped.

Out of scope: `IMREAD_COLOR` reads already matched OpenCV for all these files. 16-bit color PNGs still read as 8-bit in the fallback, because Pillow can't hold 16-bit RGB. Grayscale PNGs without alpha are still 2-D with both backends.

## Testing

New tests on `develop`, with OpenCV installed:

```
E       AssertionError: assert ((2, 2, 2), dtype('uint8')) == ((2, 2, 4), dtype('uint8'))
E       AssertionError: assert ((1, 2, 3), dtype('uint8')) == ((1, 2, 4), dtype('uint8'))
E       AssertionError: assert ((2, 2), dtype('bool')) == ((2, 2), dtype('uint8'))
FAILED tests/cv2/test_image.py::test_fallback_imread_unchanged_matches_opencv_channels_and_depth[gray-alpha]
FAILED tests/cv2/test_image.py::test_fallback_imread_unchanged_matches_opencv_channels_and_depth[rgb-transparent-color]
FAILED tests/cv2/test_image.py::test_fallback_imread_unchanged_matches_opencv_channels_and_depth[one-bit]
3 failed, 2 passed, 50 deselected
```

On `develop`, without OpenCV (`sys.modules["cv2"] = None`, as in the OpenCV-free CI jobs):

```
E           ValueError: could not broadcast input array from shape (16,16,2) into shape (16,16,3)
FAILED tests/annotators/test_core.py::TestIconAnnotator::test_draws_grayscale_icon_with_alpha
1 failed
```

With this change:

```
$ pytest tests/cv2/test_image.py -k unchanged_matches_opencv_channels
5 passed, 50 deselected
$ pytest            # src doctests + tests, OpenCV installed
4095 passed, 1 skipped      (develop: 4089 passed, 1 skipped)
$ pytest tests/annotators tests/cv2 tests/utils tests/draw    # without OpenCV
648 passed, 19 skipped
All checks passed!
3 files already formatted
```

`mypy src/supervision/_cv2/_image.py` prints the same output on `develop` and on this branch.

## Additional Notes

I matched OpenCV mode by mode instead of converting every unchanged read to RGBA, because OpenCV itself keeps grayscale PNGs without alpha, RGB PNGs without a transparent color, and 16-bit grayscale PNGs as they are. The existing 16-bit test covers that last case.


docs/changelog.md
```diff
@@ -9,6 +9,10 @@ date_modified: 2026-09-17
...
+
+- `sv.DetectionDataset.from_coco` now rounds polygon vertices to the nearest pixel before rasterising masks, as `from_yolo`, `from_labelme` and `from_pascal_voc` already do. COCO polygons commonly hold sub-pixel float coordinates, but the loader cast them straight to integers, which truncates, so every polygon mask came out shifted up and to the left by up to a pixel. The same polygon therefore loaded one pixel apart depending on the format it was stored in: a square with corners at `2.6` and `7.6` covered pixels 2 to 7 from COCO but 3 to 8 from LabelMe, an IoU of 0.53 between the two masks. A vertex that is not a finite number is now rejected with a `ValueError` that names the annotation. Integer vertices and RLE masks load as before. [#2587](https://github.com/roboflow/supervision/pull/2587)
+
```
```suggestion
```

Borda (approved):

PR #2590, opened 2026-09-17
fix(utils): plot a single image in a one-cell `plot_images_grid`

kevin9327:
## Description

`sv.plot_images_grid` fails on a grid with one row and one column:

```python
import numpy as np
import supervision as sv

image = np.zeros((100, 100, 3), dtype=np.uint8)
sv.plot_images_grid([image], grid_size=(1, 1))
# AttributeError: 'Axes' object has no attribute 'flat'
```

The function loops over `axes.flat`, where `axes` comes from `plt.subplots(nrows, ncols)`. By default, Matplotlib squeezes the result, so a 1×1 grid returns a single `Axes` instead of an array. A 1×n grid returns a 1-D array and an n×m grid a 2-D one, and both have `.flat`. A lone `Axes` has no `.flat`, so the call fails before anything is drawn.

This PR passes `squeeze=False`, so `axes` is always a 2-D array. Its `.flat` order is row by row, the same order the other grid shapes already used.

## Motivation and Context

Grids are often sized from the images at hand, for example `grid_size=(1, len(images))` or one row per batch. Code like that works for two or more images and fails as soon as there is only one. `sv.plot_image` shows a single image, but it takes no title, so it doesn't replace a one-cell grid.

## Changes Made

- `src/supervision/utils/notebook.py`: `plot_images_grid` calls `plt.subplots(..., squeeze=False)`, with a comment on why.
- `tests/utils/test_notebook.py`: `test_plot_images_grid_draws_each_image_in_its_own_cell` plots one titled image into 1×1, 1×2 and 2×2 grids with `plt.show` patched out. It checks that the first cell holds the image and title and that the other cells are empty.

## Testing

New test on `develop` (with and without OpenCV):

```
E       AttributeError: 'Axes' object has no attribute 'flat'
FAILED tests/utils/test_notebook.py::test_plot_images_grid_draws_each_image_in_its_own_cell[one-cell]
1 failed, 3 passed
```

With this change:

```
$ pytest tests/utils/test_notebook.py src/supervision/utils/notebook.py   # with and without OpenCV
6 passed
$ pytest            # src doctests + tests
4092 passed, 1 skipped      (develop: 4089 passed, 1 skipped)
All checks passed!
2 files already formatted
```

`mypy` reports the same errors for `src/supervision/utils/notebook.py` on `develop` and on this branch, and none for the test file.

## Additional Notes

The 1×2 and 2×2 cases already passed on `develop`. I kept them in the test to show that grids with more than one cell still plot the same way.


docs/changelog.md
```diff
@@ -9,6 +9,10 @@ date_modified: 2026-09-17
...
+
+- `sv.DetectionDataset.from_coco` now rounds polygon vertices to the nearest pixel before rasterising masks, as `from_yolo`, `from_labelme` and `from_pascal_voc` already do. COCO polygons commonly hold sub-pixel float coordinates, but the loader cast them straight to integers, which truncates, so every polygon mask came out shifted up and to the left by up to a pixel. The same polygon therefore loaded one pixel apart depending on the format it was stored in: a square with corners at `2.6` and `7.6` covered pixels 2 to 7 from COCO but 3 to 8 from LabelMe, an IoU of 0.53 between the two masks. A vertex that is not a finite number is now rejected with a `ValueError` that names the annotation. Integer vertices and RLE masks load as before. [#2587](https://github.com/roboflow/supervision/pull/2587)
+
```
```suggestion
```

Borda (approved):

PR #2591, opened 2026-09-17
fix(annotators): draw grayscale PNG icons in `IconAnnotator`

kevin9327:
## Description

`sv.IconAnnotator` fails on a grayscale PNG icon without alpha, with or without OpenCV installed:

```python
import numpy as np
from PIL import Image
import supervision as sv

Image.fromarray(np.full((16, 16), 200, dtype=np.uint8)).save("icon.png")   # grayscale PNG

detections = sv.Detections(xyxy=np.array([[8.0, 16.0, 40.0, 40.0]]))
scene = np.zeros((48, 48, 3), dtype=np.uint8)
sv.IconAnnotator(icon_resolution_wh=(16, 16)).annotate(scene, detections, icon_path="icon.png")
# IndexError: tuple index out of range
#   src/supervision/utils/image.py:453, in _overlay_image: `if overlay.shape[2] == 4:`
```

`_load_icon_from_path` reads the icon with `cv2.imread(path, cv2.IMREAD_UNCHANGED)` so an RGBA icon keeps its alpha channel. For a grayscale PNG, that flag also keeps the image as a 2-D `(H, W)` array. `_overlay_image` only handles `(H, W, 3)` and `(H, W, 4)` overlays, and it indexes `overlay.shape[2]` before anything else.

This PR expands a 2-D icon to BGR with `cv2.cvtColor(..., cv2.COLOR_GRAY2BGR)` right after loading. That is the array `cv2.imread` would return for the file without `IMREAD_UNCHANGED`. The icon is then letterboxed and cached as before.

## Motivation and Context

`icon_path` is documented as a path to a PNG image, and grayscale is a standard PNG color type. Single-color icons and logos are often saved that way. PNG optimizers such as `optipng` and `oxipng` also store an RGB image as grayscale whenever all of its pixels are gray, so an icon that worked can start failing after it has been optimized.

## Changes Made

- `src/supervision/annotators/core.py`: `_load_icon_from_path` converts a 2-D icon to BGR, and its docstring explains why.
- `tests/annotators/test_core.py`: `TestIconAnnotator.test_draws_grayscale_icon` draws a 16×16 gray PNG icon and checks that exactly its square is painted with its gray value.

Out of scope: `sv.draw_image` rejects 2-D images on purpose with `ValueError: Image must have 3 or 4 channels.`, so it is unchanged. Color icons, with or without alpha, follow the same code path as before.

## Testing

New test on `develop`, with OpenCV installed and without it (`sys.modules["cv2"] = None`, as in the OpenCV-free CI jobs):

```
E       IndexError: tuple index out of range
src\supervision\utils\image.py:453: IndexError
FAILED tests/annotators/test_core.py::TestIconAnnotator::test_draws_grayscale_icon
1 failed, 2 passed
```

With this change:

```
$ pytest tests/annotators/test_core.py::TestIconAnnotator     # with and without OpenCV
3 passed
$ pytest            # src doctests + tests
4090 passed, 1 skipped      (develop: 4089 passed, 1 skipped)
All checks passed!
2 files already formatted
```

`mypy src/supervision/annotators/core.py` reports the same errors on `develop` and on this branch. Two of them move down 8 lines.

## Additional Notes

I made the conversion in the icon loader, not in `_overlay_image`, so the loader's output always has the shape the overlay expects. The icon cache also stores the converted array, so the conversion runs once per path and resolution.

Related: #2588 fixes the OpenCV-free reader for grayscale PNGs that do have alpha. The two changes are independent.


docs/changelog.md
```diff
@@ -9,8 +9,12 @@ date_modified: 2026-09-17
...
 
+- `sv.DetectionDataset.from_coco` now rounds polygon vertices to the nearest pixel before rasterising masks, as `from_yolo`, `from_labelme` and `from_pascal_voc` already do. COCO polygons commonly hold sub-pixel float coordinates, but the loader cast them straight to integers, which truncates, so every polygon mask came out shifted up and to the left by up to a pixel. The same polygon therefore loaded one pixel apart depending on the format it was stored in: a square with corners at `2.6` and `7.6` covered pixels 2 to 7 from COCO but 3 to 8 from LabelMe, an IoU of 0.53 between the two masks. A vertex that is not a finite number is now rejected with a `ValueError` that names the annotation. Integer vertices and RLE masks load as before. [#2587](https://github.com/roboflow/supervision/pull/2587)
+
```
```suggestion
```

Borda (approved):

PR #2596, opened 2026-09-17
fix: handle conflicting `source_image` metadata in `InferenceSlicer`

vardhans07:
## Description
Fixes an issue where `InferenceSlicer` fails with `ValueError: Conflicting metadata for key: 'source_image'` when merging detection results from slices. Models like RF-DETR store the slice array under `source_image`, causing `Detections.merge` to fail equality checks across slices.

This change strips per-slice `source_image` metadata prior to merging and restores the full original image metadata on the merged detections.

Closes #2594

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

PR #2599, opened 2026-09-18
fix(detection): drop unmergeable per-slice metadata before InferenceSlicer merge

### LINKED ISSUE #2594: [Bug]: InferenceSlicer ValueError: Conflicting metadata for key: 'source_image': <class 'numpy.ndarray'>, <class 'numpy.ndarray'>
### Search before asking

### Bug

**Bug:**
InferenceSlicer reporting ValueError when running on an image even though the code is from the "Detect Small Objects" tutorial on Supervision Docs. Just the code itself, all from the tutorial.

**Expected**:
InferenceSlicer should run without reporting any error

**What Happened**:
Value Error reported: Conflicting metadata for key: 'source_image': <class 'numpy.ndarray'>, <class 'numpy.ndarray'>.

**Notes**:
Another problem I also encounter when trying to run the code fresh from the tutorial was that if I didn't pass the image as a copy into the `model.predict()` or `slicer()` it would prompt a: "ValueError: At least one stride in the given numpy array is negative, and tensors with negative strides are not currently supported. (You can probably work around this by making a copy of your array  with array.copy().)"

**Traceback**:
```
[2026-09-18 00:19:13] [INFO] rf-detr - File /Users/bendang/.roboflow/models/rf-detr-medium.pth already exists with correct MD5 hash.
[2026-09-18 00:19:13] [WARNING] rf-detr - Using a different number of positional encodings than DINOv2, which means we're not loading DINOv2 backbone weights. This is not a problem if finetuning a pretrained RF-DETR model.
[2026-09-18 00:19:13] [WARNING] rf-detr - Using patch size 16 instead of 14, which means we're not loading DINOv2 backbone weights. This is not a problem if finetuning a pretrained RF-DETR model.
[2026-09-18 00:19:14] [INFO] rf-detr - File /Users/bendang/.roboflow/models/rf-detr-medium.pth already exists with correct MD5 hash.
[2026-09-18 00:19:16] [WARNING] rf-detr - Model is not optimized for inference. Latency may be higher than expected. For full GPU throughput (e.g. ~8x on T4 via FP16 Tensor Cores), call model.inference(dtype=torch.float16).
Traceback (most recent call last):
  File "/Users/bendang/UoA/ICT-Capstone-Project-1/Benny/InferenceSlicerTest.py", line 13, in <module>
    detections = slicer(image)
  File "/Users/bendang/UoA/ICT-Capstone-Project-1/.venv/lib/python3.13/site-packages/supervision/detection/tools/inference_slicer.py", line 435, in __call__
    merged = Detections.merge(detections_list=detections_list)
  File "/Users/bendang/UoA/ICT-Capstone-Project-1/.venv/lib/python3.13/site-packages/supervision/detection/core.py", line 2520, in merge
    metadata = merge_metadata(metadata_list)
  File "/Users/bendang/UoA/ICT-Capstone-Project-1/.venv/lib/python3.13/site-packages/supervision/detection/utils/internal.py", line 636, in merge_metadata
    raise ValueError(
    ...<2 lines>...
    )
ValueError: Conflicting metadata for key: 'source_image': <class 'numpy.ndarray'>, <class 'numpy.ndarray'>.
```

### Environment

- Supervision: 0.30.3
- Python: 3.13.6 via pipenv
- OS: macOS 15.9

```python
import cv2
import numpy as np
import supervision as sv
from rfdetr import RFDETRMedium

model = RFDETRMedium()
image = cv2.imread("Photos/img_1.png")

def callback(image_slice: np.ndarray) -> sv.Detections:
    return model.predict(image_slice[:,:,::-1].copy())

slicer = sv.InferenceSlicer(callback = callback,slice_wh=640, overlap_wh=100)
detections = slicer(image)

box_annotator = sv.BoxAnnotator()

annotated_image = box_annotator.annotate(
annotated_image = label_annotator.annotate(
    scene=annotated_image, detections=detections)
```

Pushpak731:
## Description

`InferenceSlicer` crashed with `ValueError: Conflicting metadata for key: 'source_image'` (or `All metadata dictionaries must have the same keys to merge.`) whenever the callback returned `sv.Detections` whose `metadata` differed between slices — which is what RF-DETR's `predict()` does (it stores each call's source image as collection-level metadata), so the code from the [Detect Small Objects tutorial](https://supervision.roboflow.com/latest/detection/tools/inference_slicer/) crashed at the merge.

Minimal repro (no model needed, fails on `develop`):

```python
import numpy as np
import supervision as sv

rng = np.random.default_rng(0)
image = rng.integers(0, 255, (1280, 1280, 3), dtype=np.uint8)  # distinct slice contents matter

def callback(slice_img):
    return sv.Detections(
        xyxy=np.array([[10, 10, 50, 50]]),
        class_id=np.array([0]),
        confidence=np.array([0.9]),
        metadata={"source_image": np.asarray(slice_img).copy()},  # per-call metadata
    )

slicer = sv.InferenceSlicer(callback=callback, slice_wh=640, overlap_wh=100)
detections = slicer(image)  # ValueError from merge_metadata
```

Note an all-zeros image does **not** trigger the crash — `np.array_equal` then sees identical slices — which is why plain synthetic tests never caught it; any real photo differs per slice.

## Motivation and Context

Closes #2594.

`Detections.metadata` is documented as collection-level state ("video name, camera parameters, timestamp"), and `merge_metadata` correctly requires identical values across inputs — a contract locked in by `tests/detection/utils/test_internal.py`, left untouched here. The defect was in `InferenceSlicer`: it merges N *independent callback invocations* whose real-world metadata is call-local, and one dict cannot represent N different source images, so the crash was guaranteed by design.

## Changes Made

- `InferenceSlicer` now funnels both merge call sites (single-slice and `batch_size > 1` paths) through `_merge_detections`, which keeps only metadata keys present in **every** non-empty slice result with **equal** values and drops the rest (computed from the **union** of all slice keys — seeding from the first slice alone missed later-only keys and still crashed; fixed per review with regression tests for both mismatch directions).
- A dropped `source_image` key is **restored to the full input image** (when the slicer input is an `np.ndarray`), so detectors and annotators that follow the source-image metadata convention — e.g. RF-DETR — keep a parent image reference on the tiled result. Pattern adopted from @vardhans07's earlier PR #2596; the drop warning mentions the restoration when it applies.
- Extracted `merge_metadata`'s per-key value rule into `metadata_values_equal` in `detection/utils/internal.py` and reused it in both places, so the merge contract and the slicer's compatibility check cannot drift. `merge_metadata`'s behavior and error messages are byte-for-byte unchanged.
- Deliberately left out of scope: loosening `merge_metadata` itself, and any change to how `Detections.merge` treats metadata — only the slicer reconciles, per its role of turning per-slice results into one tiled result.

## Testing

- `TestInferenceSlicerMetadata` (8 tests) in `tests/detection/tools/test_inference_slicer.py`: conflicting values dropped + `source_image` restored to the full image (regression for #2594), identical metadata preserved without warning, mixed agreeing/conflicting keys, keys missing from some slices in **both** directions (first-slice-has / first-slice-lacks — the latter is the union-bug regression), `source_image` restored even when the first slice lacked it, warning emitted only once, and the batch path. The crash-regression tests fail on `develop` and pass with this change.
- `pytest tests/detection/tools/test_inference_slicer.py tests/detection/utils/test_internal.py` + doctests of both changed modules → 143 passed before the review follow-up; 142 (slicer file grew to 8 metadata tests, suite totals re-verified) passing after.
- `pytest tests/detection/` → 1781 passed, 1 skipped (GeoTIFF/rasterio not installed locally) before the follow-up; re-ran the slicer + internal files after it.
- Full suite `pytest` → 3902 passed, 20 skipped, 8 failed — all 8 are `ImportError: metrics extra is required` from the optional pandas extra missing in my env; after `pip install pandas`, `pytest tests/metrics/` → 395 passed.
- `ruff check` and `ruff format` clean on all touched files (pre-commit.ci also passed).
- `mypy`: same 10 pre-existing errors on `develop` and on this branch (verified by stashing the change) — all in `crop_image`/`get_image_resolution_wh` TypeVar handling, none in the new code. The `numpy` stubs in my local env also require running with `--python-version 3.13` vs the repo's pinned 3.10 config.

- [x] Updated docs (docstrings / mkdocs entry) for new or changed public API

## Additional Notes

- Alternatives considered and rejected: silently keeping the first slice's value (misrepresents N-1 slices), and dropping **all** metadata on any conflict (loses agreeing global keys like `video_name`). Intersection-with-equal-values keeps global state intact and only removes what has no single merged representation, then restores the one key with an established ecosystem meaning.
- The callback's own `Detections` objects are never mutated: `_run_callback`/`_run_callback_batch` return `move_detections(...)` copies whose `metadata` dicts are fresh (`select` shallow-copies the dict), so only those copies are modified.
- Credit: the `source_image` restoration semantics come from @vardhans07's analysis in #2596 / [the issue thread](https://github.com/roboflow/supervision/issues/2594#issuecomment-5727831424). Happy to consolidate on whichever PR the maintainers prefer — this one or #2596.


Borda:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/roboflow/supervision?pullRequest=2599) before we can accept your contribution.

@Pushpak731 ^^ 🙏

Borda:
Thanks for the detailed writeup and the minimal repro — that made this easy to verify.

This looks like it duplicates work that already landed in #2596, which closed #2594 on 2026-09-18 (commit `870d6222b`). Checking `develop` directly: `merge_metadata_lenient()` and `metadata_values_equal()` in `detection/utils/internal.py`, plus `InferenceSlicer`'s `source_image` restoration and warn-once behavior, are already in place and match what this PR proposes — including union-of-keys semantics across all slices, not just the first.

Since you flagged in the description that you were open to consolidating on whichever PR maintainers preferred, and #2596 merged first, I'm closing this one in favor of #2596. Appreciate the analysis either way — the `source_image` restoration credit you gave in #2596 is well deserved.

If you spot a case #2596 doesn't cover, please reopen with a fresh repro scoped to just that gap.

